PR #41, opened 2018-02-09
Needed refactor of getIDFromCategoryString

davidmanzanares:
PR to fix https://github.com/CartoDB/renderer-prototype/issues/23


davidmanzanares:
@IagoLast 

- I understand that you probably need more context. Feel free to ping me at any time to solve this. In any case, it is clear that it isn't easy to jump into a new codebase and that initial CRs won't be as useful as the ones that will come after. You don't need to make a deep very detail oriented CR.
- IMHO, I cannot be waiting for this.
- There is already a configured linter which is automatically run by Travis. Other than that I just expect common sense regarding code style.

IagoLast:
How can I test this? I think we should create a example under the examples folder.

davidmanzanares:
You can test it with the style commented on https://github.com/CartoDB/renderer-prototype/issues/23

The examples folder lacks all the styling stuff, this is not a good starting point since this is not the most basic functionality.

davidmanzanares:
I understand that this PR doesn't add any documentation nor tests nor examples.

However, since the status of the repo is without docs, without tests, and without examples, I don't think that this is the place to do that.

Therefore, I'm gonna merge.

IagoLast:
¯\_(ツ)_/¯

PR #46, opened 2018-02-12
Multiple examples added

davidmanzanares:


davidmanzanares:
I'm using that convention because it's easier to copy and paste the URLs into the browser.

example/styling/cielab.2.html
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Single layer | CARTO</title>
```
IagoLast: wrong titles

davidmanzanares: Yep, on all of them. However, for all the styling examples there is much copy-pasted boilerplate. We could use a JSON for all of these styling examples, right?

src/index.js
```diff
@@ -9,12 +9,15 @@
...
 
 // Source namespace has dataset and SQL
 const source = { Dataset };
 
+
 export { Layer };
 export { source };
 export { Style };
 export { parseStyle }; // TODO(@iago): temporal export, we must find a better place
+export { expressions };
```
IagoLast: we want to export this as `carto.expressions` or we want `carto.style.<expression>` ? @davidmanzanares @Jesus89

Jesus89: I vote for `carto.style.expression` but we can talk about this in the `Expressions` epic.

PR #70, opened 2018-02-15
Drawordering

davidmanzanares:
Fix for https://github.com/CartoDB/renderer-prototype/issues/20

This PR allows to client-side order the feature at drawing time. It works by issuing many rendering draw commands that filters by width range like: first render everything with width less than 1, then everything with width between 1 and 2, then...

This could be improved in the future to order by any other expression.

I removed some style commented code too.


example/styling/ordering_asc.html
```diff
@@ -0,0 +1,58 @@
...
+
+    const auth = {
+        user: 'dmanzanares-ded13',
+        apiKey: ''
+    };
+
+    const source = new carto.source.Dataset('tx_0125_copy_copy', auth, { cartoURL: 'carto-staging.com' });
+    const s = carto.expressions;
+    const style = new carto.Style({
+        width: s.floatMul(s.floatMul(s.sqrt(s.floatDiv(s.property('amount'), 5000)), 20), s.floatMul(s.floatAdd(s.floatDiv(s.zoom(), 4000), 0.01), 1.5)),
```
IagoLast: I think the expressions will be clear if we split them

src/core/renderer.js
```diff
@@ -577,7 +578,27 @@ Renderer.prototype.refresh = function (timestamp) {
 
 
     const s = 1. / this._zoom;
-    tiles.forEach(tile => {
+
+    // TODO remove hack
+    let orderer = null;
```
IagoLast: 🌈  For a newcomer with no context like me this code fragment looks like an ancient Egyptian writing. Of course it is my responsibility to know the code base and I don't (yet) but some improvements could me made in order to improve understandability.

the `Renderer.refresh` method is quite big I think if we [extract functions](https://refactoring.com/catalog/extractMethod.html) and give those functions a good descriptive name the code readability will improve a lot.

example/styling/ordering_asc.html
davidmanzanares: Done

src/core/renderer.js
davidmanzanares: I really liked the "ancient Egyptian writing", but having long functions with WebGL/OpenGL code is not frowned upon by many OpenGL developers because it actually helps to reduce errors. This is basically because if you extract a method, you need to explain quite well how that method is going to change WebGL state, and what kind of WebGL state it does expect. This is extremely error-prone. 

I recommend seeing Jonathan Blow talks, particularly this one that talks about this issue: https://youtu.be/JjDsP5n2kSM?t=27m40s

IagoLast: I could agree when the new created function has side effects but in this case we are talking of a `pure function` so there is no risk of changing webgl state.

```javascript
function goodFunctionName(tiles) {
    let orderer = null;
    let orderingMins = null;
    let orderingMaxs = null;

    if (tiles.length > 0) {
        orderer = tiles[0].style._order;
    }
    if (orderer instanceof ordering.Asc) {
        orderingMins = Array.from({ length: 16 }, (_, i) => (15 - i) * 2);
        orderingMaxs = Array.from({ length: 16 }, (_, i) => i == 0 ? 1000 : (15 - i + 1) * 2);
    }
    else if (orderer instanceof ordering.Desc) {
        orderingMins = Array.from({ length: 16 }, (_, i) => i * 2);
        orderingMaxs = Array.from({ length: 16 }, (_, i) => i == 15 ? 1000 : (i + 1) * 2);
    }
    else {
        orderingMins = [0];
        orderingMaxs = [1000];
    }
    return { orderingMins, orderingMaxs };
}
```

And in the renderer.refresh:

```javascript
const { orderingMins, orderingMaxs } = getOrdererMinandMax(tiles);
```

PR #74, opened 2018-02-15
Basic API implementation

Jesus89:
Related to: https://github.com/CartoDB/renderer-prototype/issues/43

Changes:
- [x] Added Basic API 
- [x] Added documentation for API (jsdoc)
- [x] Adder CartoError handling
- [x] Added ~35 unit tests
- [x] Updated all examples
- [x] Improve unit tests speed

Note: this PR contains no code refactor that will be addressed in other PRs.

### Source

```js
const source = new carto.source.Dataset('table'); // use default auth and config
const source = new carto.source.Dataset('table', auth);
const source = new carto.source.Dataset('table', auth, config);
const source = new carto.source.SQL('SELECT * FROM table'); // default auth and config
const source = new carto.source.SQL('SELECT * FROM table', auth, config);
```

### Style

```js
const style = new carto.Style(`
  color: rgba(0,1,0,1)
`);
const s = carto.style.expression;
const style = new carto.Style({
  color: s.rgba(0,1,0,1);
});
style.getResolution();
style.getColor();
style.getWidth();
style.getStrokeColor();
style.getStrokeWidth();
style.getOrder();
```

### Layer

```js
var layer = new carto.Layer(id, mySource, myStyle);
layer.addTo(map);
layer.setSource(source);
layer.setStyle(style);
```

### Default Auth/config

```js
carto.setDefaultAuth({
  user: 'cartojs-test',
  apiKey: '123456789012345678901234567890123456789'
});
carto.setDefaultConfig({
  serverURL: 'https://{user}.cartox.com'
});
```


package.json
```diff
@@ -21,6 +21,7 @@
     "cartocolor": "^4.0.0",
     "earcut": "^2.1.2",
     "jsep": "^0.3.2",
+    "lodash": "^4.17.5",
```
IagoLast: any chance of not using loadash? IMO we should create an `utils` folder

src/api/setup/config-service.js
```diff
@@ -0,0 +1,57 @@
...
+function checkConfig(config) {
+    if (config) {
+        if (!_.isObject(config)) {
+            throw new CartoValidationError('setup', 'configObjectRequired');
+        }
+        checkServerURL(config.serverURL);
+    }
+}
+
+function checkServerURL(serverURL) {
```
IagoLast: missing underscore prefix

package.json
Jesus89: I think we can use it right now in order to save time, but for the alpha or beta release we could improve the bundle size with this proposal if required.

src/api/style.js
```diff
@@ -0,0 +1,287 @@
...
+
+        // REVIEW THIS vv
+        this.updated = true;
+        this._changeCallback = null;
+        this._styleSpec.color.parent = this;
+        this._styleSpec.width.parent = this;
+        this._styleSpec.strokeColor.parent = this;
+        this._styleSpec.strokeWidth.parent = this;
+        this._styleSpec.order.parent = this;
+        this._styleSpec.color.notify = () => { this._changed(); };
```
IagoLast:
```javascript
this._styleSpec.color.notify = this._changed
```

src/api/style.js
```diff
@@ -0,0 +1,287 @@
+import * as _ from 'lodash';
```
IagoLast: we should review this file, maybe we can split it in smaller filers

Jesus89: Yes, of course, in other PR we should refactor this file and its interaction with the Layer and the Renderer: private methods called out of the class, on-the-fly properties. There are also some tests ignored because of the `parent` and `notify` addition in the StyleSpec properties.

This file is expected to grow with new properties. Instead of updating always the same file we should follow a method to allow extension with new files.

package.json
davidmanzanares: I have my doubts too, but I wanted some of their utils in the past and now I think it would probably save us some time. I would try to use native functions as much as possible, but I think it's fine for now.

src/api/style.js
davidmanzanares: Yes, please, don't mix refactors with features and bug fixing, the diff is already huge.

davidmanzanares:
Hey, this seems good, but the editor is broken, can you fix it? @Jesus89

Jesus89:
Hi @davidmanzanares. In the API the config object has a `serverURL`. I have replaced the old property `cartoURL` used in the example by the new one `serverURL`. Please check that all is OK.

Jesus89:
Note: two examples are not working because of a bug in the parser. I have created a ticket for that: https://github.com/CartoDB/renderer-prototype/issues/78.

PR #79, opened 2018-02-16
Hide attribution locally

IagoLast:
Hide the map attribution when served locally (url contains localhost), this is usefull in screenshot testing since fonts are rendered with quite big differences depending on the OS/browser.

This is a diff between OSX/Linux see Red pixels pointing differences in the attribution div.

diff


Jesus89:
It is not very nice to change the public examples in order to pass the tests...

IagoLast:
@Jesus89 the script only removes the attribution when served locally, public examples `documentation.carto.com/whatever` will look the same

Jesus89:
I mean this lines `<script src="../hide-attribution.js"></script>`

PR #84, opened 2018-02-16
StyleBlendTo

davidmanzanares:
Reintroduce blending between styles at `Layer`.

Adds a new `BounceEaseIn` interpolator.

### DIFF
--- a/src/api/layer.js
+++ b/src/api/layer.js
@@ -95,6 +95,29 @@ export default class Layer {
     }
 
     /**
+     * Blend the current style with another style
+     *
+     * @param {carto.Style} style - style to blend to
+     *
+     * @memberof carto.Layer
+     * @api
+     */
+    blendToStyle(style, ms = 3000, interpolator = undefined) {
+        this._checkStyle(style);
+        if (this._style) {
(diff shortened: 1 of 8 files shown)


src/api/layer.js
```diff
@@ -94,6 +94,29 @@ export default class Layer {
...
 
+    /**
+     * Blend the current style with another style
+     *
+     * @param {carto.Style} style - style to blend to
+     *
+     * @memberof carto.Layer
+     * @api
+     */
+    blendToStyle(style, ms = 3000, interpolator = undefined) {
```
Jesus89: Please, add tests for this.

test/unit/api/layer.test.js
```diff
@@ -4,14 +4,15 @@ import Layer from '../../../src/api/layer';
 
 describe('api/layer', () => {
     let source;
-    let style;
+    let style, style2;
```
Jesus89: Why are you using `style2` instead of `style`?

davidmanzanares: conceptually it makes more sense to blend to another style, when we improve the testing, we should probably need another style

PR #85, opened 2018-02-19
Add options to select basemap

### LINKED ISSUE #35: Support for additional basemaps in demo
@dv343 

As discussed, in the renderer interface that I'm using, would it be possible to add the additional CARTO basemaps?

Dark Matter: labels under, no labels
Positron: labels on top, labels under, no labels
Voyager: labels on top, labels under, no labels

thanks!

rochoa:
It lacks proper integration as the API doesn't retain the layer order when changing the style, so the new basemap style is applied but the data layer _dissapears_.

This will fix #35.


src/api/layer.js
```diff
@@ -132,7 +132,13 @@ export default class Layer {
                         dataframe.visible = false;
                     });
             });
-        });
+        };
+
+        if (map.isStyleLoaded()) {
```
Jesus89: I miss some tests to cover this.

PR #86, opened 2018-02-19
Fix addTo(map) to work after map already loaded

rochoa:
I found this problem. When the map is already loaded, the `layer.addTo(map)` doesn't have any effect.

This is just a proposal. I would love @IagoLast and @Jesus89 could take some time to think about this because I'm pretty sure we could have this problem with other API methods, e.g. `remove()`.


Jesus89:
Indeed. This is a bug that we have already talked about. I agree with this approach but IMO we need to provide error handing for the `addTo` function. We also need to add tests to cover this code.

rochoa:
❤️

PR #89, opened 2018-02-20
FSAA

davidmanzanares:
This PR adds Full Screen Anti Aliasing (FSAA) to lines and polygons rendering (to anything that is not points, actually).

MSAA is faster, but it is quite limited in WebGL 1, and it is problematic to use while integrating this library with MGL.

Points do not need this since their shader already includes a distance field based AA.

Fixes https://github.com/CartoDB/renderer-prototype/issues/36
Partially fixes https://github.com/CartoDB/renderer-prototype/issues/68

### DIFF
--- a/src/core/renderer.js
+++ b/src/core/renderer.js
@@ -88,8 +88,6 @@ Renderer.prototype._initGL = function (gl) {
 
     this._AATex = gl.createTexture();
     this._AAFB = gl.createFramebuffer();
-    gl.bindFramebuffer(gl.FRAMEBUFFER, this._AAFB);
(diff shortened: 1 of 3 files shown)


src/core/renderer.js
```diff
@@ -506,6 +512,26 @@ Renderer.prototype._computeDrawMetadata = function () {
     return drawMetadata;
 };
 
+function getOrderingRenderBuckets(tiles) {
```
IagoLast: ^^

davidmanzanares: :slightly_smiling_face:

PR #96, opened 2018-02-22
Simple map

Jesus89:
Related to: https://github.com/CartoDB/renderer-prototype/issues/87

This is a very basic implementation of a simple non-interactive carto map. For layers with animated style it repaints always. If there is no layer with animated style it repaints until all the data is loaded.

This is not documented for the public API and it will be used in the integration tests.

- [x] Added carto map
- [x] Added carto map integrator
- [x] Update layer.addTo function
- [x] Updated examples

```js
const map = new carto.Map({
  container: 'map',
  background: 'black'
};

[...]

layer0.addTo(map);
layer1.addTo(map, 'layer0');
```

### DIFF
--- a/example/basic/single-layer-mapbox.html
+++ b/example/basic/single-layer-mapbox.html
@@ -1,10 +1,9 @@
 <!DOCTYPE html>
 <html>
     <head>
-        <title>Multi layer | CARTO</title>
+        <title>Single layer | CARTO</title>
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <meta charset="UTF-8">
-        <script src="../hide-attribution.js"></script>
         <!-- Include CARTO GL JS -->
         <script src="../../dist/carto-gl.js"></script>
         <!-- Include Mapbox GL JS -->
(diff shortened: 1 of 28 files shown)


example/basic/single-layer-mapbox.html
```diff
@@ -4,7 +4,6 @@
         <title>Single layer | CARTO</title>
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <meta charset="UTF-8">
-        <script src="../hide-attribution.js"></script>
```
davidmanzanares: Why are you removing this?

src/api/integrator/mapbox-gl.js
```diff
@@ -40,17 +41,18 @@ class MGLIntegrator {
...
         this.map.setCustomWebGLDrawCallback(layerId, (gl, invalidate) => {
-            if (!this.invalidateMGLWebGLState) {
-                this.invalidateMGLWebGLState = invalidate;
+            if (!this.invalidateWebGLState) {
+                this.invalidateWebGLState = invalidate;
                 this.notifyObservers();
                 this.renderer._initGL(gl);
             }
-            paintCallback();
+            layer.paintCallback();
```
davidmanzanares: By removing the callbacks and passing the layer instead of just the layerID you are coupling the integrator to the Layer, but ok, maybe it's more clear this way

src/api/style.js
```diff
@@ -146,6 +146,10 @@ export default class Style {
         return this._styleSpec.order;
     }
 
+    isAnimated() {
+        return this.getColor().isAnimated() || this.getWidth().isAnimated();
```
davidmanzanares: Check strokes too, please

src/core/renderer.js
```diff
@@ -716,7 +716,7 @@ Renderer.prototype.refresh = function (timestamp) {
     this.computePool = [];
 
     tiles.forEach(t => {
-        if (t.style.getColor().isAnimated() || t.style.getWidth().isAnimated()) {
+        if (t.style.isAnimated()) {
```
davidmanzanares: this loop is legacy code and should probably be removed since the requestAnimationFrame logic is not going to be on the core renderer

example/basic/single-layer-mapbox.html
Jesus89: This is not used anymore since the examples are separated from the acceptance tests. This will be addressed for all the examples in the "Examples ticket".

src/api/integrator/mapbox-gl.js
Jesus89: Yes, the layer needs to match this interface now. But this code requires more work/refactor in the future.

src/core/renderer.js
Jesus89: :eyes:

src/api/style.js
Jesus89: I have borrowed this from https://github.com/CartoDB/renderer-prototype/pull/96/files#diff-ba2b399628ae0c09a9b06d41faff9b64L719. I will add the srokes.

example/basic/single-layer-mapbox.html
davidmanzanares: :+1:

PR #97, opened 2018-02-22
Fixed race condition

davidmanzanares:
The main problem was that state was modified before knowing if the style and source was ok.

With this PR, Style and Windshaft class state is only modified after a successful Maps API instantiation.

This PR adds promises to the return of setStyle/blendStyle functions so that users can receive asynchronous errors.


src/api/layer.js
```diff
@@ -100,13 +101,17 @@ export default class Layer {
...
         this._checkStyle(style);
-        if (this._style) {
-            this._style.onChange(null);
-        }
-        this._style = style;
-        this._style.onChange(this._styleChanged.bind(this));
-        // Force style changed event
-        this._styleChanged();
+        return this._styleChanged(style).then(r => {
+            console.log('success set', r);
```
Jesus89: Remove this log

src/api/layer.js
```diff
@@ -124,12 +129,19 @@ export default class Layer {
...
             style.getStrokeWidth().blendFrom(this._style.getStrokeWidth(), ms, interpolator);
-            this._style.onChange(null);
         }
-        this._style = style;
-        this._style.onChange(this._styleChanged.bind(this));
-        // Force style changed event
-        this._styleChanged();
+
+        return this._styleChanged(style).then(r => {
+            console.log('success blend', r);
```
Jesus89: Same here

src/api/layer.js
```diff
@@ -181,28 +193,37 @@ export default class Layer {
...
-                this._style._compileColorShader(this._integrator.renderer.gl, metadata);
-                this._style._compileWidthShader(this._integrator.renderer.gl, metadata);
-                this._style._compileStrokeColorShader(this._integrator.renderer.gl, metadata);
-                this._style._compileStrokeWidthShader(this._integrator.renderer.gl, metadata);
-            }
+        const originalPromise = this.requestData(style);
+        if (!originalPromise) {
+            // The previous stored metadata is still valid
+            recompile(this.metadata);
+            return Promise.resolve(undefined);
```
Jesus89: This `undefined` is not required

PR #104, opened 2018-02-23
Remove shader dead code

davidmanzanares:
This should have been removed much, much sooner.

It's mostly dead-code in the polygon renderer (tris) shader.


davidmanzanares:
Yes, stroke and width are not supported for polygons. The references to those attributes are there because the shader was initially copied from the point rendering one. I modified a long time ago this to ignore them, but I forgot to remove the rest of the code.

PR #112, opened 2018-02-26
Remove deprecated functionality

davidmanzanares:
Fixes: https://github.com/CartoDB/renderer-prototype/issues/22 https://github.com/CartoDB/renderer-prototype/issues/9


src/core/style/expressions/ramp.js
```diff
@@ -10,20 +9,15 @@ export default class Ramp extends Expression {
      * @param {*} minKey Optional
```
IagoLast: misleading doc

src/core/style/expressions/ramp.js
```diff
@@ -10,20 +9,15 @@ export default class Ramp extends Expression {
...
      * @param {*} maxKey Optional
      */
-    constructor(input, palette, minKey, maxKey) {
-        input = implicitCast(input);
-        if (minKey !== undefined) {
-            minKey = implicitCast(minKey);
-            maxKey = implicitCast(maxKey);
+    constructor(input, palette, ...args) {
+        if (args.length > 0) {
```
IagoLast: Any special reason for throwing an error on this validation? could we just ignore the extra parameters?

src/core/style/expressions/ramp.js
```diff
@@ -10,20 +9,15 @@ export default class Ramp extends Expression {
...
+        input = implicitCast(input);
         palette = implicitCast(palette);
         super({ input: input });
-        if (minKey === undefined) {
-            minKey = float(0);
-            maxKey = float(1);
-        }
-        this.minKey = minKey.expr;
-        this.maxKey = maxKey.expr;
+        this.minKey = 0;
```
IagoLast: i´d extract the defaults to constants

src/core/style/expressions/ramp.js
davidmanzanares: The reason is that is easy to try to use the old functionality, and assume that things don't work as expected because there is a bug. However, I think we should do this kind of validation more often. It's much better to throw an error than to render something with unexpected behavior (relative to the user).

PR #118, opened 2018-02-27
Improve filtering

IagoLast:
[Filtering](https://github.com/CartoDB/renderer-prototype/issues/14)
---

- Add a new shader to support  filters
- Add new filtering expressions
- Add examples and tests


src/core/style/functions.js
```diff
@@ -1,9 +1,19 @@
+/**
+ *  @api
+ *  @namespace carto.style.expressions
+ *
+ *  @description
+ *  # CARTO GL
```
IagoLast: wrong desc

PR #122, opened 2018-02-28
Metadata filtering

davidmanzanares:
Adds filtering support for drawMetadata.
Adds eval() to expressions.

Only works with viewport Metadata, not global, global metadata won't take filtering into account.

## TODO

- [x] Add tests for `blend`
- [x] Add tests for `viewport aggregations`


davidmanzanares:
I would remove:

> Implement .and test interpolator
 Implement and test opacity
 Implement and test ramp

 from the list, they are not needed for filtering, since you cannot filter by color, and linear interpolation as an approximation for other interpolations is not that bad.

IagoLast:
gonna need help for "non-global" tests and we are done 🚀

davidmanzanares:
I've completed the tests and review the entire PR. This is ready to merge, I think. Are you ok with that @IagoLast ? :rocket: 

About the viewport aggregation tests, I've seen you created a fake WebGL context, I'm not gonna block this PR for that (particularly when you didn't check the GL function calls), but I really think that this is not the way to go in general. In the future, we should refactor the styling expressions to leave non-WebGL logic apart, so we can test that without relying on fake WebGL contexts.

IagoLast:
> About the viewport aggregation tests, I've seen you created a fake WebGL context, I'm not gonna block this PR for that (particularly when you didn't check the GL function calls), but I really think that this is not the way to go in general. In the future, we should refactor the styling expressions to leave non-WebGL logic apart, so we can test that without relying on fake WebGL contexts.

100% about isolating webgl from tests, but for this case we only need a single mocked function so I'm ok with it

PR #124, opened 2018-03-01
Make instantiate function async and extract small functions

IagoLast:
- Make instantiate function async and extract small functions

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -124,15 +124,7 @@ export default class Windshaft {
 
     async _instantiate(minimunRequiredScheme, resolution, filters) {
         const conf = this._getConfig();
-        let agg = {
-            columns: {},
-            dimensions: {},
-            placement: 'centroid',
-            resolution: resolution,
-            threshold: 1,
-        };
-
-        this._updateAggregationWithMinimunRequiredScheme(agg, minimunRequiredScheme);
+        const agg = this._generateAggregation(minimunRequiredScheme, resolution);
         const select = this._buildSelectClause(minimunRequiredScheme);
         let aggSQL = this._buildQuery(select);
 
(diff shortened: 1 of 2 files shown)


src/client/windshaft.js
```diff
@@ -82,7 +82,7 @@ export default class Windshaft {
             }
             return promise;
         }
-        if (!this.url) {
+        if (!this.urlTemplate) {
```
davidmanzanares: I think this is dead code, after previous refactors, can you check it?

src/client/windshaft.js
```diff
@@ -122,70 +122,76 @@ export default class Windshaft {
...
+    async _instantiate(minimunRequiredScheme, resolution, filters) {
+        const conf = this._getConfig();
+        const agg = await this._generateAggregation(minimunRequiredScheme, resolution);
+        const select = this._buildSelectClause(minimunRequiredScheme);
+        let aggSQL = this._buildQuery(select);
+
+        const query = `(${aggSQL}) AS tmp`;
+        const metadata = await this.getMetadata(query, minimunRequiredScheme, conf);
+
+        // If the number of features is higher than the minimup set server filtering.
```
davidmanzanares: "minimup set" => "minimum, enable"

src/client/windshaft.js
```diff
@@ -122,70 +122,76 @@ export default class Windshaft {
         return this._numCategories;
     }
 
-    _instantiate(MNS, resolution, filtering) {
-        let agg = {
-            threshold: 1,
-            resolution: resolution,
+    async _instantiate(minimunRequiredScheme, resolution, filters) {
```
davidmanzanares: The first argument is called in all the project `MNS` or `minimumNeededSchema`, please, don't introduce a new alias.

src/client/windshaft.js
```diff
@@ -122,70 +122,76 @@ export default class Windshaft {
...
-                agg.dimensions[name] = name;
-            }
-        });
-        const select = MNS.columns.map(name => name.startsWith('_cdb_agg_') ? getBase(name) : name).concat(['the_geom', 'the_geom_webmercator']);
-        let aggSQL = `SELECT ${select.filter((item, pos) => select.indexOf(item) == pos).join()} FROM ${this._source._query ? `(${this._source._query}) as _cdb_query_wrapper` : this._source._tableName}`;
-        agg.placement = 'centroid';
-        const query = `(${aggSQL}) AS tmp`;
 
-        const conf = {
+        minimunRequiredScheme.columns
```
davidmanzanares: same (naming)

src/client/windshaft.js
```diff
@@ -122,70 +122,76 @@ export default class Windshaft {
...
+                    };
+                } else {
+                    aggregation.dimensions[name] = name;
+                }
+            });
+
+        return aggregation;
+    }
+
+    _buildSelectClause(minimunRequiredScheme) {
```
davidmanzanares: same (naming)

src/client/windshaft.js
```diff
@@ -287,6 +303,11 @@ export default class Windshaft {
             });
     }
 
+    _getTileUrl(x, y, z) {
+        const s = this._subdomains[Math.floor(Math.random() * this._subdomains.length) + 0];
```
davidmanzanares: Please, use a counter instead of `random()`, the counter will distribute requests more evenly. The `random()` will distribute them similarly with sufficient requests, but it will generate the same subdomain twice in a row many times.

Issue #16, opened 2018-02-01
Timestamps: proper support for temporal maps (client-side)


davidmanzanares:
Right now the only supported data types are numbers and categories in the form of strings.

We need to support timestamps and give the proper Date/Time utils to work with them.

The WWI_Ships example is a good starting point since it is working thanks to a column which is the number of days since 1917 instead of using the timestamp column.

We need to define the syntax/style API first.

I imagine something like:
`torque($date, from: '15/01/1917', to: '1/01/1922', speed: '15 days'/'1 second', range: '10 days', trail: '10 days')`

Where `from` and `to` are the start and end timestamps (should be optional, by default it should take the minimum and maximum timestamps of the dataset),
`speed` is the simulated time that passes each real-time second of the simulation, in this case 15 days per second,
`range` is the time duration in which the input is considered active (i.e. temporalSimulation returns 1),
`trail` is the time duration in which the input is considered semiactive (i.e. temporalSimulation returns a number greater than 0 and less than 1).

The output of this function is (regarding the input $date) is:
```
Output
 ^
|1.66|
|1.33|
|1.0 |        ______
|0.66|       /      \
|0.33|      /        \
|0.0 | ____/          \____
=================================> Time
```

Note:
I've used named arguments, IMO this improves clarity, but we haven't implemented this yet.

davidmanzanares:
After talking with @rochoa and discarding the named parameters example because adding it would difficult the consistency between JS API and String API. We came to the following examples:
```
s.torque(s.prop('d'), 10);
s.torque(s.prop('d'), 10, s.fade(0.1, 0.6), s.dateRange('15/01/1917', '1/01/1922'));
s.torque(s.prop('d'), 10, s.default, s.dateRange('15/01/1917', '1/01/1922'));
```
This would require:

- [ ] Adding `torque(property, durationInSeconds=10, fade=s.default, range=s.default)` which is similar to the already implemented `near()`, but it has better semantics and is more controllable (see `fade()`).
- [ ] Adding `s.DEFAULT` as a symbolic constant for default parameters, it would be accessed with just `default` in the string API.
- [ ] Adding `fade(arg1, arg2=s.default)`, which will have the following semantics: with one parameter, it will set fadeIn and fadeOut to that parameter, when called with two parameters, fadeIn and fadeOut will be assigned to them respectively.
- [ ] Adding `dateRange(from, to)`, which will parse `from` and `to` to get date timestamps.
[...]

PR #134, opened 2018-03-06
Torque

davidmanzanares:
Partially adds #16 

This PR adds a new `torque` expression to create animated maps in a more straightforward manner.


src/core/renderer.js
```diff
@@ -5,6 +5,8 @@ import { Asc, Desc } from './style/functions';
 
 const HISTOGRAM_BUCKETS = 1000;
 
+const initialTimestamp = Date.now();
```
IagoLast: why here and not in the constructor?

🌈  uppercase and snake case for global constants

src/core/style/expressions/expression.js
```diff
@@ -61,6 +61,14 @@ export default class Expression {
         return this._getChildren().map(child => child._getDrawMetadataRequirements()).reduce(schema.union, schema.IDENTITY);
     }
 
+    /**
+     * Pre-rendering routine. Should establish the current timestamp in seconds since an arbitrary point in time as needed.
+     * @param {Number} timestamp
+     */
+    _setTimestamp(timestamp) {
```
IagoLast: 🌈 if this method is public used from the renderer) why the underscore prefix?

src/core/style/expressions/expression.js
```diff
@@ -61,6 +61,14 @@ export default class Expression {
         return this._getChildren().map(child => child._getDrawMetadataRequirements()).reduce(schema.union, schema.IDENTITY);
     }
 
+    /**
+     * Pre-rendering routine. Should establish the current timestamp in seconds since an arbitrary point in time as needed.
+     * @param {Number} timestamp
```
IagoLast: `Number` --> `number`

src/core/style/expressions/torque.js
```diff
@@ -0,0 +1,103 @@
...
+ * Create an animated temporal filter (torque)
+ *
+ * @param {carto.style.expression.expression} input input to base the temporal filter,
+ * if input is a property, the beginning and end of the animation will be determined by the minimum and maximum timestamps of the property on the dataset,
+ * this can be problematic if outliers are present. Otherwise input must be a float expression in which 0 means beginning of the animation and 1 means end.
+ * @param {Number} duration duration of the animation in seconds
+ * @param {carto.style.expression.fade} fade fadeIn/fadeOut configuration
+ * @return {carto.style.expressions.torque}
+ *
+ * @example `width:    2
```
IagoLast: 🌈  wrong indentation

src/core/style/expressions/torque.js
```diff
@@ -0,0 +1,103 @@
...
+        const input = this.input.eval(feature);
+        const cycle = this._cycle.eval(feature);
+        const duration = this.duration;
+        const fadeIn = this.fade.fadeIn.eval(feature);
+        const fadeOut = this.fade.fadeOut.eval(feature);
+        return 1 - clamp(Math.abs(input - cycle) * duration / (input > cycle ? fadeIn : fadeOut), 0, 1);
+    }
+}
+
+function clamp(x, min, max) {
```
IagoLast: I think the clamp function could be moved to `utils` since is widely used

src/core/style/expressions/utils.js
```diff
@@ -1,6 +1,8 @@
 import { float, category } from '../functions';
 import Expression from './expression';
 
+export const DEFAULT = undefined;
```
IagoLast: ?

src/core/style/expressions/viewportAggregation.js
```diff
@@ -40,7 +40,11 @@ function generateAggregattion(metadataPropertyName, global) {
             return this.property._getMinimumNeededSchema();
         }
         _getDrawMetadataRequirements() {
-            return { columns: [this.property.name] };
+            if (!global) {
```
IagoLast: is this related to te issue?

src/core/style/expressions/torque.js
```diff
@@ -0,0 +1,103 @@
...
+// fadeIn and fadeOut of 0.5 seconds
+filter:       torque($day, 40, fade(0.5))
+
+`
+ *
+ * @memberof carto.style.expressions
+ * @name fade
+ * @function
+*/
+export class Fade extends Expression {
```
IagoLast: I think `Fade` should be in its own file

src/core/renderer.js
davidmanzanares: why here and not in the constructor?

Because this is just needed for numeric stability, no need to have a per-instance constant

src/core/style/expressions/expression.js
davidmanzanares: Consistency with all the other expression methods that are public to the renderer, yet private to the user.

src/core/style/expressions/torque.js
davidmanzanares: In the example? How should it be?

src/core/style/expressions/torque.js
davidmanzanares: Totally agree

src/core/style/expressions/torque.js
IagoLast: - keep the `*` at the beginning of the lines
- the example starts in the next line (after @example tag) 
- there is no need for the backtick
-

src/core/style/expressions/utils.js
davidmanzanares: @rochoa and I talked about adding a `DEFAULT` expression to be more explicit in the API to be used like: `torque($day, default, fade(5))`. I think we should have it at least on the string API, I don't have much opinion regarding the JS API

src/core/style/expressions/viewportAggregation.js
davidmanzanares: `torque()` triggers this function, and since it didn't check for `global` it made the animations slower. It is related, although the change would make sense without torque too.

src/core/style/expressions/torque.js
IagoLast:
```javascript
 * @example <caption> fadeIn of 0.1 seconds, fadeOut of 0.3 seconds </caption>
 * const s = carto.style.expressions;
 * new carto.Style({
 *  filter: s.torque($day, 40, s.fade(0.1, 0.3))
 * });
 * 
 * @example <caption>   fadeIn and fadeOut of 0.5 seconds </caption>
 * const s = carto.style.expressions;
 * new carto.Style({
 *  filter: s.torque($day, 40, s.fade(0.5))
 * });
```

src/core/style/expressions/torque.js
davidmanzanares: I haven't moved it since, at least right now, `Fade` is entirely made to be used with `torque`

src/core/style/expressions/expression.js
davidmanzanares: :+1:

src/core/style/expressions/torque.js
davidmanzanares: I haven't moved it since, at least right now, `Fade` is entirely made to be used with `torque`

src/core/style/expressions/torque.js
davidmanzanares: :+1: Thank you, updated

PR #135, opened 2018-03-07
Timestamp support

davidmanzanares:
https://github.com/CartoDB/renderer-prototype/issues/16
Add support for literal timestamps: '15/01/1917'
Add support for ´linear´ with timestamps

Example with usage included.


src/client/windshaft.js
```diff
@@ -182,8 +185,11 @@ export default class Windshaft {
         return aggregation;
     }
 
-    _buildSelectClause(MRS) {
-        return MRS.columns.map(name => name.startsWith('_cdb_agg_') ? getBase(name) : name).concat(['the_geom', 'the_geom_webmercator']);
+    _buildSelectClause(MRS, dateFields = []) {
+        return MRS.columns.map(name => name.startsWith('_cdb_agg_') ? getBase(name) : name).map(
+            name => dateFields.includes(name) ? name + '::text' : name
```
IagoLast: `::text` ? 

what is this?

src/client/windshaft.js
```diff
@@ -398,6 +408,17 @@ export default class Windshaft {
             numFields.map((name, index) => {
                 properties[index + catFields.length][i] = Number(f.properties[name]);
             });
+            datesField.map((name, index) => {
```
IagoLast: 🌈 use foreach if you dont need the resulting array

IagoLast:
I´ll add docs for the torque expression in this PR

src/client/windshaft.js
davidmanzanares: A hack, a workaround Maps API since Maps API skip SQL `date` properties if they are not casted like this. This will go away when Maps API improves timestamp support.

davidmanzanares:
"I´ll add docs for the torque expression in this PR"  Right :+1:

davidmanzanares:
@IagoLast I've just checked the sources, the `torque` docs were already there (except a note about linear()), they were introduced into master by other PR.

IagoLast:
lgtm

PR #136, opened 2018-03-08
Integration tests

Jesus89:
Related to: https://github.com/CartoDB/renderer-prototype/issues/87
---

```
yarn test:render
yarn test:render:prepare
```

- Define integration tests
    - `test.geojson`
    - `test.js`
    - `test.png`
- Render without Mapbox GL
    - Use `carto.Map`
- Add GeoJSON provider
    - Add `carto.source.GeoJSON`
    - Support for:
        - [x] Point
        - [x] LineString
        - [x] MultiLineString
        - [x] Polygon
        - [x] MultiPolygon
- Add png references generator
- Add integration tests
    - Detect test files recursively
    - Use html template for the tests
    - Load test.geojson file


src/api/map.js
```diff
@@ -36,7 +36,9 @@ export default class Map {
     }
 
     addLayer(layer, beforeLayerID) {
-        layer.requestData();
+        setTimeout(() => {
```
IagoLast: hacky smell ? 🙄

IagoLast:
I dont see the reason for creating a  BaseWindshaft  "class" specially since the base class is empty...

src/api/map.js
Jesus89: Right :joy:

src/api/util.js
```diff
@@ -2,6 +2,20 @@
  * Export util functions
```
IagoLast: +1 Moving projection functions to the utils file but I think we can seize this opportunity to do a small refactor

src/api/util.js
```diff
@@ -2,6 +2,20 @@
...
  */
 
+const DEG2RAD = Math.PI / 180;
+const EARTH_RADIUS = 6378137;
+const WM_R = EARTH_RADIUS * Math.PI; // Webmercator *radius*: half length Earth's circumference
```
IagoLast: rename to WEBMERCATOR_RADIUS

src/api/util.js
```diff
@@ -2,6 +2,20 @@
...
+const WM_2R = WM_R * 2; // Webmercator coordinate range (Earth's circumference)
+
+// Webmercator projection
+function wmProjection(latLng) {
```
IagoLast: function names should be verbs, maybe `project` ?

src/api/map.js
IagoLast: waiting for a good explanation ¬¬

src/api/source/geojson.js
```diff
@@ -0,0 +1,184 @@
...
+
+
+// TODO: add docs / API
+
+export default class GeoJSON extends Base {
+
+    constructor(data) {
+        super();
+        data = data || {};
+        if (data.type === 'FeatureCollection') {
```
IagoLast: extract method 
```javascript
  _getFeatures(data) {
        switch (data.type) {
        case 'FeatureCollection':
            this._features = data.features || [];
            break;
        case 'Feature':
            this._features = [data];
            break;
        default:
            throw Error('No valid GeoJSON data');
        }
    }
```

src/api/source/geojson.js
```diff
@@ -0,0 +1,184 @@
...
+        this._status = 'init'; // init -> metadata -> data
+        this._type = ''; // Point, LineString, MultiLineString, Polygon, MultiPolygon
+    }
+
+    bindLayer(addDataframe, removeDataframe) {
+        this._addDataframe = addDataframe;
+        this._removeDataframe = removeDataframe;
+    }
+
+    requestData() {
```
IagoLast: this method is very confusing to me, to many responsabilities, sometimes returning a promise, global state...

src/api/source/geojson.js
```diff
@@ -0,0 +1,184 @@
...
+        //
+        if (this._status === 'init') {
+            this._status = 'metadata';
+            return new Promise(resolve => {
+                let metadata = {};
+                resolve(metadata);
+            });
+        } else if (this._status === 'metadata') {
+            this._status = 'data';
+            let geometry = null;
```
IagoLast: we could extract a `decodeGeometry` method here, this method receives a feature list and returns a geometry

src/api/source/geojson.js
```diff
@@ -0,0 +1,184 @@
...
+                    }
+                }
+            }
+            const dataframe = new Dataframe(
+                { x: 0, y: 0 },
+                1,
+                geometry,
+                properties,
+            );
+            dataframe.type = this._getDataframeType(this._type);
```
IagoLast: IMO we should update the dataframe constructor to set those properties instead doing it from outside

src/api/source/geojson.js
```diff
@@ -0,0 +1,184 @@
...
+            dataframe.active = true;
+            dataframe.size = numFeatures;
+            this._addDataframe(dataframe);
+        }
+    }
+
+    _getDataframeType(type) {
+        if (type === 'Point') return 'point';
```
IagoLast: 🌈  I like how the switch statement looks for this kind of things

src/api/source/geojson.js
```diff
@@ -0,0 +1,184 @@
...
+        if (type === 'LineString' || type === 'MultiLineString') return 'line';
+        if (type === 'Polygon' || type === 'MultiPolygon') return 'polygon';
+        return '';
+    }
+
+    _computePointGeometry(data) {
```
IagoLast: rename data to `latlng` or `point`

test/integration/util.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+}
+
+function loadOptions() {
+    return {
+        delay: 100,
+        viewportWidth: 400,
+        viewportHeight: 300,
+        headless: process.platform === 'linux'
+        // waitForFn: () => window.mapLoaded
```
IagoLast: cannot we use this event with geojson maps?

test/integration/util.js
```diff
@@ -0,0 +1,68 @@
...
+        file: file,
+        geojson: geojson,
+        cartogl: path.resolve(mainDir, 'dist', 'carto-gl.js'),
+        mapboxgl: path.resolve(mainDir, 'vendor', 'mapbox-gl-dev.js'),
+        mapboxglcss: path.resolve(mainDir, 'vendor', 'mapbox-gl-dev.css')
+    }));
+}
+
+module.exports = {
+    getHTML: getHTML,
```
IagoLast: with es6 you dont need to duplicate key:value if they have the same identifier

IagoLast:
great job @Jesus89 💪 

 does it make sense to you to add unit test for the `geojson` source?

src/api/source/geojson.js
Jesus89: This is the only way with the current linter:
```js
switch (type) {
case 'Point':
    return 'point';
case 'LineString':
case 'MultiLineString':
    return 'line';
case 'Polygon':
case 'MultiPolygon':
    return 'polygon';
default:
    return '';
}
```

src/api/source/geojson.js
Jesus89: I prefer to keep data, because is not a point (`{x,y}`) it is an "encoded" array (`[0, 0]`)

src/api/util.js
Jesus89: `projectWM`, `wmProject`?

test/integration/util.js
Jesus89: It is no needed here, I added just for testing.

test/integration/util.js
Jesus89: True that

src/api/map.js
Jesus89: Now the `layer.setStyle` is an async function so in this case when `addLayer` calls to the `layer.initCallback` function the style doesn't exist at that time and the function crashes. I'm open to suggestions.

src/api/util.js
Jesus89: what kind of refactor?

src/api/source/geojson.js
Jesus89: Totally agree. That's why I added the comment to refactor it (for a future PR)

Jesus89:
Yes. I want to add unit testing and proper API documentation (and MultiPoints support) but in other PR. This is just the basic support for the Integrationt tests

.circleci/config.yml
```diff
@@ -56,11 +56,31 @@ jobs:
             - node_modules
           key: v1-dependencies-{{ checksum "package.json" }}
       - run: xvfb-run -a yarn test:acceptance
-
+  test-render:
+    docker:
+      # specify the version you desire here
+      - image: circleci/node:8-browsers
```
davidmanzanares: Just curious, why this wasn't before?

src/api/integrator/mapbox-gl.js
```diff
@@ -81,17 +77,8 @@ class MGLIntegrator {
         var c = this.map.getCenter();
         var nw = b.getNorthWest();
         var sw = b.getSouthWest();
-        var z = (Wmxy(nw).y - Wmxy(sw).y) / WM_2R;
-        this.renderer.setCenter(c.lng / 180., Wmxy(c).y / WM_R);
+        var z = (util.wmProjection(nw).y - util.wmProjection(sw).y) / util.WM_2R;
```
davidmanzanares: :+1: for moving it to `util`, but can we rename it to `projectToWebMercator` or something without the `wm`

PR #143, opened 2018-03-09
Doc ramp expression

IagoLast:
- Add buckets doc
- Add ramp doc
- Prepare colors doc


src/core/style/expressions/buckets.js
```diff
@@ -4,10 +4,59 @@ import { implicitCast } from './utils';
...
-    /*
-        If input is numeric => args is a breakpoint list
-        If input is categorical => args is a list of category names to map input
-    */
+
+    /**
+     * Given a property create "sub-groups" based on the given breakpoints.
+     * 
+     * 
+     * Imagine a trafic dataset with a speed property. We want to divide the roads in
```
Jesus89: traffic

src/core/style/expressions/buckets.js
```diff
@@ -4,10 +4,59 @@ import { implicitCast } from './utils';
...
+     *  const $procesedSpeed = s.prop('procesedSpeed');
+     *  const style = new carto.Style({
+     *  color: s.ramp(
+     *      s.buckets($procesedSpeed, 'slow', 'medium', 'high'),
+     *      s.palettes.PRISM),
+     * });
+     * ```
+     * 
+     * @param {carto.style.expressions.property} property - The property to be evaluated and interpolated
+     * @param {...carto.style.expressions.expression} breakpoints - Numeric expression containing the different breakpoints.
```
Jesus89: Is this `carto.style.expressions.property` or `carto.style.expressions.prop`.
Why the 3 dots?

IagoLast: both are valid, `s.prop` is an alias for `s.property`.

... dots means you can pass n arguments

src/core/style/expressions/palettes.js
```diff
@@ -1,5 +1,31 @@
...
+ * 
+ * Palettes are constants that allow to use {@link https://carto.com/carto-colors/|cartocolors} easily.
+ * Use them with a {@link carto.style.expressions.ramp|ramp}
+ * 
+ * The following palettes are availiable:
+ *  - Categorical:
+ *      - PRISM
+ *      - EARTH
+ *  - Numeric
+ *      - ...
```
Jesus89: Where are the numeric palettes?

IagoLast: [here](https://github.com/CartoDB/CartoColor/blob/master/cartocolor.js) tagged as cuantitative, names are going to change so I prefer to wait for a decision before copying the full list

davidmanzanares: Please, change this, we may add true numeric palettes in the future (list of numbers for bubble maps), and there are multiple "numeric" palette types I think

PR #145, opened 2018-03-09
GeoJSON doc + utests

Jesus89:
- Added API doc
- Added unit tests

### DIFF
--- a/src/api/source/geojson.js
+++ b/src/api/source/geojson.js
@@ -4,10 +4,32 @@ import * as rsys from '../../client/rsys';
 import * as util from '../util';
 
 
-// TODO: add docs / API
-
 export default class GeoJSON extends Base {
 
+    /**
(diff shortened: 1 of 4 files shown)


src/api/source/geojson.js
```diff
@@ -21,43 +46,59 @@ export default class GeoJSON extends Base {
...
             return [data];
         }
         else {
-            throw Error('No valid GeoJSON data');
+            throw new CartoValidationError('source', 'nonValidGeoJSONData');
         }
     }
 
+    _requestMetadata() {
+        return new Promise(resolve => {
```
IagoLast:
```javascript 
return Promise.resolve({});
```

?

Jesus89: This is a TODO. We will add metadata extraction in the future.

PR #147, opened 2018-03-12
Fix multilayer

davidmanzanares:
Fix for https://github.com/CartoDB/renderer-prototype/issues/146

The main problem was that in `mapbox-gl.js` only one layer initCallback was being called.

However, there was another problem, expressions with multiple parents are not supported, but `style.js` was sharing expressions between styles, this caused some WebGL warnings and it was going to explode sooner or later, so I fixed it.

### DIFF
--- a/src/api/integrator/mapbox-gl.js
+++ b/src/api/integrator/mapbox-gl.js
@@ -47,7 +47,7 @@ class MGLIntegrator {
                 this.invalidateWebGLState = invalidate;
                 this.notifyObservers();
                 this.renderer._initGL(gl);
-                layer.initCallback();
+                this._layers.map(layer => layer.initCallback());
             }
             layer.paintCallback();
             invalidate();
@@ -56,7 +56,7 @@ class MGLIntegrator {
             id: layerId,
             type: 'custom-webgl'
         }, beforeLayerID);
-        this._layers.push(layerId);
+        this._layers.push(layer);
         this.move();
     }
     needRefresh() {
--- a/src/api/style.js
+++ b/src/api/style.js
@@ -8,13 +8,13 @@ import Expression from '../core/style/expressions/expression';
 import CartoValidationError from './error-handling/carto-validation-error';
 
 
-const DEFAULT_RESOLUTION = 1;
-const DEFAULT_COLOR_EXPRESSION = s.rgba(0, 1, 0, 0.5);
-const DEFAULT_WIDTH_EXPRESSION = s.float(5);
-const DEFAULT_STROKE_COLOR_EXPRESSION = s.rgba(0, 1, 0, 0.5);
-const DEFAULT_STROKE_WIDTH_EXPRESSION = s.float(0);
-const DEFAULT_ORDER_EXPRESSION = s.noOrder();
-const DEFAULT_FILTER_EXPRESSION = s.floatConstant(1);
+const DEFAULT_RESOLUTION = () => 1;
+const DEFAULT_COLOR_EXPRESSION = () => s.rgba(0, 1, 0, 0.5);
+const DEFAULT_WIDTH_EXPRESSION = () => s.float(5);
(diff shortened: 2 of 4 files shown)


src/core/shaders/index.js
```diff
@@ -6,6 +6,8 @@ import ShaderCache from './shader-cache';
 const NUM_TEXTURE_LOCATIONS = 4;
 const shaderCache = new ShaderCache();
 
+let programID = 1;
```
Jesus89: This is not used

src/core/shaders/index.js
```diff
@@ -34,6 +36,7 @@ function compileProgram(gl, glslVS, glslFS) {
     if (!gl.getProgramParameter(this.program, gl.LINK_STATUS)) {
         throw new Error('Unable to link the shader program: ' + gl.getProgramInfoLog(this.program));
     }
+    this.programID = programID++;
```
Jesus89: This is not used

src/api/style.js
```diff
@@ -8,13 +8,13 @@ import Expression from '../core/style/expressions/expression';
...
 
 
-const DEFAULT_RESOLUTION = 1;
-const DEFAULT_COLOR_EXPRESSION = s.rgba(0, 1, 0, 0.5);
-const DEFAULT_WIDTH_EXPRESSION = s.float(5);
-const DEFAULT_STROKE_COLOR_EXPRESSION = s.rgba(0, 1, 0, 0.5);
-const DEFAULT_STROKE_WIDTH_EXPRESSION = s.float(0);
-const DEFAULT_ORDER_EXPRESSION = s.noOrder();
-const DEFAULT_FILTER_EXPRESSION = s.floatConstant(1);
+const DEFAULT_RESOLUTION = () => 1;
```
Jesus89: Why are this constants turned to functions?

src/core/shaders/index.js
davidmanzanares: "Improve shader program debugging", WebGL shaders don't have even an ID, it is a nightmare for debugging purposed if we don't add some IDs.

src/api/style.js
davidmanzanares: "However, there was another problem, expressions with multiple parents are not supported, but style.js was sharing expressions between styles, this caused some WebGL warnings and it was going to explode sooner or later, so I fixed it." we need to create a new expression each time we need them

src/api/integrator/mapbox-gl.js
```diff
@@ -47,7 +47,7 @@ class MGLIntegrator {
                 this.invalidateWebGLState = invalidate;
                 this.notifyObservers();
                 this.renderer._initGL(gl);
-                layer.initCallback();
+                this._layers.map(layer => layer.initCallback());
```
IagoLast: map --> foreach

IagoLast:
I suggest creating a `constants.js` file at  the `style.js` level.

this file can define [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) exposing the public constants.

```javascript
let  constants = {
  get FOO()  { console.log('executing internal function') ; return 'foo'; }
}

// Executes the function internally and returns a new value
constants.FOO;


// Instead using functions (misleading IMO)
contants.FOO()
```

this way we can still use constants as properties instead functions
```javascript 
import { DEFAULT_RESOLUTION} from './constants';
```

src/core/shaders/index.js
IagoLast: what about a comment?

`// programID is used to improve shader debugging`

davidmanzanares:
Seriously, guys, this is a simple bugfix. Let's focus on what it is important and don't waste 3 developers on this kind of thing.

Jesus89:
Normally bug fixes come with missing tests. Please add some tests to make sure that the same bug won't appear again.
- e2e or render test for multi layer
- unit testing for defaults in style

PR #149, opened 2018-03-12
Fix examples, improve their code style and use modern styling features

davidmanzanares:
- Fix all broken examples
- Improve their code style (auto formatting)
- Use modern styling features and modern function aliases.


IagoLast:
I think we should write the expressions multi-line to increase legibilty

davidmanzanares:
I did that on some of the longest examples, I didn't do that in the ships example since it should be transformed into a `torque` expression anyway, but that requires https://github.com/CartoDB/renderer-prototype/pull/135

Jesus89:
As you can see in the Kanban I'm working on the examples right now so in order to avoid conflicts **merge this changes** into master and I will update my branch from it.

I'm with @IagoLast in increasing the legibility of the expressions. I would use by default string expressions and only the interactive methods when required.

PR #150, opened 2018-03-13
Docs carto namespace

IagoLast:
- Document carto namespace
- Improve carto.source.Dataset docs
- Improve carto.source.SQL docs

### DIFF
--- a/src/api/source/dataset.js
+++ b/src/api/source/dataset.js
@@ -6,7 +6,19 @@ import CartoValidationError from '../error-handling/carto-validation-error';
 export default class Dataset extends BaseWindshaft {
 
     /**
-     * Create a carto.source.Dataset.
+     * A dataset defines the data that will be displayed in a layer and is equivalent
+     * to a table in the server.
+     * 
+     * If you have a table named `european_cities` in your CARTO account you could load all the
+     * data in a layer using a `carto.Dataset`. 
+     * 
+     * If you want to load data applying a SQL query see {@link carto.source.SQL|carto.source.SQL}.
+     * 
+     * Since tables in the server are protected you must provide valid credentials in order to get access to the data.
+     * This can be done {@link carto.setDefaultAuth|setting the default auth} in the carto object or providing an `auth` 
+     * object with your username and apiKey.
+     * 
+     * If your server is not hosted by CARTO you must add a third parameter that includes the serverURL.
      *
      * @param {string} tableName - The name of an existing table
      * @param {object} auth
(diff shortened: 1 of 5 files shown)


src/api/source/dataset.js
```diff
@@ -6,7 +6,19 @@ import CartoValidationError from '../error-handling/carto-validation-error';
 export default class Dataset extends BaseWindshaft {
 
     /**
-     * Create a carto.source.Dataset.
+     * A dataset defines the data that will be displayed in a layer and is equivalent
+     * to a table in the server.
+     * 
+     * If you have a table named `european_cities` in your CARTO account you could load all the
+     * data in a layer using a `carto.Dataset`.
```
Jesus89: carto.source.Dataset

src/api/style.js
```diff
@@ -295,11 +295,11 @@ export default class Style {
...
-         * @property {carto.style.expression.Base} color
-         * @property {carto.style.expression.Base} width
-         * @property {carto.style.expression.Base} strokeColor
-         * @property {carto.style.expression.Base} strokeWidth
-         * @property {carto.style.expression.Base} order
+         * @property {carto.style.expressions.Expression} color
+         * @property {carto.style.expressions.Expression} width
+         * @property {carto.style.expressions.Expression} strokeColor
+         * @property {carto.style.expressions.Expression} strokeWidth
+         * @property {carto.style.expressions.Expression} order
```
Jesus89: This is redundant. We do not have carto.source.Source for the Base class, we use carto.source.Base instead. IMO we should do the same here with the expressions.

IagoLast: I think this is out of the scope of this PR, just changed an inexistent documented object for the real one ;) 

In any case Im totally against emulating java in javascript including fake abstract "clasess". We can discuss this in the future

src/core/style/expressions/expression.js
```diff
@@ -3,7 +3,11 @@ import { blend, animate } from '../functions';
 import * as schema from '../../schema';
 
 /**
- * @description Abstract expression class
+ * Abstract expression class
+ * All expressions listed in  {@link carto.style.expressions} inherit from this class so any of them
+ * they can be used where an Expression is required as long as the types match. 
+ * 
+ * This means that you can't a numeric expression where a color expression is expected.
```
davidmanzanares: you can't  **use**  a numeric expression where a color expression is expected.

src/api/style.js
davidmanzanares: I think the problem is just the naming of the `Expression` class, in any case, it is outside the PR scope

PR #152, opened 2018-03-13
Linear fix

davidmanzanares:
The fix is trivial.

The test should work in theory, but since GeoJSON doesn't support properties, it just passes with a black image.


IagoLast:
I'd only add the hotfix without the test or I would have implemented the tests. Not sure about the amount of work that takes to improve the framework to fix this test case so do as you consider.

davidmanzanares:
I reverted the test commit, I'll add them in the future when the testing framework is more capable.

PR #153, opened 2018-03-13
Examples

Jesus89:
Related to: https://github.com/CartoDB/renderer-prototype/issues/47

- Add examples structure by categories
- Update and move previous examples
- Add index.html for the examples
- Add global index.html
- Add examples.json files for `Developer center` doc generation
- Add `yarn serve` command


examples/basics/change-source.html
```diff
@@ -0,0 +1,122 @@
...
+      apiKey: 'YOUR_API_KEY'
+    });
+
+    const source = new carto.source.SQL(`
+      SELECT *
+        FROM ne_10m_populated_places_simple
+    `);
+    const style = new carto.Style(`
+      width: 6
+      color: rgba(0.93,0.30,0.35,1.0)
```
IagoLast: shall we add spaces after commas?

examples/basics/change-source.html
```diff
@@ -0,0 +1,122 @@
...
+      strokeWidth: 1
+      strokeColor: rgba(1,1,1,1)
+    `);
+    const layer = new carto.Layer('layer', source, style);
+
+    layer.addTo(map, 'watername_ocean');
+
+    function setAllCities() {
+      layer.setSource(new carto.source.SQL(`
```
IagoLast: instead creating a new instance everytime we could use 3 global constants and set them as sources

examples/basics/change-style.html
```diff
@@ -0,0 +1,125 @@
...
+
+    carto.setDefaultAuth({
+      user: 'cartogl',
+      apiKey: 'YOUR_API_KEY'
+    });
+
+    const source = new carto.source.Dataset(`
+      ne_10m_populated_places_simple
+    `);
+    const style = new carto.Style(`
```
IagoLast: same as in setSource, instead creating new instances I think that reusing 3 objects will be better

examples/examples.json
```diff
@@ -0,0 +1,188 @@
...
+        }
+    },
+    "categories": [  
+        {  
+            "title": "Basics",
+            "samples": [  
+                {  
+                    "title": "Add a layer",
+                    "desc": "Add one CARTO layer to your map",
+                    "file": {
```
IagoLast: what is the purpose of the file field ?

examples/expressions/animation.html
```diff
@@ -0,0 +1,54 @@
...
+      user: 'cartogl',
+      apiKey: 'YOUR_API_KEY'
+    });
+
+    const source = new carto.source.Dataset(`
+      wwi
+    `);
+    const s = carto.style.expressions;
+    const style = new carto.Style({
+        width: s.mul(s.blend(1, 2, s.near(s.prop('day'), s.mod(s.mul(25, s.now()), 1000), 0, 10), s.cubic), s.zoom()),
```
IagoLast: I think we should split this in multiple lines

examples/expressions/filter.html
```diff
@@ -0,0 +1,55 @@
...
+  <script>
+    const map = new mapboxgl.Map({
+      container: 'map',
+      style: 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json',
+      center: [0, 30],
+      zoom: 2
+    });
+
+    carto.setDefaultAuth({
+      user: 'mamataakella',
```
IagoLast: we should use cartogl for every example

examples/expressions/filter.html
```diff
@@ -0,0 +1,55 @@
...
+      apiKey: 'YOUR_API_KEY'
+    });
+
+    const source = new carto.source.Dataset(`
+      pop_density_points
+    `);
+    const s = carto.style.expressions;
+    const avg = s.viewportAvg(s.prop('dn'));
```
IagoLast: I remove the viewport to reduce noise over the example

examples/expressions/order.html
```diff
@@ -0,0 +1,140 @@
...
+              s.mul(s.add(s.div(s.zoom(), 4000), 0.01), 1.5)
+          ),
+          color: s.ramp(s.prop('category'), s.palettes.prism),
+          strokeColor: s.rgba(0, 0, 0, 0.7),
+          strokeWidth: s.div(s.zoom(), 10000),
+      }));
+    }
+
+    function setAscOrder() {
+      layer.blendToStyle(new carto.Style({
```
IagoLast: If we want to show the meaning of the `order` property I think we should try to reduce the noise as much as we can

(with "noise" I mean use few expressions as possible)

examples/index.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Examples | CARTO</title>
+  <script src="https://code.jquery.com/jquery-3.2.1.min.js"></script>
```
IagoLast: use `fetch` instead

examples/expressions/animation.html
davidmanzanares: This should be replaced by torque actually, I can do it on another PR

examples/basics/change-source.html
Jesus89: Yep, but this should be changed to HEX color when available.

examples/basics/change-style.html
Jesus89: This could be done but I don't know if it could cause trouble with the current implementation...

examples/examples.json
Jesus89: In CARTO.js whe use this field to point to the 'leaflet' and 'gmaps' examples. Now for vector we use mapbox but other providers could be added in the future.

examples/expressions/filter.html
Jesus89: Yes, I have not changed some examples like this one but it is better to have simpler examples

examples/expressions/order.html
Jesus89: Same

IagoLast: same¿

examples/examples.json
```diff
@@ -0,0 +1,188 @@
...
+            "samples": [  
+                {  
+                    "title": "Add a layer",
+                    "desc": "Add one CARTO layer to your map",
+                    "file": {  
+                        "mapbox": "basics/single-layer.html"
+                    }
+                },
+                {  
+                    "title": "Add more layers",
```
IagoLast: `Add multiple layers`

examples/misc/edit-sql-expressions.html
```diff
@@ -0,0 +1,178 @@
...
+      <div class="control-inner">
+        <h3>Expressions</h3>
+        <textarea id="expressions" rows="3">
+width: 6
+color: rgba(0.93,0.30,0.35,1.0)
+strokeWidth: 0.5
+strokeColor: rgba(1,1,1,1)
+          </textarea>
+          <div class="action-container">
+            <button class="button-update" onclick="updateStyle()">
```
IagoLast: I think a `blend  style` button would be nice so the users can see the diference between a [blend](https://cartodb.github.io/renderer-prototype/docs/public/carto.Layer.html#blendToStyle) and a [update](https://cartodb.github.io/renderer-prototype/docs/public/carto.Layer.html#setStyle).

examples/misc/simple-map.html
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Simple map | CARTO</title>
```
IagoLast: I think the title is not very descriptive.

maybe `no basemap` ? 🤔

davidmanzanares: I'd say `standalone`

Jesus89: This is a custom carto.Map used now for the tests. May be we shouldn't include it now in the examples.

examples/expressions/filter.html
Jesus89: I need `pop_density_points` dataset

examples/expressions/order.html
Jesus89: Same as above. Some expression examples require clean.

Jesus89:
NOTE: some expression examples require a little cleaning (extra expressions noise)

PR #157, opened 2018-03-13
Geojson properties and metadata

davidmanzanares:
Fix https://github.com/CartoDB/renderer-prototype/issues/156

It adds (apart from GeoJSON unit tests) a render `linear` test which was developed in other branch but didn't work since the GeoJSON properties feature wasn't there.


src/api/source/geojson.js
```diff
@@ -79,14 +85,13 @@ export default class GeoJSON extends Base {
 
     _requestMetadata() {
         return new Promise(resolve => {
```
IagoLast:
```javascript
return Promise.resolve(this._metadata);
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

davidmanzanares: :rocket: Fixed

PR #159, opened 2018-03-13
Improve testing framework

Jesus89:
Related to: https://github.com/CartoDB/renderer-prototype/issues/158

- Move testing GeoJSON files to `test/common/sources`
- Implement `f-` and `x-` detection for test's folders
- Update testing documentation
- Remove acceptance tests from ci


test/common/util.js
```diff
@@ -6,10 +6,23 @@ const exquisite = require('exquisite-sst');
...
 
 function loadFiles(directory) {
     testsDir = directory;
-    return glob.sync(path.join(directory, '**', testFile));
+    let files = [];
+    let fFiles = [];
+    const allFiles = glob.sync(path.join(directory, '**', testFile));
+    allFiles.forEach(function (file) {
+        const name = getName(file);
+        if (name.indexOf('/f-') !== -1) {
```
IagoLast: [startsWith](https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Objetos_globales/String/startsWith)

test/common/util.js
```diff
@@ -6,10 +6,23 @@ const exquisite = require('exquisite-sst');
...
+            fFiles.push(file);
+        } else if (name.indexOf('/x-') === -1) {
```
IagoLast: [same](https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Objetos_globales/String/startsWith)

test/common/util.js
IagoLast: or (includes)[https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Objetos_globales/String/includes]

test/integration/render/collection-point/strokeColor/scenario.js
```diff
@@ -1,11 +1,11 @@
-/* global carto geojson */
+/* global carto sources */
```
IagoLast: instead duplicating this line in every file we can add a .eslint.js file at top level

test/common/util.js
Jesus89: startsWith does not work. includes is a better option ;)

PR #160, opened 2018-03-13
Suggestions for map examples

makella:
@Jesus89 

This is a PR for some adjustments to your example maps.

Description by file below. Some of them I couldn't figure out exactly how to modify given the example you were showing so you will find notes about those below too.

**`ramp-category.html`**:
+ I changed the color scheme to `vivid` for better contrast to demonstrate the different options
+ For the category `buckets` I changed the category to `Moda y clazado` because that is the most common category attribute and easier to see vs. `Salud`
+ I removed the `quantiles` example because we shouldn't use a classification method for categorical data

**`ramp-numeric.html`**
+ I removed `ramp auto` because it doesn't really do anything and probably not something that we want to encourage people to do
+ I added `ramp quantiles` to this example because it is appropriate for the numeric attribute
  + **note**: it isn't working I'm not sure why is it the attribute, a bug,... ?
+ I changed the color scheme to `sunset` so visually, you can see the differences better

**`order.html`**
*hopefully I got the syntax right :)*
+ I adjusted the marker `strokeColor` so it isn't so overpowering
+ I modified the `width` styling to be less complex (if the purpose was to show complexity, change back, but I figured it was to show ordering of symbol sizes more than complex styling expressions)
+ Similarly, I simplified the `strokeWidth`

There are other modifications that we should make that I wasn't able to: 

**`filter.html`**: 
+ we shouldn't display the color with `prism` it should be a sequential color scheme. [I tried to recreate that map](https://cartodb.github.io/renderer-prototype/example/mapbox.html#eyJhIjoicG9wX2RlbnNpdHlfcG9pbnRzIiwiYiI6IiIsImMiOiJtYW1hdGFha2VsbGEiLCJkIjoiaHR0cHM6Ly97dXNlcn0uY2FydG8uY29tIiwiZSI6IndpZHRoOiB6b29tKClcbmNvbG9yOiByYW1wKGJ1Y2tldHModmlld3BvcnRBVkcoJGRuKSksIHByaXNtKVxuZmlsdGVyOiBiZXR3ZWVuKCgkZG4pLDYwLDcwKVxuXG5cblxuIiwiZiI6eyJsbmciOi0xMTkuNjQwNDM1Mjg2NjIwNDYsImxhdCI6Ny4wMTM2ODE5NjYxNTA5NTV9LCJnIjowfQ==) with the editor, but didn't have luck so I couldn't adjust the styling you have.

**`viewport.html`**:
+ I would try a sequential color scheme vs. the diverging `earth`. diverging schemes are best to show above and below an average or standard deviation above or below... with quantiles, 4 classes, I'm not sure that you will see the classes that you should.
+ I'm not exactly sure what you are doing with the point `width` but the result is looking inconsistent through zoom

### DIFF
--- a/examples/expressions/filter.html
+++ b/examples/expressions/filter.html
@@ -44,7 +44,7 @@
     const avg = s.viewportAvg(s.prop('dn'));
     const style = new carto.Style({
         color: s.ramp(s.buckets(s.prop('dn'), avg), s.palettes.prism),
-        filter: s.between(s.property('dn'), 60, 70)
+        filter: s.between(s.prop('dn'), 60, 70)
 
     });
     const layer = new carto.Layer('layer', source, style);
--- a/examples/expressions/order.html
+++ b/examples/expressions/order.html
@@ -86,13 +86,10 @@
     `);
     const s = carto.style.expressions;
     const style = new carto.Style({
-        width: s.mul(
-            s.mul(s.sqrt(s.div(s.prop('amount'), 5000)), 20),
-            s.mul(s.add(s.div(s.zoom(), 4000), 0.01), 1.5)
-        ),
+        width: s.sqrt('amount'),
         color: s.ramp(s.prop('category'), s.palettes.prism),
-        strokeColor: s.rgba(0, 0, 0, 0.7),
-        strokeWidth: s.div(s.zoom(), 10000),
+        strokeColor: s.rgba(1, 1, 1, 0.7),
+        strokeWidth: 1,
     });
     const layer = new carto.Layer('layer', source, style);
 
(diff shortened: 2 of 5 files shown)


Jesus89:
Thanks @makella! :smile: 

This is awesome! Feel free to add and improve whatever you want in the examples and docs.

> ramp-category.html

:+1: 

> ramp-numeric.html

I just found a bug in the `ramp-numeric` example: calling `setRampAuto` which should be `setRampQuantiles`. I can fix it.

> order.html

:+1: 

> filter.html

I don't get what's the issue here. May be we can change the whole example to focus on the filtering expression. What do you think?

> viewport.html

I just adapted this example from the old `viewPercentage.html` example. May be @davidmanzanares could add more light here, but I think the zoom behavior could be related to a bug in the Windshaft server.

PR #161, opened 2018-03-14
Use exquisite-2.0-preview

IagoLast:
https://github.com/CartoDB/renderer-prototype/issues/151
---
This PR uses 2 new exquisite features:

- Exquisite now returns the number of different pixels instead a single boolean value.
  - Now every failing test tells the number of different pixels
  - We even can create manual thresholds "for this image allow up to 30 different pixels"
- Add a function to capture browser console
  - This way we can detect network errors, exceptions..etc.
  - By default only `warn` `error` is displayed.
  - verbose mode: `VERBOSE_LOG=true yarn test:render`

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "eslint": "^4.15.0",
-    "exquisite-sst": "IagoLast/Exquisite#return-diff-pixels",
+    "exquisite-sst": "IagoLast/Exquisite#master",
     "glob": "^7.1.2",
     "http-server": "^0.11.1",
     "jasmine-core": "^2.99.1",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "eslint": "^4.15.0",
-    "exquisite-sst": "IagoLast/Exquisite#master",
+    "exquisite-sst": "IagoLast/Exquisite#return-diff-pixels",
     "glob": "^7.1.2",
     "http-server": "^0.11.1",
     "jasmine-core": "^2.99.1",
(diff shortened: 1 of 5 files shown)


IagoLast:
I think we must add a `verbose` flag to ignore console logs, for example for `render` tests we get this output:

screen shot 2018-03-14 at 11 55 09

maybe only displaying console.errors?
or showing the full log when the test fail?

Jesus89:
> Exquisite now returns the number of different pixels instead a single boolean value

Perfect! This should be useful to detect if all the image is wrong or just a couple of pixels.

I think adding a threshold for pixels is not useful because we already have a threshold for the pixel matching, but it is good to have the number in any case.

> Add a function to capture browser console

I think we should display only the errors and exceptions in our tests, because if there is a bug it is useful to know where it is. May be the tool should have a verbose option to allow more general cases.

package.json
```diff
@@ -31,7 +31,7 @@
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "eslint": "^4.15.0",
-    "exquisite-sst": "IagoLast/Exquisite#master",
+    "exquisite-sst": "IagoLast/Exquisite#return-diff-pixels",
```
Jesus89: We should we use the master branch (or event a npm version number :D)

test/integration/render.test.js
```diff
@@ -14,6 +14,6 @@ describe('Render tests:', () => {
 function test(file) {
     it(util.getName(file), () => {
         const actual = util.testSST(file, template);
-        return chai.expect(actual).to.eventually.be.true;
+        return chai.expect(actual).to.eventually.be.at.most(1);
```
Jesus89: I would add here a comment saying that this is a temporary threshold to cover the differences in the renderer in Mac/Linux.

package.json
IagoLast: yep but I dont want to merge stuff in master while the final api is not clear, once this gets approved i´ll merge the exquisite PR and here I´ll point to master.

Finally we should use the 2.0 version of exquisite.

Jesus89: The final API is OK for this PR so you can merge it now ;)

PR #164, opened 2018-03-14
new filter example

makella:
@Jesus89 

adjusted the filter example. Basically, we are doing a quantiles classification on a filtered view between `100,150` so what you see on the map are high population density places.


Jesus89:
:tada:

PR #166, opened 2018-03-14
Allow numbers in width and strokeWidth properties

Jesus89:
Related to: https://github.com/CartoDB/renderer-prototype/issues/165

- Allow numbers in width and strokeWidth properties
- Fix expressions/order example

### DIFF
--- a/src/api/style.js
+++ b/src/api/style.js
@@ -5,7 +5,7 @@ import * as shaders from '../core/shaders';
 import { compileShader } from '../core/style/shader-compiler';
 import { parseStyleDefinition } from '../core/style/parser';
 import Expression from '../core/style/expressions/expression';
-import Float from '../core/style/expressions/float';
+import { implicitCast } from '../core/style/expressions/utils';
 import CartoValidationError from './error-handling/carto-validation-error';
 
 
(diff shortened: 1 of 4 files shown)


src/api/style.js
```diff
@@ -311,13 +312,17 @@ export default class Style {
...
             throw new CartoValidationError('style', 'nonValidExpression[color]');
         }
-        if (!(styleSpec.width instanceof Expression)) {
-            throw new CartoValidationError('style', 'nonValidExpression[width]');
-        }
         if (!(styleSpec.strokeColor instanceof Expression)) {
             throw new CartoValidationError('style', 'nonValidExpression[strokeColor]');
         }
-        if (!(styleSpec.strokeWidth instanceof Expression)) {
+        if (util.isNumber(styleSpec.width)) {
```
davidmanzanares: There is a `implicitCast()` function to do exactly this, please, use it.

Jesus89: We need control here of the type to launch the proper exceptions so I can use `implicitCast(` instead of `new float(`, but the rest should be the same.

davidmanzanares: I know, but `implicitCast()` should be the one generating the `float()` and the one checking the `isNumber()`

Jesus89: In order to do that this would be the final code:

```js
if (!(styleSpec.width instanceof Expression)) {
    styleSpec.width = implicitCast(styleSpec.width);
    if (!(styleSpec.width instanceof Expression)) {
        throw new CartoValidationError('style', 'nonValidExpression[width]');
    }
}
```

But this would allow to use strings also for `width` and `strokeWidth`, and we want to avoid that. We need explicit cast instead.

Jesus89: After talking about it we have decided to keep type checking for other PR.

We have also found that the conflictive code in `implicitCast` can be removed:
```js
if (!(value instanceof Expression) && value.type !== 'paletteGenerator' && value.type !== 'float') {
    throw new Error('value cannot be casted');
}
```

So I go ahead with it.

PR #167, opened 2018-03-14
Support local backends

jgoizueta:
I'd like to merge this because it's becoming a pain to maintain these change!

This is to be able to use a local backend (which requires specific ports in the Maps & SQL API URLs).

A local environment can be used by using a `serverURL` like this: `http://{user}.localhost.lan{local}`.
The `{local}` placeholder is to allow insertion of the ports, shouldn't be present in non-local URLs.

I've also completed support of (optional) API keys (which was lacking in some calls to the SQL API


jgoizueta:
cc @Jesus89 @IagoLast don't mean to bother you with boring stuff, but feel free to CR too

Jesus89:
Do we want this changes for the `alpha.1` tag?

jgoizueta:
Sorry for the messy commits guys 😅 I'm not sure what I did, but things look OK now, the only relevant commits are 013e6ef and a7477ea

jgoizueta:
This shouldn't be needed for alpha.1 since local envs or API keys are not going to be used in that release.

But I'll be very happy if this is merged before new conflicts arise 😅

Jesus89:
OK. It looks good to me, btw

src/client/windshaft.js
```diff
@@ -552,12 +554,13 @@ export default class Windshaft {
     }
 
     async getDatesTypes(names, query, conf) {
+        console.log('GETDATAT',conf);
```
IagoLast: do we want this? :P

src/client/windshaft.js
```diff
@@ -611,16 +614,35 @@ function getAggFN(name) {
...
+    return endpoint(conf, `${layergroup.layergroupid}/${layerIndex}/{z}/{x}/{y}.mvt`);
+}
+
+function getSQL(query, conf) {
+    let url = `${conf.sqlServerURL}/api/v2/sql?q=` + encodeURIComponent(query);
+    url = authURL(url, conf);
+    return fetch(url);
+}
+
+function authURL(url, conf) {
```
IagoLast: `addApiKey` would be a better name IMO

src/client/windshaft.js
jgoizueta: No!

src/client/windshaft.js
```diff
@@ -611,16 +614,35 @@ function getAggFN(name) {
...
-        return `${urlTemplates.url}/${conf.username}/api/v1/map/${layergroup.layergroupid}/${layerIndex}/{z}/{x}/{y}.mvt?api_key=${conf.apiKey}`;
+        return authURL(`${urlTemplates.url}/${conf.username}/api/v1/map/${layergroup.layergroupid}/${layerIndex}/{z}/{x}/{y}.mvt`, conf);
+    }
+    return endpoint(conf, `${layergroup.layergroupid}/${layerIndex}/{z}/{x}/{y}.mvt`);
+}
+
+function getSQL(query, conf) {
+    let url = `${conf.sqlServerURL}/api/v2/sql?q=` + encodeURIComponent(query);
+    url = authURL(url, conf);
+    return fetch(url);
```
IagoLast: maybe this function could return the `parsed json` so in the places where its used we get rid of the 

```javascript
const response = await getSql(q, c);
const sql = await response.json();
```

src/api/source/base-windshaft.js
```diff
@@ -39,8 +39,11 @@ export default class BaseWindshaft extends Base {
     _generateURL(auth, config) {
```
IagoLast: maybe we should rename url to urlS since this variable now contains an object with multiple urls inside

PR #169, opened 2018-03-14
changing to points for full support of numeric map types

makella:
@Jesus89 

updated numeric examples with points vs. polygons to support quantiles example


Jesus89:
Perfect!

PR #171, opened 2018-03-14
Separate functions for getting metadata and data

jgoizueta:
I find things clearer this way

### DIFF
--- a/src/api/layer.js
+++ b/src/api/layer.js
@@ -282,16 +282,9 @@ export default class Layer {
     }
     async _styleChanged(style) {
         await this._context;
-        return this.requestMetadata(style).then(metadata => {
-            if (metadata) {
-                // metadata has changed
-                this.metadata = metadata;
-            }
-            this._compileShaders(style, this.metadata);
-            if (metadata) {
-                this.requestData();
-            }
-        });
+        this.metadata = await this.requestMetadata(style);
+        this._compileShaders(style, this.metadata);
+        return this.requestData();
     }
 
     _checkId(id) {
--- a/src/api/layer.js
+++ b/src/api/layer.js
@@ -255,7 +255,7 @@ export default class Layer {
 
     initCallback() {
         this._contextInitCallback();
-        this.requestData();
+        this.requestMetadata();
     }
 
     _addToMGLMap(map, beforeLayerID) {
@@ -282,17 +282,15 @@ export default class Layer {
     }
     async _styleChanged(style) {
         await this._context;
-        const originalPromise = this.requestData(style);
-        if (!originalPromise) {
-            // The previous stored metadata is still valid
+        return this.requestMetadata(style).then(metadata => {
+            if (metadata) {
+                // metadata has changed
+                this.metadata = metadata;
+            }
             this._compileShaders(style, this.metadata);
-            return Promise.resolve();
-        }
-        // this.metadata needs to be updated, try to get new metadata and update this.metadata and proceed if everything works well
-        return originalPromise.then(metadata => {
-            this.metadata = metadata;
-            this._compileShaders(style, metadata);
-            this.requestData(style);
+            if (metadata) {
+                this.requestData();
+            }
         });
     }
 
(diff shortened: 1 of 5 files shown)


src/api/layer.js
```diff
@@ -282,17 +282,15 @@ export default class Layer {
     }
     async _styleChanged(style) {
         await this._context;
-        const originalPromise = this.requestData(style);
-        if (!originalPromise) {
-            // The previous stored metadata is still valid
+        return this.requestMetadata(style).then(metadata => {
```
IagoLast: Since this function is async you can use await

```javascript
 async _styleChanged(style) {
        await this._context;
        const metadata = await this.requestMetadata(style);
        if (metadata) {
            this.metadata = metadata;
            this._compileShaders(style, this.metadata);
            this.requestData();
        }
    }
```

Jesus89:
Thanks! :smile: This makes the code more readable and clear.

I have checked that GeoJSON - MGL connection is broken, I can take a look at it.

src/api/layer.js
Jesus89: I'm not very sure about this, because _styleChanged requires to return a Promise as it is now.

src/api/layer.js
```diff
@@ -333,13 +331,17 @@ export default class Layer {
...
         style = style || this._style;
         if (!style) {
             return;
         }
         await this._context;
-        return this._source.requestData(this._getViewport(), style);
+        return this._source.requestMetadata(style);
+    }
+
+    requestData() {
```
Jesus89: I think you need to wait for the context in this function.

```js
async requestData() {
    await this._context;
    this._source.requestData(this._getViewport());
}
```

Jesus89:
After testing the `async requestData()` I found that `mgl-integration/lines` and `mgl-integration/polygons` are not working because of the opacity. May be it is rendered multiple times.

| Reference | Out |
|---|---|
| reference | reference_out |

src/api/layer.js
Jesus89: This change does the trick:

```js
async requestData() {
    if (!this.metadata) {
        return;
    }
    this._source.requestData(this._getViewport());
}
```

src/api/layer.js
IagoLast:
```javascript
// ... 
return this.requestData();
```

?

davidmanzanares: Yes, I think `        const metadata = await this.requestMetadata(style);` is better.

src/api/layer.js
```diff
@@ -282,17 +282,15 @@ export default class Layer {
     }
     async _styleChanged(style) {
         await this._context;
-        const originalPromise = this.requestData(style);
-        if (!originalPromise) {
-            // The previous stored metadata is still valid
+        return this.requestMetadata(style).then(metadata => {
+            if (metadata) {
```
davidmanzanares: After the refactor I think this `if` is no longer needed

davidmanzanares:
I think there are two things that can be further simplified due to this refactor, apart from that, :rocket:

src/api/layer.js
Jesus89: So with both suggestions the code changes to this one, right?

```js
async _styleChanged(style) {
    await this._context;
    this.metadata = await this.requestMetadata(style);
    this._compileShaders(style, this.metadata);
    return this.requestData();
}
```

davidmanzanares: Maybe the last `return` could be removed, it depends on what kind of semantics do we want. Resolve the promise when the style is known to be valid vs resolve the promise when the style is known to be valid and required data have been downloaded.

I think the first one is better since the second one doesn't mean it has been painted yet, and we already have the `loaded` event.

In any case, the promise is mainly useful for tracking errors (`reject`), not so much for the `resolve`.

src/api/layer.js
davidmanzanares: After seeing with @jgoizueta we've seen that is needed unless we change the geojson source too. This is probably out scope then.

jgoizueta: With a Windshaft source we don't have any actual gain from that *optimization* (avoiding requesting the data if the metadata didn't change), so we could simplify the code by not making the distinction (metadata changed vs didn't). Because our Windshaft source the data already available.

But for other sources the optimization could be interesting, so I'd leave it as is.

jgoizueta:
OK, so finally for the shake of simplicity I've removed the returning of `false` in metadata if it hasn't chenged and I've applied @IagoLast / @Jesus89 refactor of `_styleChanged`

PR #173, opened 2018-03-14
Refactor dataframe

IagoLast:
- Refactor dataframe constructor
- Dataframe.setStyle is defined in the class
- Move geometry decoding logic to its own file

### DIFF
--- a/src/api/source/geojson.js
+++ b/src/api/source/geojson.js
@@ -88,17 +88,15 @@ export default class GeoJSON extends Base {
     }
 
     _requestData() {
-        const geometry = this._decodeGeometry();
-        const properties = this._decodeProperties();
-        const dataframe = new Dataframe(
-            { x: 0, y: 0 },
-            1,
-            geometry,
-            properties,
-        );
-        dataframe.type = this._getDataframeType(this._type);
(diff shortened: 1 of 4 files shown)


davidmanzanares:
the dataframe refactor is bandera

IagoLast:
@davidmanzanares quick question:

As long as I know dataframes are used to bring data to the GPU, this data comes in form of: center, scale, geom, properties, type, and size.

In the `.bind` function the decoder takes those data and generates a new {geometry, breakpoints} data suitable for the GPU.

my question is, instead decoding the data in the `.bind` could we decode the data in the constructor and keep this decoded data cached in every call to .bind?

davidmanzanares:
@IagoLast The problem with that is that you need access to the WebGL context. `bind()` is called right now by the renderer once when the dataframe is added to the dataframe, so, it is only done once.

However, there are some possible optimizations here that could be used in a mid-term future. The decode geometry pass is JS (CPU) heavy, particularly the earcut algorithm. If we find this is a performance problem we can try to solve it by using webworkers, and to do that we'll need to subdivide `bind()`

In any case, I think this is a future problem :smile:

PR #175, opened 2018-03-15
Expressions error control

davidmanzanares:
This adds, improves and test most expression type checking and the type itself of the expressions.

- Fail fast. When a parameter type is already known at constructor time the expression constructor must throw. This cannot be done when the type is a property, for those cases the error will be throw at compile time.
- "Type" fast. To accomplish "fail fast" we need to set the type of a expression as soon as possible, i.e., at constructor time if it doesn't depend on a property.
- Improved error messages. Some helpers have been defined to make consistent and more meaningful error messages. Particularly, the parameter index and name is used to form the error string.
- Testing has been done with unit tests with some helpers (in utils.js) that are shared by all expression tests. They automatically test between property expressions where the type is dynamic and regular expressions with static type.

There were many bugs founds thank these tests too.

However, I'm pretty sure that existing tests could be hardened in some cases, and I left some, not so important, expressions outside this PR.

Having a proper `bool` type is outside the scope of this PR too.


src/core/style/expressions/binary.js
```diff
@@ -343,41 +406,61 @@ function genBinaryOp(jsFn, glsl) {
...
+        return COLORS_TO_COLOR;
+    } else if (a.type == 'category' && b.type == 'category') {
+        return CATEGORIES_TO_FLOAT;
+    } else {
+        return UNSUPPORTED_SIGNATURE;
+    }
+}
+
+function getReturnTypeFromSignature(signature) {
+    if (signature == FLOATS_TO_FLOAT) {
```
Jesus89: `switch` here?

davidmanzanares: Yes, I'll change it

davidmanzanares:
:slightly_smiling_face: I thought much about the OR'ed flags, I haven't seen this pattern in JS, but I dislike the idea of using an array for that :joy:

BTW, I've just made two small fixes.

src/core/style/expressions/binary.js
davidmanzanares: Done

PR #180, opened 2018-03-16
Avoid loading the GeoJSON data multiple times

rochoa:


test/unit/api/source/geojson.test.js
```diff
@@ -54,6 +54,28 @@ describe('api/source/geojson', () => {
...
+                properties: {
+                    cartodb_id: 1
+                }
+            };
+            const source = new GeoJSON(data);
+
+            source.bindLayer(_ => _);
+            spyOn(source, '_addDataframe');
+
+            source.requestData();
```
Jesus89: The test is OK. Just a suggestion. Maybe it is better in this way.

```js
expect(source._addDataframe).toHaveBeenCalledTimes(0);
source.requestData();
expect(source._addDataframe).toHaveBeenCalledTimes(1);
source.requestData();
expect(source._addDataframe).toHaveBeenCalledTimes(1);
```

rochoa: At your command!

davidmanzanares:
"...requestData calls in other place." A small change in the viewport may trigger a tile download for some sources, but it won't trigger anything for geojson sources. The filtering is source dependant.

PR #181, opened 2018-03-16
External race condition mitigation and atomic replacement

davidmanzanares:
This PR adds:
 - Mitigation for race conditions triggered by concurrent calls to layer.setStyle() and layer.setSource().
   Multiple, close in time, calls to these functions generate race conditions as seen in examples/basics/change-source.html
 - Add support for atomicically change the source and the style. This makes the API easier when both things need to change, less error-prone, safer, and more performant.

### DIFF
--- a/src/api/layer.js
+++ b/src/api/layer.js
@@ -92,6 +92,36 @@ export default class Layer {
         this._listeners[eventType].splice(index, 1);
     }
 
+    async setAtomicSourceStyle(source, style) {
+        this._checkSource(source);
+        this._checkStyle(style);
+        this._atomicChangeUID = this._atomicChangeUID + 1 || 1;
+        const uid = this._atomicChangeUID;
+        await this._context;
+        const metadata = await source.requestMetadata(style);
+        if (this._atomicChangeUID > uid) {
+            throw new Error('Another atomic change was done before this one committed');
+        }
+
+        // Everything was ok => commit changes
+        this.metadata = metadata;
+
+        source.bindLayer(this._onDataframeAdded.bind(this), this._onDataFrameRemoved.bind(this), this._onDataLoaded.bind(this));
+        if (this._source && this._source !== source) {
+            this._source.free();
+        }
+        this._source = source;
+        this.requestData();
+
+        if (this._style) {
+            this._style.onChange(null);
+        }
+        this._style = style;
+        style.onChange(() => {
+            this._styleChanged(style);
+        });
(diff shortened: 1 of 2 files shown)


src/api/layer.js
```diff
@@ -100,13 +130,24 @@ export default class Layer {
...
      */
-    setSource(source) {
+    async setSource(source) {
         this._checkSource(source);
+        const style = this._style;
+        if (style) {
+            var metadata = await source.requestMetadata(style);
+        }
+        if (this._style !== style) {
+            throw new Error('A style change was made before the metadata was retrieved, therefore, metadata is stale and it cannot be longer consumed');
```
rochoa: This proves that this method is dangerous and we must want to just have one method for updating the `source` and the `style`.

After talking it offline with @davidmanzanares, the proposal here is to remove this method and have an `update(source, style)` one. That method will be the same as the `setAtomicSourceStyle` in here, but it will allow you to just pass the `source` and optionally the `style`, the current `style` will be re-used if not passed, but we will keep the operation atomic avoiding this problem.

@IagoLast and @Jesus89, your thoughts?

Jesus89: - We can use this API but better with an object because this forces you to pass always the same source if you want to update only the style.
```js
layer.update({
  source: ...,
  style: ...
});
```
- What happens with setStyle and blendToStyle?

IagoLast: In carto.js we got the same problem, a style update could be fired before a second one was processed. 

So we made [those methods](https://carto.com/documentation/cartojs/docs/#cartostylecartocsssetcontent) async making the user responsible for the concurrency, so if you want to perform several operations one after another you have to wait for a promise first.

```javascript
await layer.setStyle(style0);
await layer.setSource(source0);
await layer.setStyle(style2);
```

rochoa: I'm OK with Iago's proposal, but sometimes you cannot set the style and later set the source. Imagine that you go from a points source to a polygons one. So we still need a method to update both _at the same time_.

About
> ```js
> layer.update({
>  source: ...,
>  style: ...
>});
>```

The `setStyle` would be still present, so you would use `update` when you need to update both or just the source. I don't like to have `options` API in these cases.

So, I guess our best approach is to keep `setSource` and `setStyle`, and introduce that new `setAtomicSourceStyle`, but I propose to rename it to `update`.

davidmanzanares: Renamed

PR #182, opened 2018-03-16
Fix lines rendering

Jesus89:
- Fix normal in shader: apply vertexScale to normals
- Fix how corners are computed: using u, v vectors instead of normals
- Update references for line tests.

| Prev | Fixed |
|---|---|
| line-normals-corner-bug | line-fix |
| line-normals-bug | line-normals-fix |


src/core/decoder.js
```diff
@@ -139,10 +118,19 @@ function getLineNormal(a, b) {
     return normalize([-dy, dx]);
 }
 
+function getCornerVector(a, b, c) {
```
davidmanzanares: This returns the normal for all cases where the line segment is not the beginning nor the end of the line string, right?

If that is the case I would change the name, `cornerVector` sounds to me like beginning/end. I would say something like `getJointNormal`.

src/core/decoder.js
```diff
@@ -139,10 +118,19 @@ function getLineNormal(a, b) {
...
+    const u = normalize([a[0] - b[0], a[1] - b[1]]);
+    const v = normalize([c[0] - b[0], c[1] - b[1]]);
+    const sin = - u[1] * v[0] + u[0] * v[1];
+    if (sin !== 0) {
+        return [(u[0] + v[0]) / sin, (u[1] + v[1]) / sin];
+    }
```
davidmanzanares: I think it would be better (and equivalent) to return here the normal instead of relying on the caller for managing the corner-case, which I guess it's that `B` it's in the line `AC`, therefore, the angle between `AB` and `CB` is zero. (The returned value should be the normal of `AC`.

src/core/decoder.js
```diff
@@ -139,10 +118,19 @@ function getLineNormal(a, b) {
     return normalize([-dy, dx]);
 }
 
+function getCornerVector(a, b, c) {
+    const u = normalize([a[0] - b[0], a[1] - b[1]]);
+    const v = normalize([c[0] - b[0], c[1] - b[1]]);
+    const sin = - u[1] * v[0] + u[0] * v[1];
```
davidmanzanares: Can you explain this line and the return line more, please?

src/core/shaders/renderer/lines.js
```diff
@@ -22,7 +22,7 @@ void main(void) {
     color.rgb *= color.a;
     float size = 64.*texture2D(widthTex, featureID).a;
 
-    vec4 p = vec4(vertexScale*(vertexPosition)+normal*0.001*size-vertexOffset, 0.5, 1.);
+    vec4 p = vec4(vertexScale*(vertexPosition+normal*0.001*size)-vertexOffset, 0.5, 1.);
```
davidmanzanares: :+1: good catch!

src/core/decoder.js
```diff
@@ -59,69 +59,48 @@ function decodePolygon(geometry) {
...
+                    nb = getCornerVector(a, b, next) || nb;
                 }
+
+                // First triangle
+    
                 normals.push(-na[0], -na[1]);
                 normals.push(na[0], na[1]);
                 normals.push(-nb[0], -nb[1]);
+    
+                geometry.push(a[0], a[1]);
```
davidmanzanares: Nice refactor

src/core/decoder.js
Jesus89: `na` and `nb` are initialized with lineNormal. The only condition to be replaced by the jointNormal is when the jointNormal exists (when sin!=0).
I think returning the lineNormal inside the jointNormal function is mixing things because in the singularity case the jointNormal does not exists so it must return a null value.

src/core/decoder.js
```diff
@@ -59,69 +59,48 @@ function decodePolygon(geometry) {
...
                 let nb = normal;
-                //TODO bug, cartesian interpolation is not correct, should use polar coordinates for the interpolation
-                if (prev) {
-                    na = normalize([
-                        normal[0] * 0.5 + nprev[0] * 0.5,
-                        normal[1] * 0.5 + nprev[1] * 0.5,
-                    ]);
+                
+                if (i > 0) {
+                    const prev = [lineString[i + -2], lineString[i + -1]];
```
IagoLast: `"+ -2" === "-2"` ?

Jesus89: Yep, this is legacy. I will update it

IagoLast:
LGTM

I tried to pass the render tests but looks that this wont fix the linux/osx diff

Jesus89:
We will have to live with that 1px threshold in Mac. Btw,

| Linux :smile: | reference 1 |
|---|---|
| Mac :disappointed:  | reference_out 1 |

davidmanzanares:
This example is popping in the branch (not in master/nor gh-pages): https://cartodb.github.io/renderer-prototype/example/mapbox.html#eyJhIjoic2Zfc3RjbGluZXMiLCJiIjoiIiwiYyI6ImRtYW56YW5hcmVzIiwiZCI6Imh0dHBzOi8ve3VzZXJ9LmNhcnRvLmNvbSIsImUiOiJjb2xvcjogcmFtcCgkc3RfdHlwZSwgcHJpc20pIFxud2lkdGg6IDAuNSAiLCJmIjp7ImxuZyI6LTEyMi40Mzg5NDM1MzU3NTg5NCwibGF0IjozNy43NTQzNjI1NDExMTQ5fSwiZyI6MTMuOTUzMjcwODE3MjYzMn0=

I don't know why, although I guess that maybe is the change in the shader, I'll look into it

src/core/decoder.js
davidmanzanares: (this is not outdated)

PR #186, opened 2018-03-19
Refactor style

IagoLast:
https://github.com/CartoDB/renderer-prototype/issues/80
---

Minor refactor in `style.js`

### DIFF
--- a/src/api/layer.js
+++ b/src/api/layer.js
@@ -117,9 +117,7 @@ export default class Layer {
             this._style.onChange(null);
         }
         this._style = style;
-        style.onChange(() => {
-            this._styleChanged(style);
-        });
+        style.onChange(this._styleChanged);
         this._compileShaders(style, metadata);
     }
     /**
--- a/src/api/layer.js
+++ b/src/api/layer.js
@@ -315,12 +315,9 @@ export default class Layer {
     }
 
     _compileShaders(style, metadata) {
-        style.compileColorShader(this._integrator.renderer.gl, metadata);
-        style.compileWidthShader(this._integrator.renderer.gl, metadata);
-        style.compileStrokeColorShader(this._integrator.renderer.gl, metadata);
-        style.compileStrokeWidthShader(this._integrator.renderer.gl, metadata);
-        style.compileFilterShader(this._integrator.renderer.gl, metadata);
+        style.compileShaders(this._integrator.renderer.gl, metadata);
     }
+    
     async _styleChanged(style) {
         await this._context;
         const source = this._source;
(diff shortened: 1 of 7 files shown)


src/api/layer.js
```diff
@@ -117,9 +117,7 @@ export default class Layer {
             this._style.onChange(null);
         }
         this._style = style;
-        style.onChange(() => {
-            this._styleChanged(style);
-        });
+        style.onChange(this._styleChanged);
```
davidmanzanares: This seems like a bug to me, `_styleChanged` receives a style argument

src/api/style.js
```diff
@@ -172,7 +172,7 @@ export default class Style {
 
     _changed() {
         if (this._changeCallback) {
-            this._changeCallback();
+            this._changeCallback(this);
```
davidmanzanares: why?

src/api/style.js
```diff
@@ -222,7 +230,7 @@ export default class Style {
         this.filterShader = r.shader;
     }
 
-    _replaceChild(toReplace, replacer) {
+    replaceChild(toReplace, replacer) {
```
davidmanzanares: I don't think we want to expose this implementation detail (this function)

IagoLast: IMO this is a great example to use `$` prefix since is used outside `style.js` so is not private (`_` ) but is not part of the public API

src/api/style.js
IagoLast: Because [as you said](https://github.com/CartoDB/renderer-prototype/pull/186#discussion_r175479354)  the changeCallback receives a style argument ;)

davidmanzanares: I see, you changed the behavior, previously it wasn't that way

src/api/layer.js
davidmanzanares: Since you modify the `onChange` callback to give back the style the lack of an argument is not a problem.

However, I think this lacks a `bind` to `this`

src/api/style.js
davidmanzanares: Hmu... I'm ok with `$` if we all agree on that to make things consistent. Is `$` ok with you @Jesus89 ?

Jesus89: We are using this convention in carto.js so :+1:

IagoLast: I like the convention too but I think we should address this in another issue since there are lots of wrong underscored/public/private functions

src/api/layer.js
IagoLast: I'm not sure about the need of `this`. We are already binding the scope with the first this (`this`._styleChanged)

src/api/style.js
davidmanzanares: :+1:

PR #187, opened 2018-03-19
Renderlayer

davidmanzanares:
This PR adds a new non-public class `RenderLayer`. This class encapsulates dataframes/tiles of the same layer with the same style.

Previously, it was already expected that "active" dataframes shared the same style. Since that requirement is shared by many features (drawMetadata, ordering, heatmaps, **interactivity**), I think it's better to modify the renderer accordingly.


src/core/renderLayer.js
```diff
@@ -0,0 +1,25 @@
+
+export default class RenderLayer {
+    constructor() {
+        this.dataframes = [];
```
IagoLast: Any special reason to use an array? 
If we use an object we get constant add/delete time instead linear

davidmanzanares: Because "premature optimization is the root of all evil" :slightly_smiling_face: 
If we get to the point where we need O(1) for deleting the reference to a dataframe is that we already have f***ed it up, a lot.

src/core/renderer.js
```diff
@@ -141,58 +140,29 @@ class Renderer {
...
     }
 
-
-    getStyledTiles() {
-        return this.dataframes.filter(tile => tile.style && tile.visible && tile.style.colorShader);
-    }
-
-
-    _computeDrawMetadata() {
+    _computeDrawMetadata(layer) {
```
IagoLast: calling this parameter `layer` could be very confusing

davidmanzanares: core/* should never know about the API side of things, therefore, it can only mean `renderLayer`

src/core/renderer.js
```diff
@@ -141,58 +140,29 @@ class Renderer {
...
+        const tiles = layer.dataframes;
```
IagoLast: I think we should rename tiles to `dataframes`

src/core/renderer.js
IagoLast: yep nice when this knowledge is in your head.
But think about people who is reading the code for the first time in a hurry trying to fix a bug...

src/core/renderer.js
davidmanzanares: Yes, but that's outside this PR scope

IagoLast:
🚀

src/core/renderer.js
davidmanzanares: changed

PR #190, opened 2018-03-19
`And` filter bugfix

davidmanzanares:


test/unit/client/windshaft-filtering.test.js
```diff
@@ -196,6 +196,15 @@ describe('src/client/windshaft-filtering', () => {
                 }))).toBeNull();
             });
 
+            it('with the `filter: $property==5 and $property==5`', () => {
```
IagoLast: Should return null when ..

PR #191, opened 2018-03-19
Add a puppeteer-based benchmark

davidmanzanares:


IagoLast:
Context?

davidmanzanares:
Easier way to benchmark the renderer by automatically launch one example and track some metrics like average time, P99...

Not super complete, but better than nothing and really simple.

Usage: launch the script and wait for the metrics on the terminal.

profile.js
```diff
@@ -0,0 +1,45 @@
+const puppeteer = require('puppeteer');
+
+
+puppeteer.launch({
+    headless: false,
+    args: ['--disable-gpu-vsync'],
+}).then(browser => {
+    browser.newPage().then(page => {
+        page.goto('file:///home/dmanzanares/github/renderer-prototype/examples/expressions/animation.html', { waitUntil: 'networkidle0' }).then(() => {
```
Jesus89: Use __dirname here

Jesus89:
This should be moved to `test/benchmark` directory and use its own `html` example for the test. It should be nice to have also the command `yarn test:benchmark` or similar.

profile.js
davidmanzanares: ups

IagoLast:
```js
 const puppeteer = require('puppeteer');
+const path = require('path');
 
+const TEST_FILE = path.join(__dirname, '/examples/expressions/animation.html');
 
 puppeteer.launch({
     headless: false,
     args: ['--disable-gpu-vsync'],
 }).then(browser => {
     browser.newPage().then(page => {
-        page.goto('file:///home/dmanzanares/github/renderer-prototype/examples/expressions/animation.html', { waitUntil: 'networkidle0' }).then(() => {
+        page.goto(`file://${TEST_FILE}`, { waitUntil: 'networkidle0' }).then(() => {
```

davidmanzanares:
:+1: @Jesus89 can you re CR it?

PR #192, opened 2018-03-20
WIP: aggregation API filtering

jgoizueta:


jgoizueta:
Note that this requires this branch of the tiler: https://github.com/CartoDB/Windshaft-cartodb/tree/cartogl

test/unit/client/windshaft-filtering.test.js
```diff
@@ -100,6 +100,13 @@ describe('src/client/windshaft-filtering', () => {
                 )).toEqual('WHERE (("categoricalProperty" IN (\'red\',\'blue\')) OR ("numericProperty" BETWEEN 10 AND 20))');
             });
 
+            fit('should handle single quotes in text literals', () => {
```
Jesus89: :eyes: `fit`

davidmanzanares:
Can you explain a little bit more how/when filters are applied? (before/after agg)

How does it affect the `filter:` style property semantics?

jgoizueta:
What this PR does is extend the cases in which we use backend filtering. But note that this may affect performance but not the semantics of style filtering, because filters are always applied but the client anyway.

The purpose of executing some of the filtering (not all is possible in general) is to enhance performance by not having to deal with so much data.

In order to do so the filters that are applied on aggregated properties are now sent to the Maps API aggregation interface (which requires the tiler update mentionend above).  Some filtering may not be executable in the server. If we can execute part of it we do: for eample if we filter by A AND B and we can only perform A in the server we do; but if we filter by A OR B we wouldn't be able to execute anything in the server.

Besides, we still extract also the filtering that affects properties before aggregation and inject it into the SQL query. (the part of it we can as in the other case). We'll need to review this at some point because a) we don't want to depend on SQL like this b) we're referencing base columns for filtering so we shouldn't add them to the requested columns as with other references to properties in the styles.

PR #194, opened 2018-03-21
Fix source type change bug

davidmanzanares:
There is a bug that when the source type (points, polygons, lines) changes, the renderer crashes. This happens since old dataframes are not removed from the renderer, therefore, the renderer detects multiple type of geometries on the same renderLayer and fails.

This can be easily reproduced by going to the example editor and change between pluto and a point dataset.

This fixes it by removing old dataframes and resetting the renderLayer geometry.

A regression test has been added.


src/api/layer.js
```diff
@@ -47,14 +47,13 @@ export default class Layer {
...
-        this.setSource(source);
-        this.setStyle(style);
-
         this._listeners = {};
         this._renderLayer = new RenderLayer();
-
         this.state = 'init';
         console.log('L', this);
+        
+        this.setSource(source);
```
IagoLast: should we await for this to complete?

davidmanzanares: We can't, is the constructor

IagoLast:
lgtm

PR #215, opened 2018-03-22
Fix MGL repaint to avoid unnecessary repaints

davidmanzanares:


src/api/integrator/mapbox-gl.js
```diff
@@ -28,8 +28,9 @@ class MGLIntegrator {
         map.on('resize', this.move.bind(this));
 
         this.moveObservers = {};
-
         this._layers = [];
+        this._paintedLayers = 0;
+        this.map.repaint = true;
```
Jesus89: This `repaint = true` could be removed

src/api/integrator/mapbox-gl.js
```diff
@@ -50,6 +50,17 @@ class MGLIntegrator {
...
             }
             layer.paintCallback();
+            this._paintedLayers++;
+
+            if (this._paintedLayers % this._layers.length == 0) {
+                // Last layer has been painted
+                const isAnimated = this._layers.some(layer =>
+                    layer.getStyle() && layer.getStyle().isAnimated());
+                if (!isAnimated && this.map.repaint) {
+                    this.map.repaint = false;
```
Jesus89: this._paintedLayers = 0 here?

davidmanzanares: Not needed, this is the last layer, therefore (this._paintedLayers % numLayers == 0 already), and, on top of that, MGL would still end this repaint anyway.

In any case, this makes me think about multiple layer cases...

src/api/integrator/mapbox-gl.js
davidmanzanares: I hope so, I'll change it

src/api/integrator/mapbox-gl.js
davidmanzanares: After giving it a minute, I think it's ok with multiple layers

src/api/integrator/mapbox-gl.js
davidmanzanares: Changed

src/api/integrator/mapbox-gl.js
Jesus89: But is there any issue if the number of layers is not a multiple of the max size of _paintedLayers (regarding overflow) ?

davidmanzanares: If `_paintedLayers` overflows probably all the logic falls down. However, assuming one layer and 60FPS it will overflow after 9007199254740991/60 seconds which is more than 4 million years :smile:

Jesus89: We can live with it :joy:

PR #218, opened 2018-03-23
Layer refactor

IagoLast:


src/api/layer.js
```diff
@@ -125,24 +125,15 @@ export default class Layer {
...
-        source.bindLayer(this._onDataframeAdded.bind(this), this._onDataFrameRemoved.bind(this), this._onDataLoaded.bind(this));
-        if (this._source !== source) {
-            this._freeSource();
+    setSource(source) {
+        if (this._source === source) {
+            return;
         }
+        this._checkSource(source);
+        this._freeSource();
         this._source = source;
```
davidmanzanares: This refactor looks very, very, dangerous to me. You are setting the object source to the new source before waiting for checking the metadata against the style.

src/api/layer.js
```diff
@@ -157,16 +148,11 @@ export default class Layer {
...
-                this._style.onChange(null);
-            }
-            this._style = style;
-            this._style.onChange(() => {
-                this._styleChanged(style);
-            });
-            return r;
-        });
+        this._style.onChange(null);
+        this._style = style;
```
davidmanzanares: Similar problem here, this sets the style without checking that the style is valid for the source

src/api/layer.js
```diff
@@ -157,16 +148,11 @@ export default class Layer {
...
-            if (this._style) {
-                this._style.onChange(null);
-            }
-            this._style = style;
-            this._style.onChange(() => {
-                this._styleChanged(style);
-            });
-            return r;
-        });
+        this._style.onChange(null);
```
davidmanzanares: What if there is no style?

src/api/layer.js
```diff
@@ -194,25 +180,21 @@ export default class Layer {
...
-            if (this._style) {
-                this._style.onChange(null);
-            }
-            this._style = style;
-            this._style.onChange(() => {
-                this._styleChanged(style);
-            });
-            return r;
-        });
+        this._style = style;
```
davidmanzanares: Similar problem

src/api/layer.js
IagoLast: Whats the possible problem?

davidmanzanares: Leaving the layer object in an invalid state.

src/api/layer.js
IagoLast: It fails but it's impossible since the constructor forces you to set a source and a style

davidmanzanares: But the constructor set it asynchronously, right?

examples/examples.json
```diff
@@ -16,11 +16,6 @@
                     "desc": "Add multiple CARTO layers to your map",
                     "file": "basics/multi-layer.html"
                 },
-                {  
-                    "title": "Change the source",
```
davidmanzanares: Why do you remove two examples?

src/api/layer.js
```diff
@@ -179,18 +127,15 @@ export default class Layer {
...
-            });
-            return r;
-        });
+
+        await this._styleChanged(style);
+        if (this._style) {
+            this._style.onChange(null);
+        }
+        this._style = style;
+        this._style.onChange(this._styleChanged.bind(this));
```
davidmanzanares: `_styleChanged` doesn't receive a style parameter?

src/api/layer.js
```diff
@@ -225,35 +170,36 @@ export default class Layer {
...
+        this._styleChanged(style).then(() => {
             if (this._style) {
                 this._style.onChange(null);
             }
             this._style = style;
-            this._style.onChange(() => {
-                this._styleChanged(style);
-            });
-            return r;
+            this._style.onChange(this._styleChanged.bind(this));
```
davidmanzanares: same

examples/examples.json
IagoLast: Because `setSource` method was removed

src/api/layer.js
IagoLast: yes, the `style` object will call this method passing `this` as parameter

examples/examples.json
davidmanzanares: Yes, but, the functionality is still there, we should use `update()` instead

src/api/layer.js
davidmanzanares: Ok

Jesus89:
Are you agree on removing `setSource` function?

Nice code
```js
citiesLayer.setSource(allCitiesSource);
```

won't be nice anymore
```js
citiesLayer.update(allCitiesSource, citiesLayer.getStyle());
```

davidmanzanares:
@Jesus89 I agree. We can always add it later. This new form is uglier, but it is safer and its semantics are easier to understand. Also, we discourage the use of SQL, particularly for simple things that can be done by the style, like filtering.

In the future, I wouldn't be surprised if we made the second argument of `update()` optional, though.

IagoLast:
@davidmanzanares can we merge this?

davidmanzanares:
Yes :rocket:

PR #219, opened 2018-03-23
Improve lines triangulation

Jesus89:
# Related to: https://github.com/CartoDB/renderer-prototype/issues/185

**Bug**: big triangles appear for big widths on zooming

big-triangles-bug

## Model

model

## First approach

- Auto crop sharp joins (< 30 deg) avoiding big triangles for low zoom levels
- Add more JS computations
- Triangulation:
first-approach
- **Problem**: there is other source of big triangles: the constant miter length bug. Both width and miter length are constant by default for all zoom values, so for low zoom levels they keep the size adding big triangles. In order to solve it, it requires extra parameters and computations on the shader side that makes the performance go down even more.

first-approach

1-triangles-bug

## Second approach

- Simple triangulation algorithm
- Less JS computations. Better performance (2x)
- No overlapping in lines (better opacity experience)
- Auto crop sharp joins (< 60 deg)
- Triangulation:
second-approach
- This approach **solves the issue**

second-approach

2-triangles-bug-fixed

NOTE: we should revisit line triangulation in the future in order to optimize the buffer size (shared vertices, shader improvements, ...)

## Changes

- Improve line triangulation
- Improve interactive example layout
- Add and update render tests for lines


davidmanzanares:
Hey @Jesus89 can you fix the conflict?

Jesus89:
Yep, for sure

src/core/decoder.js
```diff
@@ -58,75 +57,146 @@ function decodePolygon(geometry) {
...
+            if (i <= lineString.length - 2) {
+                // If there is a next point, compute its properties
+                nextPoint = [lineString[i], lineString[i + 1]];
+                nextNormal = getLineNormal(currentPoint, nextPoint);
+                // `turnLeft` indicates that the nextLine turns to the left
+                // `joinNormal` contains the direction and size for the `miter` vertex
+                //  If this is not defined means that the join must be `bevel`.
+                let {turnLeft, joinNormal} = getJoinNormal(prevNormal, nextNormal);
+                
+                // Third triangle
```
davidmanzanares: I don't understand these two triangles, and how they are compared to the first two.
The change is quite important since previously we had 2 triangles with the joint normal. Now we are having two triangles with the line normal, another third triangle with a normal that I don't understand, and a conditional fourth triangle.

I guess these it has something to do with being the **interior** of the "curve" or being the exterior (by looking at `turnLeft`), but I don't get which triangle is each one.

Also, I'm worried about the performance of this, previously we were having 2 triangles per segment, now we have between 3 and 4.

src/core/renderer.js
```diff
@@ -399,6 +399,7 @@ class Renderer {
                 gl.enableVertexAttribArray(renderer.normalAttr);
                 gl.bindBuffer(gl.ARRAY_BUFFER, tile.normalBuffer);
                 gl.vertexAttribPointer(renderer.normalAttr, 2, gl.FLOAT, false, 0, 0);
+                gl.disable(gl.BLEND); // Disable overlapping in lines
```
davidmanzanares: This is a really important change, this should be commented with @makella 

This disables blending between lines and other lines of the same layer. Blending is still done at the layer level, without introducing a new keyword or property.

I made some quick tests and I found that this improves the quality at least on the `SF` dataset, but it is important to understand the consequences of this change, particularly, lines of the same layer that are on top of another line will hide completely the bottom line, even though there will be blending between the line layer and the layers behind.

I also found that, if we remove this particular change, lines look worse than before with opacity<1.

## Tests
### Master
current
### PR
new
### PR without this change
new-with-blending

src/core/decoder.js
Jesus89: This is shown in the PR description. The third triangle is to define the `bevel` joins. The fourth triangle is used for the `miter` joins when they are required (that's why this is conditional).

The `turnLeft` is required to get the direction of the curve for having the correct orientation for the join's triangles.

In the previous implementation we had only 2 triangles per segment but with a bug. The current implementation gets rid of the bug and enables the feature for the `miter` and `bevel` style property. (Other implementation I have seen add up to 6 triangles per segment). And keep in mind that the `round` corner implementation adds lots of triangles for each join so this is not an issue.

As described in the PR, the first approach (with less number of triangles) has 2x worse performance because of the extra computations in JS and it doesn't solve the issue. The second approach has even less computations than the `master branch`'s algorithm and a similar performance in the shader side.

src/core/renderer.js
Jesus89: Yes, I think this improves the quality of the lines when there is an opacity < 1.

The default behavior in SVG is that a line does not overlap itself.

Without this change you can see the overlapping triangles in the corners but this is something that you can not avoid for this feature without losing performance.

This is for example a GeoJSON multi-line with opacity rendered by MGL (with even more weird overlapped triangles)
selection_100

makella:
@davidmanzanares @Jesus89 

This is a complicated one and I'm not sure without doing some testing, if I can provide a well-thought-out opinion...

Is there any way I can do what David was doing between current and PR?

Jesus89:
Hi @makella!

You can try to execute the project from the sources:
1. Checkout this branch `flat-corners`
2. Install the dependencies `yarn`
3. Launch the local website `yarn serve`
4. Open http://127.0.0.1:8080/

In this website you can test the changes added in this PR.

You can open https://cartodb.github.io/renderer-prototype/ in order to compare with the `master`'s version.

makella:
ok! let me try it out.

thank yoU!

makella:
@Jesus89 @davidmanzanares

This is a really tough one! The improved rendering is great, but the disabled blending between lines is a downside.  

With my testing, I agree with David that this is a big change for how overlapping lines are displayed with opacity <1 and for me, that is a really big, noticeable change.

The SF lines example has overlapping lines, but not many so it probably isn't the best for seeing the impact this can have when the overlapping lines are important for the visualization. 

Below is an example of commuting patterns where the effect is more visible:

**Master**
[MAP Link](https://cartodb.github.io/renderer-prototype/example/mapbox.html#eyJhIjoiY29sb3JhZG9fY29tbXV0ZXJfZmxvdyIsImIiOiIiLCJjIjoiY2FydG9nbCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IHNxcnQoJHdvcmtlcnNfaW5fZmxvdykvMTBcbmNvbG9yOiBvcGFjaXR5KHJhbXAobGluZWFyKCR3b3JrZXJzX2luX2Zsb3csNTAwLDEwMDApLGFnX2dybnlsKSwwLjUpXG5cblxuXG4iLCJmIjp7ImxuZyI6LTEwNC45MzY2OTUxNTI3NDU1OSwibGF0IjozOS43MDU1MzA4Nzg3MzUyNn0sImciOjEwLjE1NTUzMDQzMDcwMzk3M30=)
screen shot 2018-04-02 at 9 46 36 am



**Branch**
In this version, 
+ we are visually "summarizing" the information since we lose the blending. 
+ in addition, the lines that are overlapping, seem to have degraded rendering as compared to master

screen shot 2018-04-02 at 9 45 46 am

I wish there was an in between.

I definitely think we should look at this more closely, but for now, I would say, let's keep what we have. There aren't too many times when people will make lines as large as the initial example from @Jesus89. 

Opinions?

Jesus89:
Hi!

If we want to visualize the information of the overlapped lines we should enable the gl.BLEND definitely. I'm going to update the branch to revert the BLEND for the lines.

We need to take into account that this bug shows more with big widths, but it is present in all sizes. This for example is for `width: 4`:

selection_126

makella:
@Jesus89 ok great! i'll take a look at the same example when the blend option is there

Jesus89:
@makella branch updated. You can try again with the BLEND enabled and let us know.

makella:
@Jesus89 @davidmanzanares 

Ok, so this is better for blending purposes. 

Similar to what David was noticing, I'm seeing "more" triangles. In the SF lines example, since there are more rounded line features (vs. straight for commuters), that makes sense.

Right now, it makes most sense to NOT have the off-shooting triangles (most important) and to keep our ability to visualize overlapping lines, (blend).

For now, let's go with this version.

As mentioned earlier, I would love to take a closer look at this together (for example, how line joins are decided, etc.).

rochoa:
Which worries me about this PR is that we are fixing a scenario to make another worse: the new triangles we create for improving the corners are not dealing with the blending, creating several artefacts.

new-with-blending

Jesus89:
Yes this happens in some joins because of the approach that solves the issue. It seems inevitable to avoid both behaviors. As I commented in other occasions this is how MGL render lines with opacity:

selection_100

There are even more artifacts in the joins.

I will dedicate a bit of free time to reevaluate the first approach to see if it can be improved. Also, if someone has any idea or enlightenment about it, it would be appreciated.

src/renderer/decoder.js
```diff
@@ -115,75 +115,144 @@ function decodePolygon (geometry) {
...
+ * multiplied by a factor of `1/sin(theta)` to reach the intersecction of the wide lines.
+ * Theta is the angle between the vectors `v` and `u`. But instead of computing the angle,
+ * the `sin(theta)` (with sign) is obtained directly from the vectorial product of `v` and `u`
+ */
+function getJoinNormal (prevNormal, nextNormal) {
+    const u = [prevNormal[1], -prevNormal[0]];
+    const v = [-nextNormal[1], nextNormal[0]];
+    const sin = v[0] * u[1] - v[1] * u[0];
+    const factor = Math.abs(sin);
+    const miterJoin = factor > 0.866; // 60 deg
```
jgoizueta: With `sin` only we don't have enough information for choosing miter or bevel: factor is small when the angle is close to zero, but also when it's close to 180º.
For the latter I guess the right choice is using miter (although the difference between bevel and mitter for those small changes of direction might be undetectable, were you accounting for that?)

Jesus89: Miter is the default, and bevel only used for small angles.

Jesus89:
Yes, moving to CR. Since lines bug is fixed I think we can address Polygon stroke in other PR after merging this one.

src/renderer/decoder.js
jgoizueta: But also angles like these:

screen shot 2018-07-30 at 14 57 34

In my branch I ended up computing also the cosine to detect these cases: https://github.com/CartoDB/carto-vl/pull/765/files#diff-295612d0417d7be833afc41c3e8a2f63R294

jgoizueta:
I think any of the solutions without spikes looks great with the overlap prevention using z depth.

@davidmanzanares has been looking at some overlaps the still appear in some cases and it turns out that when lines are clipped by tile borders, the line fragment have different feature ids, so we still have a problem because of this. (And problems at tile borders would also appear apart from line clipping, because the depth buffer is cleared for each tile).

src/renderer/decoder.js
jgoizueta: After talking with @davidmanzanares this is probably not worth taking care of: the case depicted is is where the lack of miter is most noticeable; for angles closer to 180º it's hardly visible, so it might not be worth adding the miter triangle in these cases (angles between 60 and 180).

src/renderer/Renderer.js
```diff
@@ -423,6 +428,14 @@ export default class Renderer {
                 freeTexUnit++;
             }
 
+            if (tile.type === 'line' /* || tile.type === 'polygon' */) {
+                gl.clearDepth(1);
```
davidmanzanares: Trolling myself here, I think that calling `gl.depthRange(0,1)` here would be the same thing to do before using the Z buffer

src/renderer/shaders/geometry/line/lineVertexShader.glsl
```diff
@@ -34,7 +34,11 @@ void main(void) {
     color.rgb *= color.a;
     float size = decodeWidth(texture2D(widthTex, featureID).a);
 
-    vec4 p = vec4(vertexScale*(vertexPosition)+normalScale*normal*size-vertexOffset, 0.5, 1.);
+    // 64 is computed based on RTT_WIDTH and the depth buffer precision
+    // 64 = 2^(BUFFER_BITS)/RTT_WIDTH = 2^16/1024 = 64
+    float z = mod(featureID.y, 1./64.)*63. + featureID.x / (64.);
```
davidmanzanares: After seeing it with @jgoizueta this morning, the Z value here is in the [-1, 1] coordinate space (WebGL thing), so we should do `z = z * 2. - 1.;`

src/renderer/decoder.js
Jesus89: There is a fallback to bevel joins for <60º. And if we add bevel also for >60º we cover all the angles with bevel. I think we can fallback bevel also for >160º, for example.

jgoizueta:
It seems that the effect of `gl.enable(gl.DEPTH_TEST)` affects other Mapbox GL layers
(this can be checked with @Jesus89 's `debug/lines-geojson.html`example in https://github.com/CartoDB/carto-vl/tree/lines-resolution when the VL layer is added, the MGL layer disappears)

So we should restore previous DEPT_TEST configuration; I don't know if it's simply a matter of disabling it at some point or we have to keep track of the depth parameters (Func, ...) 

cc @davidmanzanares

davidmanzanares:
Unfortunately, I think this could be a bug in our MGL patch.

MGL controls the state of all the depth-related WebGL state see https://github.com/CartoDB/mapbox-gl-js/blob/carto/src/gl/context.js#L164 and https://github.com/CartoDB/mapbox-gl-js/blob/carto/src/gl/context.js#L34

I suspect that https://github.com/mapbox/mapbox-gl-js/compare/master...CartoDB:carto#diff-badfdd9928c3556e38b4a2f4ad759263R25 is not working as expected for at least one of the modified depth-related WebGL states.

Adding `gl.disable(gl.DEPTH_TEST)` to our renderer shouldn't be needed since it is supposed to have been invalidated. However, before this patch, we were already leaving the WebGL state with the `DEPTH_TEST` disabled. Therefore, I think that your proposal to add that `gl.disable` will leave the PR is a "not worse than master" state regarding MGL integration.

Since an upstream patch to create custom layers is on the way, I think we can live for a while knowing that there is probably a hidden bug in our patch.

Jesus89:
I have added little performance tests for decoding lines (10000 x 100 it). I found that the output value in master is ~25 and in this branch ~21 so this branch is faster.

davidmanzanares:
> (10000 x 100 it).

?

Jesus89:
Sorry, 10000 times inside a `for` loop and runs: 100.

davidmanzanares:
Ah, ok :+1:

src/renderer/decoder.js
Jesus89: Hi @jgoizueta. I have reviewed your proposal `!(factor < 0.866 && cos > 0.5);` and saw that this covers better the angle detection for near 180º, so I'm going to use it in this branch. Thanks.

Jesus89:
OK, I have tested performance and also the lines debug example. It seems that now we are doing a better job at joins (Orange VL, Blue MGL):

`Auto bevel join`
selection_230

IMO we are done with this. We can merge it and I'll continue with the polygon stroke implementation from this point.

NOTE: it seems that our lines are 1% wider than MGL. We should check that to know if the bug is in our roof or not.

| width 100 | width 99 |
|---|---|
| selection_231 | selection_232 |

Jesus89:
Final note was fixed in: https://github.com/CartoDB/carto-vl/pull/789

PR #224, opened 2018-03-26
Point interactivity

davidmanzanares:
Fix https://github.com/CartoDB/renderer-prototype/issues/198


src/core/dataframe.js
```diff
@@ -65,22 +65,45 @@ export default class Dataframe {
         this.style = style;
```
Jesus89: I think this `setStyle` function can be removed

davidmanzanares: Yes, that's legacy code

davidmanzanares: Fixed

src/core/dataframe.js
```diff
@@ -61,26 +61,45 @@ export default class Dataframe {
...
             default:
                 return [];
         }
     }
 
-    _getPointsAtPosition(pos) {
-        console.log(pos);
-        return [];
+    _getPointsAtPosition(p, style) {
+        p = wToR(p.x, p.y, { center: this.center, scale: this.scale });
```
Jesus89: `strokeWidth` is not taken into account here. I think we should add it to cover the whole point.

davidmanzanares: Ups, good catch, I totally forgot about that.

davidmanzanares: Solved!

test/unit/core/dataframe.test.js
```diff
@@ -1,12 +1,61 @@
...
+            };
+            const feature2 = {
+                properties: {
+                    id: 2
+                }
+            };
+            const style = {
+                getWidth: () => ({
+                    eval: () => {
+                        return 0.5;
```
Jesus89: Use width 1 to allow the "click inside point but not over stroke" test

davidmanzanares: eh? :confused:

test/unit/core/dataframe.test.js
```diff
@@ -1,12 +1,61 @@
...
+            it('should return an empty list when there are no points at the given position', () => {
+                expect(dataframe.getFeaturesAtPosition({ x: 0.5, y: 0.5 }, style)).toEqual([]);
+                expect(dataframe.getFeaturesAtPosition({ x: 5, y: 5 }, style)).toEqual([]);
+                expect(dataframe.getFeaturesAtPosition({ x: 1.0, y: 1.0 + 1.001 / 1024 }, style)).toEqual([]);
+            });
+            it('should return a list containing the features at the given position', () => {
+                expect(dataframe.getFeaturesAtPosition({ x: 0.0, y: 0.0 }, style)).toEqual([feature1]);
+                expect(dataframe.getFeaturesAtPosition({ x: 1.0, y: 1.0 }, style)).toEqual([feature2]);
+                expect(dataframe.getFeaturesAtPosition({ x: 1.0, y: 1.0 + 0.999 / 1024 }, style)).toEqual([feature2]);
+            });
```
Jesus89: Using width: 1.
- 0,0 inside point
- 0.6, 0.6 inside stroke

davidmanzanares: The inside stroke test is already being tested with  `x: 1.0, y: 1.0 + 0.999 / 1024 `

test/unit/core/dataframe.test.js
```diff
@@ -1,12 +1,61 @@
...
+                }),
+                getStrokeWidth: () => ({
+                    eval: () => {
+                        return 0.5;
+                    }
+                })
+            };
+            dataframe.renderer = { _zoom: 1, gl: { canvas: { height: 1024 } } };
+            it('should return an empty list when there are no points at the given position', () => {
+                expect(dataframe.getFeaturesAtPosition({ x: 0.5, y: 0.5 }, style)).toEqual([]);
```
Jesus89: Using width: 1.
- 0,1.1 outside stroke
- 0,5 outside stroke

davidmanzanares: ???

src/core/dataframe.js
```diff
@@ -61,26 +61,46 @@ export default class Dataframe {
...
+            const featureID = i / 2;
+            const center = {
+                x: points[i],
+                y: points[i + 1],
+            };
+            const f = {};
+            columnNames.forEach(name => {
+                f[name] = this.properties[name][featureID];
+            });
+            const scale = (styleWidth.eval(f) + styleStrokeWidth.eval(f)) * widthScale;
```
Jesus89: I think we should use `(styleWidth.eval(f) + styleStrokeWidth.eval(f)) * widthScale * 0.5` here

davidmanzanares: Why? have you manually tried the example? This matched the rendering in my tests

test/unit/core/dataframe.test.js
Jesus89: With the same width and widthStroke the point is fully covered by the stroke and you can not tests the case when the mouse is inside the point but not over the stroke.

src/core/dataframe.js
Jesus89: Yep, so I am missing something. This is the model, right?
point

So may be the 0.5 is implicit in widthScale.
Could you spot the missing part (just for my clarification :wink:)

davidmanzanares: I see your point, I think it's due to `1 / this.renderer.gl.canvas.height`. The vertex shader converts the coordinates to the [-1,1] range, therefore, a `2 / this.renderer.gl.canvas.height` is what we are doing at render time. Here the `2` is being canceled by the `0.5`.

test/unit/core/dataframe.test.js
davidmanzanares: The inside stroke case is already being tested with x: 1.0, y: 1.0 + 0.999 / 1024

Jesus89: :+1:

test/unit/core/dataframe.test.js
Jesus89: :+1:

src/core/dataframe.js
Jesus89: Ok. It would be nice to have a comment with this information in the code to avoid future misunderstandings.

Jesus89:
Great! Now maths and code are aligned. Thanks :D

Merge time!

Issue #229, opened 2018-03-27
Use unique featureID


Jesus89:
The problem is that current featureID is based on position and depends on the tile, so the same feature may have > 1 featureIDs if it covers different tiles.

- How can we identify Windshaft sources (dataset, SQL) features? cartodb_id
- How can we identify GeoJSON source features? cartodb_id

Geometries
- [x] Points
- [x] Lines
- [x] Polygons

Jesus89:
Closed by: https://github.com/CartoDB/renderer-prototype/pull/230

PR #230, opened 2018-03-27
Use unique featureid

IagoLast:
#229
----

- Use `cartodb_id` as feature id
- Set `cartodb_id` on GeoJSON sources
- Request `cartodb_id` to windshaft sources (Dataset, SQL)
- Update interactivity examples: points, lines, polygons
- Add unit tests
- Add user tests


IagoLast:
🍐 ✨

examples/interactivity/lines/feature-click.html
```diff
@@ -0,0 +1,122 @@
+<!DOCTYPE html>
```
davidmanzanares: We've changed the examples from their directory like 10 times in 4 weeks, I really don't care, but this is much noise in the PRs, it makes impossible to check for actual change on the examples and it doesn't get along with the browser history, please, commit to one path and stick to it in following PRs. Thank you

examples/interactivity/polygons/feature-click.html
```diff
@@ -93,24 +93,25 @@ <h2></h2>
         console.log('CLICK OUT', event.features);
       });
     });
-    
+
```
davidmanzanares: Another :rainbow: , I don't know why there are so many spaces like this one, I see lots of PRs, including mine that removed them, but I don't know how they are introduced in the first place, it would be nice to get rid of them :slightly_smiling_face:

examples/interactivity/polygons/feature-click.html
```diff
@@ -5,11 +5,11 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta charset="UTF-8">
```
davidmanzanares: I cannot access the examples from http://127.0.0.1:8080/examples/ I guess it's because of the path change

src/client/windshaft.js
```diff
@@ -72,6 +72,10 @@ export default class Windshaft {
         const MNS = style.getMinimumNeededSchema();
         const resolution = style.getResolution();
         const filtering = windshaftFiltering.getFiltering(style);
+        // FIXME: force to include `cartodb_id` in the MNS columns
```
davidmanzanares: Why Fixme?

src/client/windshaft.js
```diff
@@ -469,6 +475,9 @@ export default class Windshaft {
             t.categoryNames.map(name => categoryIDs[name] = this._getCategoryIDFromString(name));
             columns.push(t);
         });
+        // FIXME: force columns to include `cartodb_id` in the Metadata
```
davidmanzanares: I thought this wasn't needed after all

examples/interactivity/lines/feature-click.html
Jesus89: Yes. The issue with that is that we have the same feature events with different geometries so they are now organized by geometry. Agree. We should move files in other PRs to avoid this diff issue :wink:

examples/interactivity/polygons/feature-click.html
Jesus89: You need to enter in http://127.0.0.1:8080/examples/interactivity because they are not added yet in the `examples.json` file

examples/interactivity/lines/feature-click.html
davidmanzanares: :+1:

examples/interactivity/polygons/feature-click.html
Jesus89: I have checked my editor (Atom) to "Ensure Single Trailing Newline". @davidmanzanares and @IagoLast you should check also your config preferences.

src/client/windshaft.js
Jesus89: Just to mark that this is a pseudo-hack and we should revisit it in the future.

examples/interactivity/polygons/feature-click.html
davidmanzanares: I accessed the example directly, but, don't we want to add them in the `json` file?

src/client/windshaft.js
davidmanzanares: `FIXME` sounds to me like "FIXME ASAP - possible root of serious bugs and really bad things", not like Maps API requires a little bit of care, but it's your call

examples/interactivity/polygons/feature-click.html
Jesus89: Yes but not now. These are "internal" examples. We need to discuss with @makella how the examples must be and may be add new examples (oriented to the final user) and then add the reviewed examples to the json file.

src/client/windshaft.js
Jesus89: Yep, it seems that, with the changes in MNS, this is not required. I'll remove it.

src/client/windshaft.js
Jesus89: We can use TODO instead.

PR #231, opened 2018-03-27
Add interactivity docs

davidmanzanares:
Fixes https://github.com/CartoDB/renderer-prototype/issues/227 and https://github.com/CartoDB/renderer-prototype/issues/201

### DIFF
--- a/src/api/interactivity.js
+++ b/src/api/interactivity.js
@@ -102,8 +102,6 @@ export default class Interactivity {
     constructor(layerList) {
         checkLayerList(layerList);
         this._init(layerList);
-
-        this._prevFeatures = [];
     }
 
     /**
@@ -138,6 +136,7 @@ export default class Interactivity {
         this._emitter = mitt();
         this._layerList = layerList;
         this._subscribeToIntegratorEvents(layerList[0].getIntegrator());
+        this._prevFeatures = [];
     }
 
     _subscribeToIntegratorEvents(integrator) {
--- a/src/api/interactivity.js
+++ b/src/api/interactivity.js
@@ -3,6 +3,66 @@ import Layer from './layer';
 import { WM_R, projectToWebMercator } from './util';
 import { wToR } from '../client/rsys';
 
+
+/**
+ * 
+ * FeatureEvent objects are fired by {@link carto.Interactivity|Interactivity} objects.
+ * @typedef {Object} FeatureEvent
+ * @property {object} coordinates LongLat coordinates in {lng: 0, lat:0} form
+ * @property {object} position pixel coordinates in {x: 0, y: 0} form
+ * @property {Array<carto.Feature>} features array of {@link carto.Feature}
+ * @api
+ */
+
+/**
+ * 
+ * Feature objects are provided by {@link carto.FeatureEvent} events.
+ * @typedef {Object} Feature
+ * @property {Object} properties Object with the feature properties in {propertyName1: 12.4, propertyName2: 'red'} form
+ * @api
+ */
+
+
+
+/**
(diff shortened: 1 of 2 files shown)


src/api/interactivity.js
```diff
@@ -3,6 +3,66 @@ import Layer from './layer';
 import { WM_R, projectToWebMercator } from './util';
 import { wToR } from '../client/rsys';
 
+
+/**
+ * 
+ * FeatureEvent objects are fired by {@link carto.Interactivity|Interactivity} objects.
```
IagoLast: what is a feature event? what is it used for?

src/api/interactivity.js
```diff
@@ -3,6 +3,66 @@ import Layer from './layer';
...
+ * @typedef {Object} FeatureEvent
+ * @property {object} coordinates LongLat coordinates in {lng: 0, lat:0} form
```
IagoLast: missing id property

src/api/interactivity.js
```diff
@@ -3,6 +3,66 @@ import Layer from './layer';
...
+
+
+/**
+ * featureClick events are fired when the user clicks on features. The list of features behind the cursor is provided.
+ *
+ * @event featureClick
+ * @type {FeatureEvent}
+ * @api
+ */
+/**
```
IagoLast: add white line to split

src/api/interactivity.js
```diff
@@ -11,29 +71,68 @@ const EVENTS = [
...
+    *
+    * Interactivity purpose is to allow the reception and management of user-generated events, like clicking, over layer features.
+    *
+    * To create a Interactivity object an array of {@link carto.Layer} is required.
+    * Events fired from interactivity objects will refer to the features of these layers and only these layers.
+    *
+    * @param {Array<carto.Layer>} layerList - Array of {@link carto.Layer}, events will be fired based on the features of these layers. The array cannot be empty, and all the layers must be attached to the same map.
+    *
+    * @example
+    * const myLayer = new carto.Layer('layer0', mySource, myStyle);
```
IagoLast: this example is wrong, we must wait for the . `layer.loaded` event to create the interactivity.

In the future, I think we should make the addtoMap async

src/api/interactivity.js
```diff
@@ -11,29 +71,68 @@ const EVENTS = [
...
         this._init(layerList);
-        
-        this._prevFeatures = [];
-    }
 
-    _init(layerList) {
-        this._emitter = mitt();
-        this._layerList = layerList;
-        this._subscribeToIntegratorEvents(layerList[0].getIntegrator());
+        this._prevFeatures = [];
```
IagoLast: this should be moved to the `init` method

src/api/interactivity.js
davidmanzanares: It is an `Object`, usage is end-app scope, not ours.

src/api/interactivity.js
davidmanzanares: Missing on `Feature`, not here, and that's not implemented on `master`

src/api/interactivity.js
davidmanzanares: Yes

src/api/interactivity.js
davidmanzanares: This is certainly inconsistent (although it was there before this PR), however, I doubt we need a `_init` method in the first place, it is pure side-effect and it is only called by the constructor, also, it is 4 lines long

src/api/interactivity.js
davidmanzanares: Fixed

PR #234, opened 2018-03-28
Hex

davidmanzanares:


src/core/style/expressions/hex.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ *
+ * Create a color from their hexadecimal description
+ *
+ * @param {string} hexadecimalColor - color in the form #AA88AA
+ * @return {carto.style.expressions.hex}
+ *
+ * @example <caption>Display blue points.</caption>
+ * const s = carto.style.expressions;
+ * const $type = s.property('type');
```
Jesus89: $type not used

src/core/style/expressions/hex.js
```diff
@@ -0,0 +1,40 @@
+import Expression from './expression';
+import { checkString, hexToRgb, getStringErrorPreface } from './utils';
+
+/**
+ *
+ * Create a color from their hexadecimal description
```
IagoLast: Create a color from `its` hexadecimal description?

src/core/style/expressions/utils.js
```diff
@@ -15,12 +15,18 @@ export function implicitCast(value) {
 }
 
 export function hexToRgb(hex) {
+    // TODO allow for #F00 and #F004 notations
```
IagoLast: Maybe we should point in the doc this forms are not supported

Jesus89:
Added support for `#ABC` colors

PR #238, opened 2018-04-02
Named colors

IagoLast:
- Add new `color` expression allowing named colors.
- Add an example
- Add naive tests


src/core/style/expressions/color.js
```diff
@@ -0,0 +1,54 @@
...
+        const color = this._nameToRGB(this.name);
+        this.inlineMaker = () => `vec4(${(color.r / 255).toFixed(4)}, ${(color.g / 255).toFixed(4)}, ${(color.b / 255).toFixed(4)}, ${(1).toFixed(4)})`;
+    }
+
+    _nameToRGB(name) {
+        const colorRegex = /rgb\((\d{1,3}), (\d{1,3}), (\d{1,3})\)/;
+        const fakeDiv = document.createElement('div');
+        fakeDiv.style.color = name;
+        document.body.appendChild(fakeDiv);
+        const rgbSring = getComputedStyle(fakeDiv).color;
```
davidmanzanares: Is this browser-independent?

IagoLast: https://caniuse.com/#feat=getcomputedstyle

davidmanzanares: And that means it not only works, but that it works consistently so that the regex can work?

IagoLast: Good point:

I created this block:
https://bl.ocks.org/IagoLast/raw/137e831dd1f7d065a4ccd2ec2b365407/

And I'm testing against different system/browser configurations:

screen shot 2018-04-02 at 13 28 20

screen shot 2018-04-02 at 13 29 29


screen shot 2018-04-02 at 13 30 11

screen shot 2018-04-02 at 13 31 54

davidmanzanares: Looks good to me

src/core/style/expressions/color.js
```diff
@@ -0,0 +1,54 @@
...
+ * const style = new carto.Style({
+ *   color: s.color('red');
+ * });
+ *
+ * @memberof carto.style.expressions
+ * @name hex
+ * @function
+ * @api
+ */
+export default class Color extends Expression {
```
davidmanzanares: I would change the name, the String API could be improved to automatically use these without function calls, with JS, I don't know if this is so useful, maybe `NamedColor` is better

IagoLast: ok, `NamedColor`

examples/expressions/color.html
```diff
@@ -78,6 +78,10 @@
         <input type="radio" name="style" onclick="setOpacityGreen()" id="green">
         <label for="green">Opacity - Green</label>
       </li>
+      <li>
+        <input type="radio" name="style" onclick="setNamedColor()" id="named">
```
Jesus89: id="orange"

examples/expressions/color.html
```diff
@@ -78,6 +78,10 @@
...
+        <label for="green">Named Color - Orange</label>
```
Jesus89: for="orange"

examples/expressions/color.html
```diff
@@ -142,6 +149,10 @@
     function setOpacityGreen() {
       layer.blendToStyle(styleGreen);
     }
+
+    function setNamedColor(){
```
Jesus89: `setNamedColorOrange`

Jesus89:
I don't really like the syntax `namedColor('orange')`. I know we have the same issue for the column properties as in the hex colors (`hex('#ABC')`) but here there are so many letters.

Maybe @makella could help us with this, but we need to thought also about implicit cast for these text colors.

davidmanzanares:
@Jesus89 Yes, `namedColor` is ugly, but the String API could use names without this. 

In fact, @IagoLast , it would be nice if you add that feature, it's another `else if` block in https://github.com/CartoDB/renderer-prototype/blob/master/src/core/style/parser.js#L133

Regarding the JS API, if you come up with a good name perfect, but `color` is too generic, and, in any case, I don't think this feature is very useful outside the String API scope.

IagoLast:
@Jesus89 @davidmanzanares can you update the `named` reference in the integration tests?

Jesus89:
Yes :point_up:

Jesus89:
What is the final String API?

makella:
hey guys, sorry missed this earlier. 

so what is the final notation for HEX and named colors?

PR #241, opened 2018-04-02
Flex properties

davidmanzanares:
This PR lift the maximum number of properties limitation.

Previously, there was a fixed limit of properties: `4`. After the `cartodb_id` PR was merged, that limit was being hit my multiple examples.

This removes the windshaft.js limitation, and it lifts the webgl limitation.

Now, WebGL texture units are shared between property textures and auxiliar (styling expressions like `ramp`) textures.

I manually checked that the examples work now, but a test wouldn't harm.

Also, the number of required texture units is not controlled (this was the previous case, too). This should be controlled, but how is not clear. One option is to throw an error if we reached the current hardware limit, another option is to throw an error if we reach a fixed limit (this is important since, if we take the first route, one map could work on one hardware but could not work on another hardware).

cc: @rochoa


Jesus89:
Looks good to me, but I miss the tests. How can this feature be tested with our current tests (unit, render, user, e2e)? (an example)

davidmanzanares:
A render test should do the trick, I'll do it

src/client/windshaft.js
```diff
@@ -406,7 +406,10 @@ export default class Windshaft {
     }
 
     _decodeMVTLayer(mvtLayer, metadata, mvt_extent, catFields, numFields, datesField) {
-        var properties = [new Float32Array(mvtLayer.length + 1024), new Float32Array(mvtLayer.length + 1024), new Float32Array(mvtLayer.length + 1024), new Float32Array(mvtLayer.length + 1024)];
+        const properties = [];
+        for (let i=0; i<catFields.length+numFields.length; i++){
```
IagoLast: 🌈 missing whitespace between operators

davidmanzanares:
@Jesus89 I created the test I checked that it fails before this PR.

src/client/windshaft.js
davidmanzanares: Fixed

PR #253, opened 2018-04-03
Allow non-loaded layers in interactivity constructor layerlist

### LINKED ISSUE #250: Allow non-ready layers in carto.Interactivity
Now a `carto.Interactivity` object must be instantiated inside a `loaded` event callback. But this implementation has errors when multiple layers are used:

```js
// layer2
layer1.on('loaded', () => {
  const interactivity = carto.Interactivity([layer1, layer2]);
});
layer1.addTo(map);
layer2.addTo(map);
```

The solution is to allow non-ready layers in the `carto.Interactivity` constructor and manage internally the errors and async cases.

```js
const interactivity = carto.Interactivity([layer1, layer2]);
layer1.addTo(map);
layer2.addTo(map);
```

davidmanzanares:
Fix #250

### DIFF
--- a/src/api/interactivity.js
+++ b/src/api/interactivity.js
@@ -104,7 +104,7 @@ export default class Interactivity {
             // Allow one layer as input
             layerList = [layerList];
         }
-        checkLayerList(layerList);
+        preCheckLayerList(layerList);
         this._init(layerList);
     }
 
@@ -139,8 +139,11 @@ export default class Interactivity {
     _init(layerList) {
         this._emitter = mitt();
         this._layerList = layerList;
-        this._subscribeToIntegratorEvents(layerList[0].getIntegrator());
         this._prevFeatures = [];
+        Promise.all(layerList.map(layer => layer._context)).then(() => {
+            postCheckLayerList(layerList);
+            this._subscribeToIntegratorEvents(layerList[0].getIntegrator());
+        });
     }
 
     _subscribeToIntegratorEvents(integrator) {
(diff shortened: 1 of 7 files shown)


src/api/interactivity.js
```diff
@@ -139,8 +139,11 @@ export default class Interactivity {
     _init(layerList) {
         this._emitter = mitt();
         this._layerList = layerList;
-        this._subscribeToIntegratorEvents(layerList[0].getIntegrator());
         this._prevFeatures = [];
+        Promise.all(layerList.map(layer => layer._context)).then(() => {
```
IagoLast: I think we need some kind of public method `isLoaded` instead of relying on private properties.

What happens if in the future we need the context and `metadatum` to consider a layer loaded?

test/unit/api/interactivity.test.js
```diff
@@ -53,15 +53,7 @@ describe('api/interactivity', () => {
...
-        it('should throw an error when the layers have no integrator', () => {
-            layer.getIntegrator = () => undefined;
-
-            expect(() => new Interactivity([layer])).toThrowError('Invalid argument, all layers must belong to some map');
-
-            layer.getIntegrator = () => mockIntegrator;
-        });
-
-        it('should throw an error when the layers have different integrator', () => {
+        xit('should throw an error when the layers have different integrator', () => {
```
Jesus89: These test should be updated to cover the new scenario

src/api/interactivity.js
davidmanzanares: "What happens if in the future we need the context and metadatum to consider a layer loaded?" => "If in the future" is a smell already

test/unit/api/interactivity.test.js
davidmanzanares: This is what I talked about off-channel

davidmanzanares:
User tests should be fixed, yes

davidmanzanares:
User tests fixed, IMO, we should merge this

davidmanzanares:
Thank you!

PR #255, opened 2018-04-03
Throw an error on shared styles by  multiple layers

davidmanzanares:
Fix https://github.com/CartoDB/renderer-prototype/issues/212

### DIFF
--- a/test/unit/api/layer.test.js
+++ b/test/unit/api/layer.test.js
@@ -53,6 +53,14 @@ describe('api/layer', () => {
                 new Layer('layer0', source, {});
             }).toThrowError('The given object is not a valid style. See "carto.Style".');
         });
+
+
+        it('should throw an error if a style is already added to another layer', () => {
+            new Layer('layer1', source, style);
+            expect(() => {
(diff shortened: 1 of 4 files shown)


test/unit/api/layer.test.js
```diff
@@ -53,6 +53,14 @@ describe('api/layer', () => {
...
             }).toThrowError('The given object is not a valid style. See "carto.Style".');
         });
+
+
+        it('should throw an error if a style is already added to another layer', () => {
+            new Layer('layer1', source, style);
+            expect(() => {
+                new Layer('layer2', source, style);
+            }).toThrowError(/[\s\S]*shared[\s\S]*/g);
+        });
```
IagoLast: I think we must check the .blendto and the .setStyle functions also

davidmanzanares: true

davidmanzanares: done for blendto, for setstyle see https://github.com/CartoDB/renderer-prototype/issues/256

PR #257, opened 2018-04-03
WIP Allow shared sources by multiple layers

### LINKED ISSUE #211: Allow shared sources by multiple layers

davidmanzanares:
Fix #211 
Lacks tests


davidmanzanares:
This approach is too problematic, sharing a source by multiple layers is a huge deal since sources don't expect to add previously added dataframes, therefore, when a layer is added with a previously used source old dataframes present on the former layer won't be added to the new one. Resolving this problem is actually too complex.

PR #260, opened 2018-04-04
Shared sources 2

### LINKED ISSUE #211: Allow shared sources by multiple layers

davidmanzanares:
Fix #211 

It's done by cloning sources, which is not a problem since sources are immutable from the user perspective. Of course, this is not free in terms of performance, but I prefer to add this now than to waste days in this corner-case.

However, I thought about a related performance concern: using one Maps API instantiation for multiple Windshaft sources. For this case, I think that some global private state should be stored in windshaft.js to support it transparently, and I think that won't clash with this PR (I hope).

Note: https://github.com/CartoDB/renderer-prototype/pull/257 PR was a former attempt that never made it.


Jesus89:
This is not very nice, but I understand the motivation.

I think a unit test in `layer.test.js` is missing to ensure that we don't break this functionality for any source (Dataset, SQL, GeoJSON).

davidmanzanares:
What do you want me to test? that `_clone()` exists? I already added a test that covered shared sources, and, of course, the common case path of a unique source is tested in all render tests.

Anyway, it was true that MGL with shared sources was an untested path, I added a test for that.

Jesus89:
Since Dataset and SQL are not covered in render tests we should add unit tests for that. I would add:

- [ ] Unit test in Layer with a Dataset: check that the layer.getSource() != Dataset source
- [ ] Unit test in Layer with a SQL: check that the layer.getSource() != SQL source
- [ ] Unit test in Layer with a GeoJSON: check that the layer.getSource() != GeoJSON source

In the `describe('constructor'` in `layer.test.js`

Jesus89:
You can test instead that the `source._clone()` function is called once, when the Layer is created.

src/api/layer.js
```diff
@@ -42,6 +42,7 @@ export default class Layer {
     }
 
     _init(id, source, style) {
+        source = source._clone();
```
Jesus89: I think this line is not needed (extra _clone)

davidmanzanares: Why? the constructor doesn't clone it, and there can be two layers with the same source passed in both cases through the constructor

davidmanzanares:
Oh, I understand your point now, will add those unit tests

src/api/layer.js
Jesus89: I think this is done in `this.update(source, style);` at the end of the `_init` function

davidmanzanares: Yep, you're right

PR #261, opened 2018-04-04
Fix ordering for limit case

davidmanzanares:
Fix https://github.com/CartoDB/renderer-prototype/issues/258


Jesus89:
I'm wondering how a render test for this case would be

davidmanzanares:
Since ordering has very little expectations from the API point of view, you'll be just testing the implementation

PR #263, opened 2018-04-04
Control that resolution value is in range

davidmanzanares:
Fix https://github.com/CartoDB/renderer-prototype/issues/209

### DIFF
--- a/test/unit/api/style.test.js
+++ b/test/unit/api/style.test.js
@@ -94,7 +94,7 @@ describe('api/style', () => {
                 };
                 expect(function () {
                     new Style(styleSpec);
-                }).toThrowError('`resolution` must be less than 255.');
+                }).toThrowError('`resolution` must be less than 256.');
             });
 
             it('should throw an error when color is not a valid expression', () => {
--- a/test/unit/api/style.test.js
+++ b/test/unit/api/style.test.js
@@ -85,7 +85,7 @@ describe('api/style', () => {
                 };
                 expect(function () {
                     new Style(styleSpec);
-                }).toThrowError(/[\s\S]*must.*greater[\s\S]*/g);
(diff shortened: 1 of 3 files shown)


test/unit/api/style.test.js
```diff
@@ -79,6 +79,24 @@ describe('api/style', () => {
                 }).toThrowError('`resolution` must be a number.');
             });
 
+            it('should throw an error when resolution is too small', () => {
+                const styleSpec = {
+                    resolution: 0
+                };
+                expect(function () {
+                    new Style(styleSpec);
+                }).toThrowError(/[\s\S]*must.*greater[\s\S]*/g);
```
Jesus89: I like to use the full string exception. `'`resolution` must be greater than $0.'`

test/unit/api/style.test.js
```diff
@@ -79,6 +79,24 @@ describe('api/style', () => {
...
+            });
+
+            it('should throw an error when resolution is too big', () => {
+                const styleSpec = {
+                    resolution: 10000
+                };
+                expect(function () {
+                    new Style(styleSpec);
+                }).toThrowError(/[\s\S]*must.*less[\s\S]*/g);
```
Jesus89: Same here:
'`resolution` must be less than 255.'

test/unit/api/style.test.js
davidmanzanares: That repeats the entire error message two times, why coupling so much the test when it is not needed?

Jesus89: The error strings are part of the API, as a function or a namespace. If you change it, the interface you provide to the user changes and we need to detect that the API has changed.

In this case if you change min or max resolution (API) no test fails. Adding the full text this case is also covered.

PR #270, opened 2018-04-09
Mapbox gl integration

IagoLast:
https://github.com/CartoDB/renderer-prototype/issues/44
---

This PR uses our own fork from a CDN in the examples while `carto-vl` support is not included in the mapboxgl repo.

cc @rochoa @Jesus89 @davidmanzanares

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,7 +19,7 @@ TBD
   <!-- Include CARTO GL JS -->
   <script src="../../dist/cartovl.js"></script>
   <!-- Include CARTO Mapbox GL JS fork -->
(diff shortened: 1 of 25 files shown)


Jesus89:
As we talked off-line, we should upload those packages to our CDN (S3 bucket) and use it from there.

I would also remove our fork of mapboxgl from `npm`.

IagoLast:
> I would also remove our fork of mapboxgl from npm

why?

Jesus89:
We will use our CDN instead of unpkg.

IagoLast:
> We will use our CDN instead of unpkg.

ok but in any case this is not important now

PR #277, opened 2018-04-10
Visualization

davidmanzanares:
# Visualization

As discussed off-channel, we want to test the idea of replacing `Style` objects by a more powerful `Viz` ones. This includes:

- [x] Support custom variables
    - [x] JS API
    - [x] String API
- [x] Allow accessing aliases with the JS API
- [x] Support global* dataviews 
- [x] Support referencing custom variables inside style: aliases
- [x] Control circular dependencies
- [x] Support multiple references to the same alias:
    - [x] multiple shader bindings to expressions
    - [x] aliases notify all parents
    - [x] aliases replaceChild work as expected 
- [x] Fix tests
- [x] Add tests
- [x] Add variable expression
- [x] Use a preliminary prefix instead of `@` in the parser
- [x] Remove variable functionality from `property` expression.
- [x] Remove feature properties
- [x] Add `@` to parser
- [x] Fix interactivity tests
- [x] Fix examples

To be fixed by a following PR:
- [ ] Improve namespace 
- [ ] Viz rename
- [ ] Fix docs
- [ ] Add docs

Outside of the scope of this PR:

- [ ] Support viewport* dataviews
    - [x] points
    - [ ] lines
    - [ ] polygons

I want to centralize the discussion of the topic here. In this PR, that at this moment is just words, also, I want to make this PR thread the source of all related news, implemented features, implementation problems, API problems...

## Rationale
We want to support **dataviews** sooner or later. Right now, we found some problems regarding the users who want to get some properties on **interactivity**, but without styling by them. **Dataviews** are a similar case: computations that can already be expressed in terms of expressions (see viewport* expressions) cannot be accessed without using them in a `Style` object.

One option would be to add new API objects to solve this, but that has at least two main problems:
- Users don't want two APIs to do one thing
- The current implementation relies on having only one object that affects the source/data extraction: the `Style`. Having similar objects would probably make the library much more fragile and difficult to maintain.

Replacing `Style` by `Viz`, and allowing `Viz` to have variables/custom properties, which can be accessed in JS, but that doesn't need to be used in the rendering style, solve the issue.

## Examples
```
const airbnbViz = new carto.Viz(`
    columns:
        $name
        $owner
        $type

    variables:
        vpmaxprice: viewportMax($price)
        sum_price: sum($price)
        priceComparison:  viewportAvg($price)/globalAvg($price) - 1
        priceRangeLow:    viewportPercentile($price, 1)
        priceRangeHigh:   viewportPercentile(sum($price), 99)

        // Availability widget
        numAvailable:     viewportSum($available)
        availabilityPct:  viewportSum($available)/viewportCount()

    style:
      width: ramp(linear($price, $priceRangeLow, $vpmaxprice), temps)
      color: rgb(255, 76, 89)
    `);


const layer = new carto.Layer('layer', source, airbnbViz);

blablabla.on('move', () => {
    updateMyAwesomePriceWidget({
        priceComparison: airbnbViz.priceComparison.eval(),
        priceRange: [airbnbViz.priceRangeLow.eval(), airbnbViz.priceRangeHigh.eval()]
    });
    updateMyAwesomeAvailabilityWidget({
        numAvailable: airbnbViz.numAvailable.eval(),
        availabilityPct: airbnbViz.availabilityPct.eval(),
    });
});

const interactivity = new carto.Interactivity(layer);
interactivity.on('featureHover', featureEvent => {
    if (!featureEvent.features) {
        updateMyAwesomePopup(null);
    } else {
        updateMyAwesomePopup({
            name:   featureEvent.features[0].properties.name,
            normalizedPrice:  featureEvent.features[0].properties.price / featureEvent.features[0].viz.avgPrice
        });
    }
});



const footTrafficViz = new carto.Viz(`
heatmap:    sum($count)
heatmapFn:  ramp(linear($heatmap, viewportMin($heatmap), viewportMax($heatmap)), temps)


// Foot traffic widget
total:  viewportSum($count)
`);
```


IagoLast:
I don't see the advantages of having a Viz object in fact, reminds me the old `viz.json` which by the way had to be replaced for specific APIs in carto.js.


IMO we should centralize all data/column related logic in the source to keep high cohesion between concepts so:

> All columns used by the interactivity, dataviews and styles are defined in the source (we could eventually add syntax sugar to use directly properties defined by the style)

This is equivalent to carto.js where the documentation says: 

> Use a source to reference the data used in a dataview or a layer

So a layer is an special case of dataview where the data (specified by the source) is displayed in a "drawing" called map.

To specify how the map will be shown you need to use a `style`.

```js
const source = new carto.Source('dataset', {
  columns: ['price'],
  aggregated_columns: {
      avgPrice: {
          aggregate_function: 'avg',
          aggregated_column: 'price'
      }
});

const style = new carto.style(`
  color: red,
  width: $price / $avgPrice
`);
```

### Dataviews

```js
const source = new carto.Source('dataset', {
  columns: ['column0', 'column1'],
  aggregated_columns: {
      value: {
          aggregate_function: 'avg',
          aggregated_column: 'value'
      }
});

const dataview = new carto.Dataview(source);
dataview.addField('viewportAvg', 'column0');
dataview.on(data => doSomething(data));
```


### Interactivity
```js
const source = new carto.Source('dataset', {
  columns: ['price', 'name'],
  aggregated_columns: {
      avgPrice: {
          aggregate_function: 'avg',
          aggregated_column: 'price'
      }
});

const style = new carto.style(`...`);
const layer = new carto.layer(source, style);
const interactivity = new carto.Interactivity(layer);

interactivity.on('featureHover', featureEvent => {
    if (!featureEvent.features) {
        updateMyAwesomePopup(null);
    } else {
        updateMyAwesomePopup({
            name:   featureEvent.features[0].properties.name,
            normalizedPrice:  featureEvent.features[0].properties.price / featureEvent.features[0].properties.avgPrice
        });
    }
});
```


This API looks clean, simple, easy to use and easy to understand to me. Please feel free to point its drawbacks.

We can discuss the some points:

 - can we refer to a column in the style without defining it in the source?
 - what features are listed in the interactivity?
 -

rochoa:
First of all, do we plan to have a programmatic API? If so, I want to see how the programmatic API would be.

rochoa:
> - Everything is merged in a single object (a simila and more general concept of vizzualization was tested in the past and clients wanted low-level apis)

Can you be more specific? Something that user didn't like and it's similar here.

> - We need a new meta-lang: (+ documentation, + parsing, + learn how to use it)

This is not different from our current styling language, is it?

 
> I don't see the advantages of having a Viz object in fact, reminds me the old `viz.json` which by the way had to be replaced for specific APIs in carto.js.

The viz.json is way more than styling, it also contains metadata for the rest of the elements: map zoom/center, legends, annotations, pop-ups, attributions.

> This is equivalent to carto.js where the documentation says: 
> > Use a source to reference the data used in a dataview or a layer

Raster relies on server side all the time, that why using the source makes sense. However, here, there are interactions that happens in the client. For instance, the dataview is computed in the client.

> So a layer is an special case of dataview where the data (specified by the source) is displayed in a "drawing" called map.

That's true, but that means that the logic/manipulation for all the filtering/aggregations should happen in the *Source*.


> To specify how the map will be shown you need to use a `style`.
> 
> ```js
> const source = new carto.Source('dataset', {
>   columns: ['price'],
>   aggregated_columns: {
>       avgPrice: {
>           aggregate_function: 'avg',
>           aggregated_column: 'price'
>       }
> });
> 
> const style = new carto.style(`
>   color: red,
>   width: $price / $avgPrice
> `);
> ```

Not a good example, `$price/$avgPrice`. Don't you see much more expression friendly to do:

```js
Viz(`
color: red
width: $price/avg($price)
`)
```

> ```js
> const source = new carto.Source('dataset', {
>   columns: ['column0', 'column1'],
>   aggregated_columns: {
>       value: {
>           aggregate_function: 'avg',
>           aggregated_column: 'value'
>       }
> });
> 
> const dataview = new carto.Dataview(source);
> dataview.addField('viewportAvg', 'column0');
> dataview.on(data => doSomething(data));
> ```

What's "_`addField('viewportAvg', 'column0');`_"?


> This API looks clean, simple, easy to use and easy to understand to me. Please feel free to point its drawbacks.

Probably it's easy to reason about at the beginning. But I don't agree on being simple: it forces you to declare some variable for later reference. We should add types! 😈

makella:
From a less technical POV, I really like the ability to define variables that can be used interchangeably for visualization, interactivity, widgets etc. From a styling language point of view, it is nice that this simplifies some of the complexity that will be inherent with advanced expressions and variables are also intuitive and known in CartoCSS.

While we were working on the interactivity examples, it occurred to me that it doesn't make a ton of sense to have to style with the variables that you want to show in something like a pop-up or hover. For example, in order to get a category (country name) and a numeric value (population), in a pop-up, you HAVE to style by category which in this case, requires that we have 195 unique colors! This defeats the purpose of good cartographic practices. There is no way that anyone will be able to distinguish each unique color in the map.

@davidmanzanares in this example, `temps` should be numbers, right? and could `color` also become a visualization variable? 

```
width: ramp(linear($price, $priceRangeLow, $vpmaxprice), temps)
color: rgb(255, 76, 89)
```

IagoLast:
@rochoa 

> Can you be more specific? Something that user didn't like and it's similar here.

The cartojsv3 -> cartojsv4 migration was about moving from a declarative API (viz.json) to a programmatic one because the users demanded a higher control. I see a small similarity here since we are using a "super-object" to manage all stuff.

> This is not different from our current styling language, is it?

The styling language is an entirely new thing.  [Despite we already have methods to define columns and aggregations as a simple json](https://carto.com/docs/carto-engine/maps-api/tile-aggregation/) we are going to extend the `styling?` language with new stuff...

> Raster relies on server side all the time, that why using the source makes sense. However, here, there are interactions that happens in the client. For instance, the dataview is computed in the client.

The dataview is computed from the client using the data and columns specified in the source :P 

> Not a good example, $price/$avgPrice. Don't you see much more expression friendly to do:

As I said, we can add some sugar to the API so this kind of things could be expressed as:

```js
const style = new carto.style(``
  width: $price/ avg($price)
`);
```

> What's "addField('viewportAvg', 'column0');"?

David told me one of the main advantages of this api is you can add local expressions for dataviews. The code snippet is an API proposal where you can include in the dataview the viewportAvg of the `column0``


> it forces you to declare some variable for later reference

If we add the sugar we remove this constraint and both apis are exaclty the same

src/api/style.js
```diff
@@ -185,7 +187,7 @@ export default class Style {
             this._styleSpec.strokeColor,
             this._styleSpec.strokeWidth,
             this._styleSpec.filter,
-        ].filter(x => x && x._getMinimumNeededSchema);
+        ].concat(Object.keys(this._styleSpec.variables).map(varname => this._styleSpec.variables[varname])).filter(x => x && x._getMinimumNeededSchema);
```
IagoLast: IMO this would be easier to read and understand
```js
.concat(this._getVariables())
.filter(x => x && x._getMinimumNeededSchema);
```


```js
_getVariables() {
  return Object.values(this._styleSpec.variables);
}
```

src/api/style.js
```diff
@@ -195,10 +197,52 @@ export default class Style {
         this._compileStrokeColorShader(gl, metadata);
         this._compileStrokeWidthShader(gl, metadata);
         this._compileFilterShader(gl, metadata);
+
+        Object.values(this._styleSpec.variables).map(v => {
```
IagoLast:
```js
_getVariables().map(v => v._bind(metadata));
```

Jesus89:
# New API

I think this approach to enhance the expressions language has potential to solve some of the issues between Source/Style by grouping all into the new object Viz. This also enables to extract data using the expression language that could be used to create widgets. But having all-in-one: declaring scope for feature.columns, declaring variables, declaring aggregation parameters and style properties in the same object could be a little confusing for the user so we need to make sure that the final API does what we want it to do in a consistent and coherent way.

To discuss about the final spec I would like to work with these three use cases trying to get all possible uses of the API, the basic cases and the corner cases. I'm not totally agree with the presented API but I think its enough to discuss the cases and build on top of it.

I will use `$property` to reference actual columns/properties and `@variable` to reference viz variables.

# Use cases

## 1. Update style

> Edit a feature's style or all features' style

```js
const viz = new carto.Viz(`
  variables {
    color: rgb(255, 0, 0)
    disabledColor: opacity(@color, 0.5)
  }
  style {
    color: @disabledColor
    strokeColor: @color
  }
`);
```

OR

```js
const s = carto.style.expressions;
const viz = new carto.Viz({
  variables: {
    color: s.rgb(255, 0, 0),
    disabledColor: s.opacity(s.variable('color'), 0.5)
  },
  style: {
    color: s.variable('disabledColor'),
    strokeColor: s.variable('color')
  }
});
```

### Update feature's style on enter/leave

#### Style properties

```js
int.on('featureEnter', event => {
  for (let feature of event.features) {
    feature.style.color.blendTo(s.rgb(0, 255, 0));
    features.style.strokeColor.blendTo(s.rgb(0, 255, 0));
  }
});
int.on('featureLeave', event => {
  for (let feature of event.features) {
    feature.style.color.reset();
    features.style.strokeColor.reset();
  }
});
```

#### Variables

```js
int.on('featureEnter', event => {
  for (let feature of event.features) {
    feature.variables.color.blendTo(s.rgb(0, 255, 0));
  }
});
int.on('featureLeave', event => {
  for (let feature of event.features) {
    feature.variables.color.reset();
  }
});
```

### Update all features' style

#### Style properties

```js
viz.style.color.blendTo(s.rgb(0, 255, 0));
viz.style.strokeColor.blendTo(s.rgb(0, 255, 0));
viz.style.color.reset();
viz.style.strokeColor.reset();
```

#### Variables

```js
viz.variables.color.blendTo(s.rgb(0, 255, 0));
viz.variables.color.reset();
```

### Questions

- Can we edit style properties after creating the Viz object? YES
- Can we edit/add/remove variables after creating the Viz object? YES
- Can we use a variable in other variables? YES

## 2. Show columns and aggregated columns

> Click in a feature and show its name and average pop_max

```js
const viz = new carto.Viz(`
  columns [
    $name
  ]
  variables {
    avg_pop_max: avg($pop_max)
  }
  misc {
    resolution: 0.5
    placement: 'centroid'
  }
`);
```

OR

```js
const s = carto.style.expressions;
const viz = new carto.Viz({
  columns: [
    s.prop('name')
  ],
  variables: {
    avg_pop_max: s.avg(s.prop('pop_max'))
  },
  misc: {
    resolution: 0.5,
    placement: 'centroid'
  }
});
```

```js
const layer = new carto.Layer(source, viz);
const interactivity = new carto.Interactivity(layer);
interactivity.on('featureClick', event => {
  if (event.features) {
    show(event.features[0].columns.name);
    show(event.features[0].variables.avg_pop_max);
  }
});
```

### Questions

- Can we edit/add/remove columns after creating the Viz object?
- Are the columns used in the style, for example `pop_min` in `color: quantiles($pop_min, 5)` available in the feature columns? Or only the specified columns and variables are available?
- Are non-feature-dependent variables available in the feature? For example, `priceComparison: viewportPercentile($price, 1)`?
- Should we use better names for aggregation function `avg`, `sum`, `min`, `max`, `mode`? YES

### 3. Widgets with filters

> Create category widget to show the average(sum) of the most populated places shown in the map

```js
const viz = new carto.Viz(`
  variables {
    sum_pop_max: sum($pop_max)
    vavg: viewportAvg(@sum_pop_max)
  }
  style {
    width: sqrt(@sum_pop_max) / 1000
  }
  misc {
    filter: @sum_pop_max > 5000000
    resolution: 0.5
    placement: 'centroid'
  }
`);
```

OR

```js
const s = carto.style.expressions;
const viz = new carto.Viz({
  variables: {
    sum_pop_max: s.sum()
    avgp: s.viewportAvg(s.variable('sum_pop_max'))
  },
  style: {
    width: s.sqrt(s.variable('sum_pop_max')) / 1000
  },
  misc {
    filter: s.gt(s.variable('sum_pop_max'), 5000000),
    resolution: 0.5,
    placement: 'centroid'
  }
});
```

```js
const layer = new carto.Layer(source, viz);
layer.on('loaded' () => {
  createWidget({
    average: viz.variables.avgp.eval();
  })
});
```

### Questions

- Can we edit the filter after creating the Viz object?
- What happens if the user tries to evaluate a feature-dependent variable? Error
- What about Formula, Histogram and Time Series widgets? It would be nice to replicate them by using this API.

davidmanzanares:
> What about Formula, Histogram and Time Series widgets? It would be nice to replicate them by using this API.

I think Formula widgets can already be replicated with viewport* and global* functions, histogram can be workarounded with percentile functions, but additional support should be added. Time series is probably too far away right now.

src/core/renderer.js
```diff
@@ -160,7 +160,9 @@ class Renderer {
         const strokeColorRequirements = style.getStrokeColor()._getDrawMetadataRequirements();
         const strokeWidthRequirements = style.getStrokeWidth()._getDrawMetadataRequirements();
         const filterRequirements = style.getFilter()._getDrawMetadataRequirements();
-        let requiredColumns = [widthRequirements, colorRequirements, strokeColorRequirements, strokeWidthRequirements, filterRequirements]
+        const styleVariables = style._styleSpec.variables;
```
IagoLast: If you create the `getVariables` function you won't be accessing a private field here.

src/core/renderer.js
```diff
@@ -280,6 +282,10 @@ class Renderer {
 
         const drawMetadata = this._computeDrawMetadata(renderLayer);
 
+        Object.values(style._styleSpec.variables).map(v => {
```
IagoLast: `Object.values(style._styleSpec.variables)` --> `variables`

src/core/style/expressions/expression.js
```diff
@@ -35,6 +39,26 @@ export default class Expression {
         return this;
     }
 
+    _prefaceCode(glslCode){
```
IagoLast: `_prefaceCode` --> `_generatePrefaceCode` ?

IagoLast:
could we split the viz parameters?

```js
const viz = new carto.Viz({style, variables, columns, misc});
```

Every parameter could be expressed in string notation or as an object:


```js
const viz = new carto.Viz({
  style:  `
    color: red
    width: @sum_pop_max
 `,
  columns: `
    $price, $population
  `,
  variables: `
    sum_pop_max: sum($pop_max)
  `
});
```

The usage is pretty similar but is easier to parse

Jesus89:
# Viz API

After the meeting we have made the following decisions about the Text/JS APIs.
- Using `$pop_max` to reference actual column/property (SQL/GeoJSON). Access in JS with `s.prop('pop_max')` or `s.property('pop_max')`.
- Using `@my_color` to reference viz variables. Access in JS with `s.var('my_color')` or `s.variable('my_color')`.
- In the JS API we group all the variables in a `variables` object.
- Not using `style`, `misc` namespaces: all style properties are in the viz object:
  - `color`, `strokeColor`, `width`, `strokeWidth`, `filter`, `resolution`, `placement`.
- Style properties can be obtained from the viz object `viz.color`.
- Style properties can be updated, evaluated and reset: `blendTo(...)`, `get value / eval()`, `reset()`.
- Variables can be obtained from the viz `viz.variables.myVar` [or with a function `viz.getVar('myVar')`].
- Variables can be updated, evaluated and reset: `blenTo(...)`, `get value / eval()`, `reset()`.
- Variables can be used in other variables.
- Each feature contains a viz object with its style properties and the variables.
- We should use better naming for aggregation function `avg`, `sum`, `min`, `max`, `mode`:
  - `propertyAvg`, `clusterAvg`, ...
- If the user tries to evaluate a feature-dependent variable an `Error` will be thrown.

This is a big step, but we need to think about more use cases and verify if this makes sense in all the examples or we need to improve/change something.

I have updated the previous use cases with the new specification.

# Use cases

## 1. Update style

> Edit a feature's style or all features' style

```js
const viz = new carto.Viz(`
  @color: rgb(255, 0, 0)
  @disabledColor: opacity(@color, 0.5)

  color: @disabledColor
  strokeColor: @color
`);
```

OR

```js
const s = carto.style.expressions;
const viz = new carto.Viz({
  variables: {
    color: s.rgb(255, 0, 0),
    disabledColor: s.opacity(s.var('color'), 0.5)
  },
  color: s.var('disabledColor'),
  strokeColor: s.var('color')
});
```

### Update all features' style

#### Style properties

```js
viz.color.blendTo(s.rgb(0, 255, 0));
viz.strokeColor.blendTo(s.rgb(0, 255, 0));
viz.color.reset();
viz.strokeColor.reset();
```

#### Variables

```js
viz.variables.color.blendTo(s.rgb(0, 255, 0));
viz.variables.color.reset();
```

### Update feature's style on enter/leave

#### Style properties

```js
int.on('featureEnter', event => {
  for (let feature of event.features) {
    feature.viz.color.blendTo(s.rgb(0, 255, 0));
    features.viz.strokeColor.blendTo(s.rgb(0, 255, 0));
  }
});
int.on('featureLeave', event => {
  for (let feature of event.features) {
    feature.viz.color.reset();
    features.viz.strokeColor.reset();
  }
});
```

#### Variables

```js
int.on('featureEnter', event => {
  for (let feature of event.features) {
    feature.variables.color.blendTo(s.rgb(0, 255, 0));
  }
});
int.on('featureLeave', event => {
  for (let feature of event.features) {
    feature.variables.color.reset();
  }
});
```

## 2. Show columns and aggregated columns

> Click on a feature and show its name and average pop_max

```js
const viz = new carto.Viz(`
  @name: $name
  @avg_pop_max: avg($pop_max)

  resolution: 0.5
  placement: CENTROID
`);
```

OR

```js
const s = carto.style.expressions;
const viz = new carto.Viz({
  variables: {
    name: s.prop('name')
    avg_pop_max: s.avg(s.prop('pop_max'))
  },
  resolution: 0.5,
  placement: s.CENTROID
});
```

```js
const layer = new carto.Layer(source, viz);
const interactivity = new carto.Interactivity(layer);
interactivity.on('featureClick', event => {
  if (event.features) {
    show(event.features[0].viz.variables.name);
    show(event.features[0].viz.variables.avg_pop_max);
  }
});
```

## 3. Widgets with filters

> Create category widget to show the average(sum) of the most populated places shown on the map

```js
const viz = new carto.Viz(`
  @sum_pop_max: sum($pop_max)
  @vavg: viewportAvg(@sum_pop_max)

  width: sqrt(@sum_pop_max) / 1000
  filter: @sum_pop_max > 5000000
  resolution: 0.5
  placement: CENTROID
`);
```

OR

```js
const s = carto.style.expressions;
const viz = new carto.Viz({
  variables: {
    sum_pop_max: s.sum(s.prop('pop_max'))
    vavg: s.viewportAvg(s.var('sum_pop_max'))
  },
  width: s.sqrt(s.var('sum_pop_max')) / 1000,
  filter: s.gt(s.var('sum_pop_max'), 5000000),
  resolution: 0.5,
  placement: s.CENTROID
});
```

```js
const layer = new carto.Layer(source, viz);
layer.on('loaded' () => {
  createWidget({
    average: viz.variables.vavg.eval();
  })
});
```

davidmanzanares:
:+1: As I see it, this just needs the fixing of the interactivity tests and the interactivity examples (and the renaming / namespace change that will go outside this PR).

src/core/style/parser.js
```diff
@@ -39,13 +39,23 @@ function parseStyleNamedExpr(styleSpec, node) {
     if (node.operator != ':') {
         throw new Error('Invalid syntax');
     }
+    if (node.left.name.length && node.left.name[0] == '@') {
+        node.left.name = '__cartovl_variable_' + node.left.name.substr(1);
```
Jesus89: With the patch in `jsep` is this renaming required?

davidmanzanares: Yes, it is required because the variables are inserted, also, on the style root namespace to ease the implementation of interactivity.

PR #278, opened 2018-04-10
Fix agg as viewport/global/quantiles params

Jesus89:
This PR recovers some fixes implemented in the `source-columns-agg` branch that have value by itself out of the box:

- Fix using aggregations in viewport functions. Ex. `vieportMax(sum($prop))`
- Fix using aggregations in global functions. Ex. `globalAvg(sum($prop))`
- Fix using aggregations in quantiles function. Ex. `quantiles(avg($prop))` 
- Remove redundant `cartodb_id` property from `feature.properties` since it is the `feature.id`

### DIFF
--- a/src/core/style/expressions/aggregation.js
+++ b/src/core/style/expressions/aggregation.js
@@ -14,7 +14,7 @@ function genAggregationOp(aggName, aggType) {
     return class AggregationOperation extends Expression {
         constructor(property) {
             checkInstance(aggName, 'property', 0, Property, property);
-            super({ property: property });
+            super({ property });
             this._aggName = aggName;
             this.type = aggType;
         }
@@ -27,6 +27,9 @@ function genAggregationOp(aggName, aggType) {
         get numCategories() {
             return this.property.numCategories;
         }
+        eval(feature) {
+            return feature[schema.column.aggColumn(this.property.name, aggName)];
+        }
         //Override super methods, we don't want to let the property use the raw column, we must use the agg suffixed one
         _compile(metadata) {
             super._compile(metadata);
@@ -37,9 +40,6 @@ function genAggregationOp(aggName, aggType) {
                 preface: '',
                 inline: `${getGLSLforProperty(schema.column.aggColumn(this.property.name, aggName))}`
             };
-        }
-        eval(feature) {
-            return feature[schema.column.aggColumn(this.property.name, aggName)];
         }
         _postShaderCompile() { }
         _getMinimumNeededSchema() {
--- a/src/core/style/expressions/aggregation.js
+++ b/src/core/style/expressions/aggregation.js
@@ -21,6 +21,9 @@ function genAggregationOp(aggName, aggType) {
         get name() {
             return this.property.name;
         }
+        get aggName() {
+            return this._aggName;
+        }
         get numCategories() {
             return this.property.numCategories;
         }
--- a/src/core/style/expressions/quantiles.js
+++ b/src/core/style/expressions/quantiles.js
@@ -2,13 +2,14 @@ import Expression from './expression';
 import { float } from '../functions';
 import { checkNumber, checkInstance, checkType } from './utils';
 import Property from './property';
+import * as schema from '../../schema';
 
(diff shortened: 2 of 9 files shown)


src/core/style/expressions/quantiles.js
```diff
@@ -2,13 +2,14 @@ import Expression from './expression';
...
 import Property from './property';
+import * as schema from '../../schema';
 
 let quantilesUID = 0;
 
 function genQuantiles(global) {
     return class Quantiles extends Expression {
         constructor(input, buckets) {
-            checkInstance('quantiles', 'input', 0, Property, input);
+            checkInstance('quantiles', 'input', 0, Property, input && (input.property || input));
```
davidmanzanares: Clarify, pls

src/core/style/expressions/aggregation.js
```diff
@@ -14,16 +14,22 @@ function genAggregationOp(aggName, aggType) {
...
             checkInstance(aggName, 'property', 0, Property, property);
-            super({ property: property });
+            super({ property });
             this._aggName = aggName;
             this.type = aggType;
         }
         get name() {
             return this.property.name;
         }
+        get aggName() {
```
IagoLast: I don't think getters are a good idea...

test/integration/user/api/interactivity.test.js
```diff
@@ -70,7 +70,7 @@ describe('Interactivity', () => {
                     expect(event.features[0].style.color.blendTo).toBeDefined();
                     expect(event.features[0].style.color.reset).toBeDefined();
                     expect(event.features[0].style.reset).toBeDefined();
-                    expect(event.features[0]).toEqual({ id: 0, layerId: 'layer', properties: { cartodb_id: 0 }, style: jasmine.any(Object) });
+                    expect(event.features[0]).toEqual({ id: 0, layerId: 'layer', properties: { }, style: jasmine.any(Object) });
```
davidmanzanares: I wouldn't filter it

test/unit/api/style.test.js
```diff
@@ -165,11 +165,11 @@ describe('api/style', () => {
 
                 expect(actual).toEqual(jasmine.any(Style));
                 expect(actual.getResolution()).toEqual(1);
-                expect(actual.getColor()).toEqual(s.rgba(1, 0, 0, 1));
-                expect(actual.getWidth()).toEqual(s.float(10));
-                expect(actual.getStrokeColor()).toEqual(s.rgba(0, 0, 1, 1));
-                expect(actual.getStrokeWidth()).toEqual(s.float(15));
-                expect(actual.getOrder()).toEqual(s.asc(s.width()));
+                expect(actual.getColor().expr).toEqual(s.rgba(1, 0, 0, 1).expr);
```
davidmanzanares: I think that `eval()` would be the correct way to achieve this

src/core/style/expressions/aggregation.js
davidmanzanares: Disagree

Jesus89: Don't you like the getter? I think we should avoid accessing underscored variables

test/integration/user/api/interactivity.test.js
Jesus89: I think `cartodb_id`should not appear in properties because it is an internal way to achieve the feature.id. Since feature.id === feature.properties.cartodb_id we are duplicating always the information

davidmanzanares: I disagree, but we can always change it later if we really found it's better, I won't worry now :+1:

src/core/style/expressions/aggregation.js
IagoLast: I preffer `getAggregationName`

src/core/style/expressions/quantiles.js
Jesus89: This is added to allow aggregation as input parameters for quantiles. Ex `quantiles(sum($prop), 5)`. In this case we should check input.property instead of input.

Issue #28, opened 2018-02-06
Set up library distribution


davidmanzanares:
[UPDATED 2018-10-01]

Discussion
---
- Decide on versioning schema (eg. `0.0.0-beta-x`)
- What to do with mapbox-gl? (https://github.com/CartoDB/renderer-prototype/issues/242)
- Setup distribution / polyfills for the supported browsers ([babel-env](https://babeljs.io/docs/plugins/preset-env/))

Tasks
----
- [x] Create a npm package`@carto/cartovl`
- [x] Transpiling.
- [-] Polyfills (after discussing we are not adding polyfills for now)

We are ready to go now.

Jesus89:
Related to: https://github.com/CartoDB/renderer-prototype/issues/45

VictorVelarde:
I've made a quick review. Some notes:

#### Build system
We have a build based on **Webpack 4** with two options: `build:dev` and `build:min`
- build:dev creates 4 files:
  - `carto-vl.js` (1.1MB)
  - `e396fd0b112b6fc7099e.worker.js` (134KB)
  - `carto-vl.js.map` (1MB)
  - `e396fd0b112b6fc7099e.worker.js.map` (138KB)

- build:min creates 2 files:
  - `carto-vl.min.js` (**287KB**)
  - `8041724bec85fcf46240.worker.js` (41KB)

The code currently contains several modern features (>=ES6), such as `const`, `let`, `arrow functions`, `async/await`, `includes`, `fetch`, `export & import`...). Some of these are not fully supported by old browsers, such as IE11 (a useful tool to check those compatibilities is https://caniuse.com). 

For a broader compatibility we could apply a **babel** transpiling task.

##### Adding babel test
There are currently 3 active webpack loaders:
  * webpack-glsl-loader
  * svg-inline-loader
  * worker-loader

Adding a fourth one: 'babel-loader', with 'preset-env', results in (`build:min`):
- carto-vl.min.js (475KB vs previous 287KB)
- 8041724bec85fcf46240.worker.js (49KB vs previous 41KB)

Do we really want to transpile our library?  I'm not sure if it makes sense with our target.

VictorVelarde:
Pinging you for a quick opinion on 🔝 ... @rochoa @davidmanzanares @elenatorro @jgoizueta ?

rafaelfernandezgis:
[...]

PR #282, opened 2018-04-10
Rename package

IagoLast:
- Rename package (https://github.com/CartoDB/carto-vl/issues/28)
- Remove private flag ([only paid orgs can have private scope packages](https://docs.npmjs.com/misc/scope))
- Specify 0.0.0 version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
-  "private": true,
-  "name": "cartovl",
-  "version": "0.1.0",
+  "name": "@carto/carto-vl",
+  "version": "0.0.0-alpha-0",
   "description": "CARTO Vector library",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


Jesus89:
I should keep the `"private": true` to avoid publishing the package by accident until we want to do it.
We have already released the 0.1.0 version. I think the next version should be the 0.2.0. I don't understand why we go back to 0.0.0-alpha-0.

IagoLast:
> I should keep the "private": true to avoid publishing the package by accident

I want to publish the package so some teams can start using it internally

>  don't understand why we go back to 0.0.0-alpha-0.

I didnt know we released the 0.1.0, IMO we should start with a 0.0.0-apha-x naming without worring about semver (at least for the alpha)

rochoa:
Don't care much about keeping it `private`, that was important before agreeing in the name.

I prefer to stick to SemVer as it what our README says. I would go with 0.2.0 to make clear is something newer than renderer-prototype@0.1.0, even if it's a different name. It's like when someone decided to jump Node.js from 0.12.x to 4.0.0 because of io.js previous versions.

Issue #283, opened 2018-04-11
Error in retina display with click area


IagoLast:
Points:

[![https://gyazo.com/031dcd2a7df40cbf4b53471c597a1786](https://i.gyazo.com/031dcd2a7df40cbf4b53471c597a1786.gif)](https://gyazo.com/031dcd2a7df40cbf4b53471c597a1786)

Same for lines

[![https://gyazo.com/3394dd3bd0da94044d370ff9eafff1e6](https://i.gyazo.com/3394dd3bd0da94044d370ff9eafff1e6.gif)](https://gyazo.com/3394dd3bd0da94044d370ff9eafff1e6)

It gets interesting when you zoom in

[![https://gyazo.com/469ef5dd4ea419d3a441d058bc2172a4](https://i.gyazo.com/469ef5dd4ea419d3a441d058bc2172a4.gif)](https://gyazo.com/469ef5dd4ea419d3a441d058bc2172a4)

Polygons are working fine

TODO
---

- [x] Points

PR #284, opened 2018-04-11
283 interactivity errors

IagoLast:
Related with https://github.com/CartoDB/carto-vl/issues/283


davidmanzanares:
Works well :rocket:

Jesus89:
I think we should add more user tests to cover this code

Issue #298, opened 2018-04-16
Hex is broken


davidmanzanares:
https://cartodb.github.io/carto-vl/example/mapbox.html#eyJhIjoidHhfMDEyNV9jb3B5X2NvcHkiLCJiIjoiIiwiYyI6ImNhcnRvZ2wiLCJkIjoiaHR0cHM6Ly97dXNlcn0uY2FydG8uY29tIiwiZSI6IndpZHRoOiAzXG5jb2xvcjogaGV4KCdGMDAnKSIsImYiOnsibG5nIjoyLjE3LCJsYXQiOjQxLjM4fSwiZyI6MTN9

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/300

PR #300, opened 2018-04-16
Fix hex

Jesus89:
Related to: https://github.com/CartoDB/carto-vl/issues/298

- Fix Hex expression
- Add render test for Hex
- Fix fullscreen button

selection_132


example/mapbox.html
```diff
@@ -38,6 +38,7 @@
             min-height: 40px;
             height: auto;
             max-height: 400px;
+            font-family: monospace;
```
davidmanzanares: Great! a monospaced font is really needed, but, can we set a larger font size too?

Jesus89: Yes. Is 14px ok for you?

davidmanzanares: Thanks!

PR #306, opened 2018-04-17
Improve startup time

davidmanzanares:
Small change, but quite effective.

I checked 2 different examples and startup times were reduced: 3.8=> 2.1 and 4.5=>3.3

This improves the startup time since it avoids waiting for the context to request the metadata. Request metadata doesn't need the context, so the change should be safe.

We could even improve this by start requesting data without having the context, but in that case, we'll need to wait after the response since we need the context to upload the dataframes to the GPU.


davidmanzanares:
:smile:

PR #313, opened 2018-04-23
Move editor example

Jesus89:
Moving the editor example to `examples/editor/index.html`.

In order to allow redirection from the previous editor URL, the specific HTML in `gh-pages` should be replaced by:

```html
<html>
  <script language="javascript" type="text/javascript">
    var hash = window.location.hash;
    window.location='https://cartodb.github.io/carto-vl/examples/editor/index.html'+hash;
  </script>
</html>
```


Jesus89:
Done!

PR #311, opened 2018-04-18
Workaround WIP

davidmanzanares:

PR #315, opened 2018-04-23
Use vendor file from https://github.com/CartoDB/mapbox-gl-js/commit/4…

davidmanzanares:
Use vendor file from https://github.com/CartoDB/mapbox-gl-js/commit/4b58dc3ced816bb36199fc284ebe6f93e5f9e29c

Update MGL vendor file to get a fix for the same as https://github.com/CartoDB/carto-vl/pull/311

The problem was visible on the country borders of the basemap, particularly/usually in Africa, where many country borders disappeared.

The issue was that we were modifying one of the VAOs of MGL, and we weren't invalidating that. The patched version binds the default VAO to avoid modifying MGL VAOs

### DIFF
--- a/test/integration/user/api/interactivity.test.js
+++ b/test/integration/user/api/interactivity.test.js
@@ -63,7 +63,7 @@ describe('Interactivity', () => {
         });
     });
 
-    fdescribe('when the user clicks on the map', () => {
+    describe('when the user clicks on the map', () => {
         let interactivity;
 
         describe('and the click is in a feature', () => {
@@ -128,8 +128,8 @@ describe('Interactivity', () => {
                 });
                 layer.on('loaded', () => {
                     // Move mouse inside a feature
-                    map.fire('click', { lngLat: { lng: -10, lat: -10 } });
-                    map.fire('mousemove', { lngLat: { lng: 10, lat: 10 } });
+                    map.fire({ type: 'click', lngLat: { lng: -10, lat: -10 } });
+                    map.fire({ type: 'mousemove', lngLat: { lng: 10, lat: 10 } });
                 });
             });
 
@@ -142,7 +142,7 @@ describe('Interactivity', () => {
                 });
                 layer.on('loaded', () => {
                     // Move mouse inside a feature
-                    map.fire('mousemove', { lngLat: { lng: 10, lat: 10 } });
+                    map.fire({ type: 'mousemove', lngLat: { lng: 10, lat: 10 } });
                 });
             });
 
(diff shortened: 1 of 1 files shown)


davidmanzanares:
Is this ready to merge? @IagoLast

Jesus89:
- I would use our fork as a `npm` dependency
- Are we covering the same use cases than before?
- Why are you using `click()` and `map.fire({ type`. Could we use the same function for all the user tests?

IagoLast:
@Jesus89 

> I would use our fork as a npm dependency

+1 (different issue)

> Are we covering the same use cases than before?

Nope, less cases IMO we should try to keep all cases even when we need to use mapboxgl private methods.

> Why are you using click() and map.fire({ type. Could we use the same function for all the user tests?

As pointed [here](https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.45.0-beta.1) the `map.fire` functionality was moved to be private (but still there with a slightly different API) Some events can be simulated but other ones are very hard to.  Since api changes are very easy to update I think we should rewrite the old tests (keeping them all) using the new API and forget about this.

davidmanzanares:
Hey, I don't like the idea of using `map.fire`. That's a private method of MGL. We shouldn't depend on that, MGL patch versions and major changes with breaking changes notes can break our tests with no reason.

(https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.45.0-beta.1)

IagoLast:
> Hey, I don't like the idea of using map.fire. That's a private method of MGL. 

Agreed BUT:

- The right fix this will take lot of effort and time while the hotfix is just a few minutes.
- We only need to check this when we change the `mapbox-gl` version, this happened 1 time in 2 months and I think we could find worst problems than that when this change happens.
- Even if the change breaks this, Is only test code so our product won't break.

So right, is not the perfect solution but I don't think is a good idea to waste time working on a shinny fix for a few tests while we have thousands of hacks in the real code.

davidmanzanares:
> I think we could find worst problems than that when this change happens.

It shouldn't, if it happens it is a bug.

IagoLast:
> It shouldn't, if it happens it is a bug.

I mean, since we are using `mapbox-gl` through a custom fork it could happen they change some internal and break our project right?

IagoLast:
oh, nice @Jesus89 ;)

Jesus89:
We close this PR since we have solved this issue in #332 and #339

IagoLast:
should we delete the branch?

PR #331, opened 2018-04-24
Compute metadata based on viz to reduce properties

### LINKED ISSUE #317: GeoJSON improvements
- [x] Performance when having several categories
  - [x] Property lookup is slow due to linear search.
  - [x] Only use required columns based on `viz.getMinimumNeededSchema` as in Windshaft.
- [x] Smell: `requestMetadata` doesn't use `viz`.
- [x] JSON.parse/JSON.stringify for cloning the orignal data.
- [x] If present, it should use cartodb_id from original data. Use case: using SQL API to retrieve GeoJSON.

rochoa:
The DataFrame is computed on demand, without having to precompute features
from the original data. This removes the clone of the data. It also allows using the
cartodb_id from original data.

This fixes #317.


rochoa:
@Jesus89, using negative cartodb_id was on purpose. How was that affecting the tests?

Jesus89:
OK. If this is a feature we should update the interactivity tests. Let me do it

rochoa:
I updated this PR with @davidmanzanares' one.

Jesus89:
:+1:

Issue #336, opened 2018-04-24
Tests are broken on Retina displays


davidmanzanares:
As reported by @rochoa 

Could you take a look @IagoLast ?

IagoLast:
3 Render tests are broken:


|Test| actual|expected|diff|
| --- |:-----:|:---------:|:---:|
| color/named|![reference_out](https://user-images.githubusercontent.com/2657897/39191196-b93be130-47d6-11e8-99fb-11d7d710d358.png)|![reference](https://user-images.githubusercontent.com/2657897/39191212-c3fffc82-47d6-11e8-8b3c-e039873cc196.png)|![diff](https://user-images.githubusercontent.com/2657897/39191232-cbdf8c10-47d6-11e8-83b4-8a094d1e3760.png)
| ramp/custom-category-others |![reference_out](https://user-images.githubusercontent.com/2657897/39191319-fedab040-47d6-11e8-945a-890b19eaeae9.png)|![reference](https://user-images.githubusercontent.com/2657897/39191334-048f0f5e-47d7-11e8-86b1-14ee235b4037.png)|![diff](https://user-images.githubusercontent.com/2657897/39191376-1de962c4-47d7-11e8-8613-ddb455e4e256.png)
|ramp/custompalette|![reference_out](https://user-images.githubusercontent.com/2657897/39191459-45bbd3e0-47d7-11e8-8946-0369b59dc0d1.png)|![reference](https://user-images.githubusercontent.com/2657897/39191474-4bd43cd6-47d7-11e8-9c36-ed6ad88421d0.png)|![diff](https://user-images.githubusercontent.com/2657897/39191478-4f11c224-47d7-11e8-85cf-340773ecf683.png)

rochoa:
Running the tests in my retina laptop I get:

```
  Render tests:
    1) /collection-point/color
    ✓ /collection-point/default (1445ms)
    2) /collection-point/strokeColor
    3) /collection-point/strokeWidth
    4) /collection-point/width
    ✓ /color/cielab (1457ms)
    5) /color/hex
    ✓ /color/hsl (1498ms)
    ✓ /color/hsla (1477ms)
    ✓ /color/hsv (1480ms)
    ✓ /color/hsva (1479ms)
    ✓ /color/named (1494ms)
    6) /color/opacity
    7) /color/rgb
    ✓ /color/rgba (1487ms)
    8) /line-string/color
    ✓ /line-string/default (1520ms)
[...]

PR #338, opened 2018-04-24
Add render threshold

IagoLast:
Fixes https://github.com/CartoDB/carto-vl/issues/336


Jesus89:
I'm not very happy with this threshold adjustment. What is the real reason for the issue? Retina, the color space?

Jesus89:
Try the same: configuring your color space to Display P3 or Color LCD to SD 170M-A

PR #341, opened 2018-04-25
Improve editor load

Jesus89:
- Improve load for the maps (remove initial basemap and position)
- Fix load when using incognito mode
- `blendToViz` returns a Promise

load-improved

### DIFF
--- a/examples/editor/index.js
+++ b/examples/editor/index.js
@@ -120,17 +120,12 @@ map.on('load', () => {
             if (layer) {
                 $('#loader').addClass('spin');
                 document.getElementById('feedback').style.display = 'none';
-                const promise = layer.blendToViz(new carto.Viz(v));
-                if (promise) {
-                    promise.then(() => {
-                        $('#loader').removeClass('spin');
-                    }).catch(error => {
-                        handleError(error);
-                        $('#loader').removeClass('spin');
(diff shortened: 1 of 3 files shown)


examples/editor/index.js
```diff
@@ -220,8 +211,16 @@ map.on('load', () => {
     } else {
         barcelona();
     }
+
+    map.on('moveend', () => {
```
davidmanzanares: No need for the extra arrow function, right?

test/unit/api/layer.test.js
```diff
@@ -97,30 +97,33 @@ describe('api/layer', () => {
     });
 
     describe('.blendToViz', () => {
-        it('should not throw an error if a valid viz is passed', () => {
+        it('should resolve the promise if a valid viz is passed', (done) => {
             const layer = new Layer('layer0', source, viz);
-            expect(function () {
-                layer.blendToViz(viz2);
-            }).not.toThrow();
+            // Mock for _vizChanged
```
davidmanzanares: This is testing the implementation

Jesus89: This is actually testing that `_checkId` passes without errors. So `_vizChanged` should be mocked.

PR #356, opened 2018-04-30
Polygon stroke

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/pull/356
- [x] With GeoJSON
- [x]  Support for polygon holes
- [x] With Windshaft  without clipping artifacts
- [x] Interactivity should take stroke into account
- [x] Fix tests
- [x] Test new functionality
- [x] Docs
    - [x] One example
- [x] Take last vertex on polygon holes into account
- [x] Take multi-clipping case into account

Probably for another PR:
- [ ] Use correct joint normal
- [ ] Investigate where and how the additional client-side clipping must be done


src/core/dataframe.js
```diff
@@ -159,30 +159,43 @@ export default class Dataframe {
...
         const p = wToR(pos.x, pos.y, { center: this.center, scale: this.scale });
         const vertices = this.decodedGeom.vertices;
+        const normals = this.decodedGeom.normals;
         const breakpoints = this.decodedGeom.breakpoints;
         let featureIndex = 0;
         const features = [];
+        // The viewport is in the [-1,1] range (on Y axis), therefore a pixel is equal to the range size (2) divided by the viewport height in pixels
+        const widthScale = (2 / this.renderer.gl.canvas.clientHeight) / this.scale * this.renderer._zoom;
+        const columnNames = Object.keys(this.properties);
+        const vizWidth = viz.getWidth();
```
Jesus89: `getStrokeWidth()`?

davidmanzanares: Ups, really good catch!

Issue #358, opened 2018-04-30
Add JS/String examples in JSDoc


Jesus89:
The goal is to define a way to include both JS/String APIs in the inline examples. These examples will be accessible in the Developer Center documentation by using the tabs (String, JavaScript).

For example, we can add the following documentation and then split the content in the dev-center side. (We need to define the interface):

```js
/*
 * @example <caption>String</caption>
 * const viz = new carto.Viz(`
 *   width: 5 * 5
 * `);
 * @example <caption>JavaScript</caption>
 * const s = carto.expressions;
 * const viz = new carto.Viz({
 *   width: s.mul(5, 5);
 * });
 */
```

Or

```js
/*
 * @example
 * const viz = new carto.Viz(`
 *   width: 5 * 5
 * `);
 * ---
 * const s = carto.expressions;
 * const viz = new carto.Viz({
 *   width: s.mul(5, 5);
 * });
 */
```

Etc...

The final result will be something like:
- String (tab 1)
```js
const viz = new carto.Viz(`
  width: 5 * 5
`);
```
- JavaScript (tab 2)
```js
const s = carto.expressions;
const viz = new carto.Viz({
  width: s.mul(5, 5);
});
```

@csubira @arianaescobar

Jesus89:
```
@example <caption>String: Using a color scheme</caption>
const viz = new carto.Viz(`
  filter: ramp($type, PRISM);
`);

@example <caption>JavaScript: Using a color scheme</caption>
const s = carto.expressions;
const viz = new carto.Viz({
  filter: s.ramp(s.prop('type'), s.palettes.PRISM);
});
```

Jesus89:
Closed by 371

PR #371, opened 2018-05-04
Inline doc examples

Jesus89:
Related to: https://github.com/CartoDB/carto-vl/issues/358

Changes:

- Use an array for `in`, `nin` functions instead of ...arg.
- Use an array for `buckets` function instead of ...arg.
- Add String API examples to all the expressions
- Add `PI` constant to expressions lang


src/core/viz/expressions/belongs.js
```diff
@@ -59,10 +69,14 @@ function NIN_INLINE_MAKER(categories) {
...
     return class BelongExpression extends BaseExpression {
-        constructor(value, ...categories) {
+        constructor(value, categories) {
             value = implicitCast(value);
             checkExpression(name, 'value', 0, value);
 
+            categories = categories || [];
+            if (!Array.isArray(categories)) {
+                categories = [categories];
+            }
```
jgoizueta: So... we admit `s.in(s.prop('type'), 'metropolis')` for `s.in(s.prop('type'), ['metropolis'])`, right? Which is nice for the user fingers... but will be allow it also in the String language? ...in all functions that take arrays? ...is it worth the effort?

src/core/viz/expressions/binary.js
```diff
@@ -189,12 +252,21 @@ export const GreaterThan = genBinaryOp('greaterThan',
  * @param {carto.expressions.Base|number} y - Numeric expression
  * @return {carto.expressions.Base} Numeric expression
  *
- * @example <caption>Compare two numbers to show only elements with price >= 30</caption>
+ * @example <caption>Compare two numbers to show only elements with price greater or equal than 30.</caption>
```
jgoizueta: grammar police: /than/to

src/core/viz/expressions/belongs.js
Jesus89: Good point! I added this code because of the tests, but the point is that we must use always [] for all these cases, so we will use arrays also in the String API.

src/core/viz/expressions/binary.js
```diff
@@ -189,12 +252,21 @@ export const GreaterThan = genBinaryOp('greaterThan',
...
  * const s = carto.expressions;
  * const viz = new carto.Viz({
- *   filter: s.gte(s.prop('price'), 30);
+ *   filter: s.gte(s.prop('price'), 30)
  * });
  *
+ * @example <caption>Compare two numbers to show only elements with price greater or equal than 30. (String)</caption>
```
jgoizueta: same, /than/to

src/core/viz/expressions/binary.js
```diff
@@ -215,12 +287,21 @@ export const GreaterThanOrEqualTo = genBinaryOp('greaterThanOrEqualTo',
  * @param {carto.expressions.Base|number} y - Numeric expression
  * @return {carto.expressions.Base} Numeric expression
  *
- * @example <caption>Compare two numbers to show only elements with price < 30</caption>
+ * @example <caption>Compare two numbers to show only elements with price lower than 30.</caption>
```
jgoizueta: grammar police again: is lower OK? (usually "less" is used)

jgoizueta: mmm 🤔 for "price" it sounds right, maybe the grammar police was too zealous here

src/core/viz/expressions/binary.js
```diff
@@ -241,12 +322,21 @@ export const LessThan = genBinaryOp('lessThan',
  * @param {carto.expressions.Base|number} y - Numeric expression
  * @return {carto.expressions.Base} Numeric expression
  *
- * @example <caption>Compare two numbers to show only elements with price <= 30</caption>
+ * @example <caption>Compare two numbers to show only elements with price lower or equal than 30.</caption>
```
jgoizueta: 👮 "lower than or equal to" or perhaps "less than or equal to"

src/core/viz/expressions/binary.js
```diff
@@ -241,12 +322,21 @@ export const LessThan = genBinaryOp('lessThan',
...
  * const s = carto.expressions;
  * const viz = new carto.Viz({
- *   filter: s.lte(s.prop('price'), 30);
+ *   filter: s.lte(s.prop('price'), 30)
  * });
  *
+ * @example <caption>Compare two numbers to show only elements with price lower or equal than 30. (String)</caption>
```
jgoizueta: 👮 again

src/core/viz/expressions/binary.js
```diff
@@ -267,12 +357,21 @@ export const LessThanOrEqualTo = genBinaryOp('lessThanOrEqualTo',
  * @param {carto.expressions.Base|number} y - Numeric expression
  * @return {carto.expressions.Base} Numeric expression
  *
- * @example <caption>Compare two numbers to show only elements with price === 30</caption>
+ * @example <caption>Compare two numbers to show only elements with price equal than 30.</caption>
```
jgoizueta: 👮

src/core/viz/expressions/binary.js
```diff
@@ -267,12 +357,21 @@ export const LessThanOrEqualTo = genBinaryOp('lessThanOrEqualTo',
...
  * const s = carto.expressions;
  * const viz = new carto.Viz({
- *   filter: s.eq(s.prop('price'), 30);
+ *   filter: s.eq(s.prop('price'), 30)
  * });
  *
+ * @example <caption>Compare two numbers to show only elements with price equal than 30. (String)</caption>
```
jgoizueta: 👮

src/core/viz/expressions/binary.js
```diff
@@ -293,12 +392,21 @@ export const Equals = genBinaryOp('equals',
  * @param {carto.expressions.Base|number} y - Numeric expression
  * @return {carto.expressions.Base} Numeric expression
  *
- * @example <caption>Compare two numbers to show only elements with price !== 30</caption>
+ * @example <caption>Compare two numbers to show only elements with price not equal than 30.</caption>
```
jgoizueta: 👮

src/core/viz/expressions/binary.js
```diff
@@ -293,12 +392,21 @@ export const Equals = genBinaryOp('equals',
...
  * const s = carto.expressions;
  * const viz = new carto.Viz({
  *   filter: s.neq(s.prop('price'), 30);
  * });
  *
+ * @example <caption>Compare two numbers to show only elements with price not equal than 30. (String)</caption>
```
jgoizueta: 👮

src/core/viz/expressions/clusterAggregation.js
```diff
@@ -64,18 +76,22 @@ export const ClusterMax = genAggregationOp('max', 'number');
 export const ClusterMin = genAggregationOp('min', 'number');
 
 /**
- * Aggregate using the maximum value. This operation disables the access to the property
+ * Aggregate using the mode value. This operation disables the access to the property
```
jgoizueta: Maybe we should add (here and in all the other cluster functions): "except within other cluster aggregate functions" or something similar?

src/core/viz/expressions/ramp.js
```diff
@@ -12,30 +13,28 @@ import { cielabToSRGB, sRGBToCielab } from '../colorspaces';
...
 *
-* @example <caption> Display points with a different color depending on the `category` property. (We assume category has discrete values) </caption>
+* @example <caption>Display points with a different color depending on the `category` property. (We assume category has discrete values).</caption>
 * const s = carto.expressions;
 * const viz = new carto.Viz({
-*   color: s.ramp(s.prop('category'), s.palettes.PRISM),
+*   color: s.ramp(s.prop('category'), s.palettes.PRISM)
 * });
 *
-* @example <caption> Display points with a different color depending on the `speed` property. (We assume category has continuos numeric values)</caption>
```
jgoizueta: Why have you remove this example?
By the way: 👮 /continuos/continuous

src/core/viz/expressions/clusterAggregation.js
Jesus89: :+1:

src/core/viz/expressions/ramp.js
Jesus89: This example adds nothing new. The same as before but changing the column's name.

Issue #370, opened 2018-05-04
Support comments in Viz


Jesus89:
Just ignore comments inside viz lang:

- [x] `// line-comment`
- [x] `/* block-comment */`

Jesus89:
Closed by #372

PR #372, opened 2018-05-04
Support String API comments

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/370

- Clean one-line comment in viz string
- Clean multi-line-comments in viz string
- Add unit tests

### DIFF
--- a/src/core/viz/parser.js
+++ b/src/core/viz/parser.js
@@ -25,7 +25,7 @@ export function parseVizExpression(str) {
 
 export function parseVizDefinition(str) {
     prepareJsep();
-    const ast = jsep(str);
+    const ast = jsep(cleanComments(str));
     let vizSpec = { variables: {} };
     if (ast.type == 'Compound') {
         ast.body.map(node => parseVizNamedExpr(vizSpec, node));
@@ -178,4 +178,13 @@ function cleanJsep() {
     jsep.removeIdentifierChar('#');
     jsep.addLiteral('true');
     jsep.addLiteral('false');
+}
+
+/**
+ * Remove comments from string
+ * - // one line comments
+ */
+function cleanComments(str) {
+    return str
+        .replace(/\/\/.*/g, '');
 }
--- a/src/core/viz/parser.js
+++ b/src/core/viz/parser.js
@@ -182,9 +182,11 @@ function cleanJsep() {
 
 /**
  * Remove comments from string
- * - // one line comments
- */
+ * - // one-line comments
+ * - /* multi-line
+ *     comments */
 function cleanComments(str) {
     return str
-        .replace(/\/\/.*/g, '');
+        .replace(/\/\/.*/g, '')
+        .replace(/\/\*(\*(?!\/)|[^*])*\*\//g, '');
 }
--- a/src/core/viz/parser.js
+++ b/src/core/viz/parser.js
@@ -184,7 +184,7 @@ function cleanJsep() {
  * Remove comments from string
  * - // line comments
  * - /* block comments
- * - ' " Keep comments inside single and double quotes
+ * - Keep comments inside single and double quotes tracking escape chars
  * Based on: https://j11y.io/javascript/removing-comments-in-javascript/
  */
 export function cleanComments(str) {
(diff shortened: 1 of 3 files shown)


src/core/viz/parser.js
```diff
@@ -179,3 +179,14 @@ function cleanJsep() {
...
+/**
+ * Remove comments from string
+ * - // one-line comments
+ * - /* multi-line
+ *     comments */
+export function cleanComments(str) {
+    return str
+        .replace(/\/\/.*/g, '')
+        .replace(/\/\*(\*(?!\/)|[^*])*\*\//g, '');
+}
```
jgoizueta: Mmmm, what about "//" and "/*" inside quotes?

Jesus89: I see. We need stronger regex :muscle:

Jesus89:
Hi @jgoizueta. I have updated the cleaning function, using a new (more efficient) algorithm that takes into account the content of the quotes.

src/core/viz/parser.js
```diff
@@ -179,3 +179,78 @@ function cleanJsep() {
...
+ * - ' " Keep comments inside single and double quotes
+ * Based on: https://j11y.io/javascript/removing-comments-in-javascript/
+ */
+export function cleanComments(str) {
+    var mode = {
+        singleQuote: false,
+        doubleQuote: false,
+        blockComment: false,
+        lineComment: false
+    };
```
jgoizueta: 🤔 So this is basically a state machine but you keep track of the state with a boolean per state...  so you do `if (mode.singleQuote)` instead of `mode == SINGLEQUOTE`. Am I missing anything?

test/unit/core/viz/parser.test.js
```diff
@@ -1,12 +1,58 @@
...
+            const str = 'width: 1// line comment\n';
+            expect(cleanComments(str)).toBe('width: 1\n');
+        });
+
+        it('should keep line comments inside simple quotes', () => {
+            const str = '@var: \'// line comment\'';
+            expect(cleanComments(str)).toBe('@var: \'// line comment\'');
+        });
+
+        it('should keep line comments inside simple quotes', () => {
```
jgoizueta: typo /simple/double

src/core/viz/parser.js
```diff
@@ -179,3 +179,78 @@ function cleanJsep() {
...
+
+    // Adding chars to avoid index checking
+    str = ('_' + str + '_').split('');
```
jgoizueta: What's the reason for splitting and working with an array of one-char strings`str=['1', '2', '3']` instead of a string `str='123'` ?. In both cases we can index individual characters as `str[i]`, right?

src/core/viz/parser.js
```diff
@@ -179,3 +179,78 @@ function cleanJsep() {
...
+            str[i] = '';
+            continue;
+        }
+
+        if (mode.lineComment) {
+            if (str[i+1] === '\n' || str[i+1] === '\r') {
+                mode.lineComment = false;
+            }
+            if (i+1 < l) {
+                str[i] = '';
```
jgoizueta: Oh, maybe that's the reason for the split?

src/core/viz/parser.js
jgoizueta: Now I see some characters get eliminated in the loop with `str[i]=''`

Jesus89: It is because of string [i] is read-only.

src/core/viz/parser.js
Jesus89: Yep ^

src/core/viz/parser.js
```diff
@@ -179,3 +179,78 @@ function cleanJsep() {
...
+        lineComment: false
+    };
+
+    // Adding chars to avoid index checking
+    str = ('_' + str + '_').split('');
+
+    for (var i = 0, l = str.length; i < l; i++) {
+
+        if (mode.singleQuote) {
+            if (str[i] === '\'' && str[i-1] !== '\\') {
```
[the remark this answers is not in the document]
Jesus89: Good point!

Jesus89:
I have added the code + tests for the escape char cases

Issue #63, opened 2018-02-13
Examples for beta release


rochoa:
TBD with @inigomedina.

makella:
@davidmanzanares @rochoa after going back and forth, back and forth, I started a different style today to document some basic examples.

I will meet with Cillas tomorrow to discuss further details.

I'm still not done (because I couldn't decide what to do) and what I am putting together needs work still, but at a basic level, you can see some example maps (just screenshots for now), basic description, and associated styling. Will continue chipping away at this the remainder of the week to make it more robust and cover all the "basics"

https://github.com/CartoDB/renderer-prototype/wiki/Map-Examples

makella:
@rochoa @davidmanzanares 

I think I've gotten through the point map types...

Right now, the biggest things we are missing are additional classification methods and classed bubble maps for points

I put some resources together please have a look. 

_As a side note, these are by no means "final docs" and as you will notice are still incomplete in some areas_

+ [Styling and Operator Tables](https://github.com/CartoDB/renderer-prototype/wiki/Styling-and-Operator-Tables)
+ [Styling Points](https://github.com/CartoDB/renderer-prototype/wiki/Styling-Points)

arianaescobar:
@makella and I were talking about the examples yesterday, trying to make sense about what extra work we might need for the beta release and I just created [this document](https://docs.google.com/spreadsheets/d/1hpoA1ysvU-5JZt6lLNnstetQ0GlkhHSOZR55N6n5DNc/edit?usp=sharing) that will help us to quickly visualize what's needed.

makella:
❤️ this!

arianaescobar:
To discuss: for the examples, should we use javascript or the string API?

makella:
We have been using string, so it was easier for me to modify, I'd say let's stick with that! @Jesus89 cool with you?

arianaescobar:
Yeah, Jesus and I agree on using string, but there are a few using javascript. We should change those then.

arianaescobar:
To-do: 

- [ ] Add spinner to examples

makella:
👍

PR #380, opened 2018-05-09
63 Examples update for beta release

arianaescobar:
Fixes: #63


arianaescobar:
Hi @Jesus89,

Mamata and I are pickier than you. 
Feel free to just merge.

Love, 
The walking dead versions of Mamata and Ariana

Jesus89:
I see some conflicts... Let's fix them.

And what about the To-do:
  - [ ] Add spinner to examples
  - [ ] CSS file for UI components (@piensaenpixel)
  - [ ] @Jesus89 to review examples to remove unnecessary stuff

(This is part of my review :stuck_out_tongue: )

Jesus89:
Review (after merging master)
- animate URL is not correct
- viewport example generates an error in the shader
- zoom example has not the proper zoom
- interact-multiple-layers `viz.getWidth` error
- interactive-based-styling `viz.getWidth` error
- pop-ups  `viz.getWidth` error
- denver-trees `viz.getWidth` error
- spen-data `customPalette(): invalid fourth parameter color[3]`
- us-population-by-country fails in the shader
- taxi-pickups `Unimplemented globalAggregation` `Cannot read program of undefined` (renderer)

arianaescobar:
Thank you for the review!

Can you help us out with those things you found? 🙏🏻☹️

examples/interactivity/pop-ups.html
```diff
@@ -125,52 +125,14 @@
       strokeColor: opacity(BlueViolet, 0.8)
     `);
     const layer = new carto.Layer('layer', source, viz);
-
```
arianaescobar: Wait, why?
We did this on purpose.

Jesus89: Ah, I thought we want unit-examples without extra functionality. This example looks exactly like the interactive-style plus pop-ups. As a developer, I would appreciate small and clean examples.

But, despite this, if you want to add this I could revert the commit.

arianaescobar: Yup, that's definitely the idea. But since we already introduced the interactivity based styling on the previous example, we thought it wouldn't be a big deal keeping it on the following one. Also, because this is something that a lot of people will want to make, so better giving it to them at once :)

Jesus89: OK. It makes sense :+1:

PR #383, opened 2018-05-09
Fix properties initialized to 0

Jesus89:
Property check for undefined properties.
Fixes using `strokeWidth: 0` in the JS API


Jesus89:
Using the util functions to improve readability

PR #385, opened 2018-05-09
Viewport histogram

davidmanzanares:
- [x] Support categories
- [x] Support numerics
- [x] Tests
- [x] Docs
- [x] Example

To be done later:
- [ ] Support nested histograms like: for each month get the distribution (histogram) of the loss cause.


examples/expressions/histogram.html
```diff
@@ -0,0 +1,62 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>Single layer | CARTO</title>
```
Jesus89: Histogram

davidmanzanares: Ops, fixed

Issue #384, opened 2018-05-09
resolution lack of blendTo method


davidmanzanares:
As reported by @rubenmoya we lack a method to change the resolution without doing a `blendToViz`. This is highly inconsistent since all the other properties have a `blendTo` method.

The reason for this behavior is that `resolution` only allows numbers, not even expressions.

I think we should make `resolution` a first-class citizen and allow numeric expressions there, with the obvious constraint of being feature independent, including metadata operators. This kind of limitation is not so unusual on the viz lang, but the "just a number" limitation is.

davidmanzanares:
No need to fix after https://github.com/CartoDB/carto-vl/pull/393

PR #393, opened 2018-05-10
Allow setting the resolution directly

davidmanzanares:
Previously, the only way to change the resolution was to use `layer.blendToViz`. This allows to set it directly with `viz.resolution = 5`.

I thought about using `blendTo`, but it didn't make much sense since the resolution cannot be animated, so I decided to stick to the current way, but improve it.

Address: https://github.com/CartoDB/carto-vl/issues/384


Jesus89:
I think a test for that would be nice :)

PR #396, opened 2018-05-11
Update event

davidmanzanares:
Adds a new `update` event which is fired every time that the map is moved, the viz changes (including animation) or new data is loaded. This is very practical for reacting to render external widgets.

cc: @rubenmoya

### DIFF
--- a/src/api/layer.js
+++ b/src/api/layer.js
@@ -220,6 +220,7 @@ export default class Layer {
             return;
         }
         this._source.requestData(this._getViewport());
+        this._needUpdate = true;
     }
 
     hasDataframes() {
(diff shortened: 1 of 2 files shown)


Jesus89:
+Tests please ;)

davidmanzanares:
How do you imagine testing this? I mean, you would need to know when dataframes are added, when map is moved... I don't want to create a lot of fake internals and test the implementation

Jesus89:
I'm not very happy using a generic event for lots of things, so we should revisit this in the future. But since its part of the API it should have tests to cover all its functionality.

To tackle this we must implement "user tests" to work with a real map (div). Let me help you with that ;)

Jesus89:
UPDATE: added user tests and a bit of refactoring/renaming.

LGTM, but @davidmanzanares take a look please :wink:

davidmanzanares:
:+1: Thanks!

PR #399, opened 2018-05-11
Get all metadata through Maps API

jgoizueta:
WIP
This removes all usage of the SQL API.
It requires https://github.com/CartoDB/Windshaft-cartodb/pull/952


src/client/windshaft.js
```diff
@@ -738,6 +600,40 @@ function authURL(url, conf) {
...
+function adaptColumnType(type) {
+    if (type === 'string') {
+        return 'category';
+    }
+    return type;
+}
+
+// generate a promise under certain assumptions/choices; then if the result changes the assumptions,
+// repeat the generation with the new information
+async function repeatablePromise(initialAssumptions, assumptionsFromResult, promiseGenerator) {
```
davidmanzanares: I don't like much this part, but if it's going to be temporal, ok

src/client/windshaft.js
```diff
@@ -56,11 +53,12 @@ export default class Windshaft {
...
     }
 
-    _getInstantiationID(MNS, resolution, filtering) {
+    _getInstantiationID(MNS, resolution, filtering, choices) {
         return JSON.stringify({
             MNS,
             resolution,
-            filtering: this.metadata && this.metadata.featureCount > MIN_FILTERING ? filtering : null
+            filtering: choices.backendFilters ? filtering : null,
+            choices: choices
```
davidmanzanares: `options` wouldn't be better?

src/client/windshaft.js
```diff
@@ -132,6 +130,9 @@ export default class Windshaft {
      *  - When the filter conditions changed and the dataset should be server-filtered.
      */
     _needToInstantiate(MNS, resolution, filtering) {
+        if (!this.metadata) {
+            return true;
```
davidmanzanares: Isn't this redundant? I guess that `this._MNS` will always be undefined the first time, therefore, the next check should return true every time this condition happens, right?

Issue #403, opened 2018-05-11
Rename public method `eval` to a `.value` getter


davidmanzanares:
Right now to access to data-view like data like histograms, global and viewport aggs we do:
`const max = viz.variables.globalMaxLoss.eval()`
I think it would be more consistent to have:
`const max = viz.variables.globalMaxLoss.value`

It would be similar to:
`const featureMax = feature.variables.featureMaxLoss.value`

davidmanzanares:
Fixed by #421

IagoLast:
Not fixed, this was a preliminary refactor 😅

PR #425, opened 2018-05-16
403 eval to value

IagoLast:
#403 
---

Note for the reviewer:
--- 
I have no idea of what I'm doing


Jesus89:
Found that this breaks point/lines detection :stuck_out_tongue:

Already hot-fixed in master.

PR #427, opened 2018-05-17
Use cartovl user instead of cartogl

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/366

### DIFF
--- a/src/api/integrator/mapbox-gl.js
+++ b/src/api/integrator/mapbox-gl.js
@@ -62,7 +62,7 @@ class MGLIntegrator {
     }
 
     addLayer(layer, beforeLayerID) {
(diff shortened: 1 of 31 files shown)


src/api/integrator/mapbox-gl.js
```diff
@@ -62,7 +62,7 @@ class MGLIntegrator {
     }
 
     addLayer(layer, beforeLayerID) {
-        const callbackID = `_cartoGL_${uid++}`;
+        const callbackID = `_cartovl_${uid++}`;
```
jgoizueta: So, according to you @davidmanzanares this id was introduced to be able to remove layers which isn't yet implemented... I guess it's nice to use vl in those internal ids as well anyway

Issue #375, opened 2018-05-08
Tile border artifact


davidmanzanares:
It's not very noticeable, but there is a tile boundary artifact:

https://cartodb.github.io/carto-vl/examples/editor/index.html#eyJhIjoibW5tYXBwbHV0byIsImIiOiIiLCJjIjoiZG1hbnphbmFyZXMiLCJkIjoiaHR0cHM6Ly97dXNlcn0uY2FydG8uY29tIiwiZSI6ImNvbG9yOiByYW1wKGxpbmVhcihsb2coJG51bWZsb29ycyksIDEsIDQpLCBFYXJ0aClcbnN0cm9rZUNvbG9yOiBvcGFjaXR5KHdoaXRlLCAuMikiLCJmIjp7ImxuZyI6LTczLjkzNjA3MzM0ODU1MDksImxhdCI6NDAuODA2NzkxMTA0NjY1NTl9LCJnIjoxNS44ODM3OTgwNzM5ODg5NzUsImgiOiJEYXJrTWF0dGVyIn0=

![bug-tile](https://user-images.githubusercontent.com/1692175/39757925-a46b0cb0-52ce-11e8-9e60-ac37d946bbfe.png)

Could this be related to https://github.com/CartoDB/carto-vl/issues/365 ?

PR #430, opened 2018-05-18
Correct clipping

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/375

Before/After:
clipping

The current approach is much more correct, it may not be 100% correct yet, but right now I cannot see any artifacts on the mnmappluto dataset. (All previously perceived artifacts have been fixed). In fact, I see the same at bigger buffersizes (tested at 16, set at 1 to avoid unnecessary overhead)

### How is done

I coded the Sutherland-Hodgman algorithm to clip polygons on windshaft. I removed the broken (aproximated) previous "clipping".


src/core/decoder.js
```diff
@@ -38,14 +38,8 @@ function decodePoint(vertices) {
...
-function clip(x) {
-    if (x > 1) {
-        return 1;
-    }
-    if (x < -1) {
-        return -1;
-    }
-    return x;
+function isClipped(l) {
+    return l[0] == -1 || l[0] == 1 || l[1] == -1 || l[1] == 1;
```
Jesus89: Could we use `abs` here?

src/client/windshaft.js
```diff
@@ -431,39 +432,109 @@ export default class Windshaft {
...
+                polygon.flat.push(v[0], v[1]);
+            });
+        }
+    }
+    _intersect(a, b, c, d) {
+        //If AB intersects CD => return intersection point
+        // Intersection method from Real Time Rendering, Third Edition, page 780
+        const sub = ([ax, ay], [bx, by]) => ([ax - bx, ay - by]);
+        const dot = ([ax, ay], [bx, by]) => (ax * bx + ay * by);
+        const perpendicular = ([x, y]) => [-y, x];
```
Jesus89: I think we should declare these functions out of the _intersect function to improve performance.

src/client/windshaft.js
```diff
@@ -431,39 +432,109 @@ export default class Windshaft {
...
+        // Sutherland-Hodgman Algorithm to clip polygons to the tile
+        // https://www.cs.drexel.edu/~david/Classes/CS430/Lectures/L-05_Polygons.6.pdf
+        const clippingEdges = [
+            p => p[0] <= 1,
+            p => p[1] <= 1,
+            p => p[0] >= -1,
+            p => p[1] >= -1,
+        ];
+        const clippingEdgeIntersectFn = [
+            (a, b) => this._intersect(a, b, [1, -10], [1, 10]),
```
Jesus89: Why 10?

jgoizueta: Anything larger than the extended tile would do?

src/core/decoder.js
davidmanzanares: How?

src/client/windshaft.js
davidmanzanares: "premature optimization is the root of all evil" :)

src/client/windshaft.js
davidmanzanares: Yep (Anything larger than the extended tile should do it), I ran into some numeric problems when using `1`, the result was wrong by a very small amount, but just enough to mark it as no intersection which could create serious artifacts. The reason anything larger will work is that at that point we already know that there is an intersection point (because one point is outside the AABB and the other is inside).

src/client/windshaft.js
Jesus89: C'mon :P

src/core/decoder.js
Jesus89: `Math.abs(l[0]) == 1 ||  Math.abs(l[1]) == 1` ?

davidmanzanares:
I made a quick performance comparison. It seems that performance changes in the sense that, the windshaft decoder is slower and the dataframe decoder is faster. However, the total time is basically the same.

`master`
per-master
This branch:
perf-branch

davidmanzanares:
ping :eyes: :slightly_smiling_face:

src/core/decoder.js
davidmanzanares: Why is that better?

Jesus89: Not very huge improvement though, just a suggestion to improve readability. But it is fine that way :+1:

PR #441, opened 2018-05-21
Accept anything with a SELECT as a SQL source

### LINKED ISSUE #422: SQL source validation
The SQL source does some simplistic [validation](https://github.com/CartoDB/renderer-prototype/blob/05b907f22ae3804886d45dedb58b8859b78e2adb/src/api/source/sql.js#L71-L72
) of the query parameter (does it look like `SELECT ... FROM ...`?) which can be an inconvenient if using complex queries like [this one](https://gist.github.com/jgoizueta/2fbf74989a915521220a6e6d5965341c#file-index-html-L52-L74).

I propose removing the validation completely (leave it to the backend to parse the SQL) or simplyfy it even further (for example to `/\bSELECT\b/i`)

jgoizueta:
Fixes #422 

Let's do this so we can use funky SQL in the guides examples

### DIFF
--- a/src/api/source/sql.js
+++ b/src/api/source/sql.js
@@ -68,7 +68,7 @@ export default class SQL extends BaseWindshaft {
         if (query === '') {
             throw new CartoValidationError('source', 'nonValidQuery');
         }
-        var sqlRegex = /(SELECT|select)\s+.*\s+(FROM|from)\s+.*/;
+        var sqlRegex = /\bSELECT\b/i;
         if (!query.match(sqlRegex)) {
             throw new CartoValidationError('source', 'nonValidSQLQuery');
         }


src/api/source/sql.js
```diff
@@ -68,7 +68,7 @@ export default class SQL extends BaseWindshaft {
         if (query === '') {
             throw new CartoValidationError('source', 'nonValidQuery');
         }
-        var sqlRegex = /(SELECT|select)\s+.*\s+(FROM|from)\s+.*/;
+        var sqlRegex = /\bSELECT\b/i;
```
Jesus89: If the user is a `lowercase`fan this could be not enough, maybe `\b(SELECT|select)\b`?

jgoizueta: The `i` is for case insensitive, so it should match SELECT, select as well as Select or sElEct

Jesus89: Ah, you are right, I missed that part!

PR #443, opened 2018-05-21
Fix parser

davidmanzanares:
Fix parser identifier treatment. Returning `undefined` cause `viz.js` to use the defaults when an error should be thrown. For example with: `color: asd`.


davidmanzanares:
Regression test added

PR #446, opened 2018-05-22
Style guides

arianaescobar:


makella:
Let's merge these changes if we can! Lots of improvements in this round of guides.

PR #457, opened 2018-05-23
Linear default to min/max value

davidmanzanares:


src/core/viz/expressions/linear.js
```diff
@@ -28,6 +30,12 @@ import { checkExpression, checkLooseType, implicitCast, checkType } from './util
 export default class Linear extends BaseExpression {
     constructor(input, min, max) {
         input = implicitCast(input);
+
+        if (min == undefined && max == undefined) {
```
Jesus89: What about only one (min or max) undefined?

davidmanzanares: I don't think we want to support that. If the user doesn't put the `max` it's probably because one parenthesis is in the wrong place. If the user wants to avoid specifying a `min` he would need to do something like: `linear($wadus, undefined, 123)` which doesn't feel very good to me either.

Jesus89: Then, I think it's OK

davidmanzanares: Great! :rocket:

PR #458, opened 2018-05-24
Remove metabindings

IagoLast:
Did a find and is only used in this two places so I think is unused


davidmanzanares:
@IagoLast I think you started this branch from a branch you didn't want to :slightly_smiling_face:

PR #460, opened 2018-05-24
Reorganize shaders

IagoLast:
- Move all shaders to `core/shaders` folder
- Keep one file per class
- Add 3 folders: `common`, `style` and geometry each one with its own shaders and glsl code.
- Update compileProgram function to return parameters instead using side effects.


davidmanzanares:
"Update compile function to return parameters instead using side effects." Which of the compile functions? xD

IagoLast:
@davidmanzanares [compileProgram](https://github.com/CartoDB/carto-vl/pull/460/files#diff-98cac74a56f005c64f83d5b0063eb4fcR20)

davidmanzanares:
Did you cancel the test? :confused:

Issue #382, opened 2018-05-09
Cache linked shader programs instead of just vertex and fragment shaders


davidmanzanares:
We're already caching VS and FS compilations, but a final `gl.linkProgram` is not under that cache, and the following WebGL calls are very heavy.

As seen by @makella , when doing something as trivial as removing a lot of style from a comment inside the editor, the program gets linked once per letter, and the performance is completely destroyed, with pauses of more than 20 seconds. It seems that on top of that, something is making each call slower than the one before:

![bug-program](https://user-images.githubusercontent.com/1692175/39821758-37614e46-53a9-11e8-8a2a-f6afa62efefd.png)

We should cache the linked programs too to avoid this problem.

PS: great profiling image :joy:

davidmanzanares:
Fixed by https://github.com/CartoDB/carto-vl/pull/461 and https://github.com/CartoDB/carto-vl/pull/463

PR #461, opened 2018-05-25
Cache webgl programs

IagoLast:
#382
---

### DIFF
--- a/src/core/shaders/utils.js
+++ b/src/core/shaders/utils.js
@@ -1,8 +1,8 @@
-import ProgramCache from './ProgramCache';
+import Cache from './Cache';
 
 let programID = 1;
-const shaderCache = new ProgramCache();
-const programCache = new ProgramCache();
+const shaderCache = new Cache();
+const programCache = new Cache();
 
 /**
  * Compile a webgl program.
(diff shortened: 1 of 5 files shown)


src/core/shaders/utils.js
```diff
@@ -1,32 +1,30 @@
...
-        program: program,
-        programID: programID++
-    };
+    const program = createProgram(gl, VS, FS);
+    programCache.set(gl, code, { program: program, programID: programID++ });
+    return { program, programID };
 }
 
 function compileShader(gl, sourceCode, type) {
-    if (shaderCache.has(gl, sourceCode)) {
```
davidmanzanares: But, don't remove the shader (vertex/fragment) one!

This is still needed. A program may need to be recompiled, but not all parts may need to be recompiled. Actually, this is a very typical behavior since restyling modifies the fragment shader but the vertex shader is fixed.

src/core/shaders/utils.js
```diff
@@ -1,26 +1,28 @@
-import ShaderCache from './ShaderCache';
+import ProgramCache from './ProgramCache';
 
 let programID = 1;
+const shaderCache = new ProgramCache();
```
davidmanzanares: `shaderCache = new ProgramCache`?

IagoLast:
can we 🚀  this?

PR #462, opened 2018-05-25
Minor fixes in the guides

Jesus89:


docs/guides/05-introduction-to-styling.md
```diff
@@ -4,75 +4,55 @@ This guide walks through making a variety of visualizations for points, lines, a
...
 
-### Resolution
+#### Resolution
 Use `resolution` to aggregate points into clusters.
 
 ```
 resolution: 5
 ```
 
-### Opacity
```
arianaescobar: Why did you remove this? @Jesus89 @IagoLast

docs/guides/05-introduction-to-styling.md
```diff
@@ -4,75 +4,55 @@ This guide walks through making a variety of visualizations for points, lines, a
...
 
 ```
-color: rgba(232,66,244,0.5)
-```
-```
-color: opacity(ramp($animals,Prism),0.5)
+filter: between($numfloors, 10, 120)
 ```
 
-## Color by value
```
arianaescobar: This was done on purpose. The separation between the basic styling properties and the more "advanced" ones. We want to revert this change.

docs/guides/05-introduction-to-styling.md
```diff
@@ -97,7 +77,7 @@ To color features by the most common category, you can use the expression [`top`
 color: ramp(top($animals, 4),Prism)
 ```
 
-### Style by number
```
arianaescobar: Also, the naming is the same we followed in the examples, so they relate better. So we should revert these too.

docs/guides/05-introduction-to-styling.md
Jesus89: Opacity is not a `style property`. And `filter` was missing so I replaced it.

I just realized that we should move it to another place instead.

arianaescobar: Maybe not in the reference, but when you are trying to teach people about styling or designing a map, opacity is definitely part of it. Like the `a` in RGBA. We have to mention it.

docs/guides/05-introduction-to-styling.md
Jesus89: The problem here is that we have lots of levels so things are not looking good in the dev-center. With this approach, I think it is clearer (at least for me). But feel free to update it.

docs/guides/05-introduction-to-styling.md
Jesus89: Same

> The problem here is that we have lots of levels so things are not looking good in the dev-center. With this approach, I think it is clearer (at least for me). But feel free to update it.

docs/guides/05-introduction-to-styling.md
arianaescobar: **Form follows function.** We shouldn't compromise the way we teach our product because the styling is not ideal. Let's not change the content, let's improve the styling.

docs/guides/05-introduction-to-styling.md
arianaescobar: 👇

> **Form follows function.** We shouldn't compromise the way we teach our product because the styling is not ideal. Let's not change the content, let's improve the styling.

docs/guides/05-introduction-to-styling.md
makella: @Jesus89 the way we are defining opacity has always been a hard thing to explain (personally, I felt as though we should have a style property called `opacity` because it is very, very common as common as `color` and `width` in my opinion.

Since it isn't a "styling property" from the dev side but is indeed a styling property, where are you suggesting we put it?

If it is removed from a guide that walks through the basics of styling, it is likely that people will not find it.

PR #465, opened 2018-05-25
Fix variable _resolveAliases method to avoid having repeated names

davidmanzanares:
Easy fix, not so easy to debug problem. This was causing a major performance problem when using variables. Multiple repeated names were being added to `childrenNames`, which is used by almost all recursive methods in the expressions. This caused that something as trivial as an `isAnimated` method call cost 20 ms. By counting the number of visited nodes in those recursive calls I saw this: 55313 visited nodes!

Before fix profiling:
perf-bug

After fix profiling:
perf-fix

After the fix, the map I was debugging (editor => ethnic example) recovered the 60fps :small_airplane:

### DIFF
--- a/src/core/viz/expressions/basic/variable.js
+++ b/src/core/viz/expressions/basic/variable.js
@@ -47,6 +47,7 @@ export default class Variable extends BaseExpression {
     _resolveAliases(aliases) {
         if (aliases[this.name]) {
             this.childrenNames.push('alias');
+            this.childrenNames = [...(new Set(this.childrenNames))];
             this.alias = aliases[this.name];
         } else {
             throw new Error(`variable() name '${this.name}' doesn't exist`);


src/core/viz/expressions/basic/variable.js
```diff
@@ -47,6 +47,7 @@ export default class Variable extends BaseExpression {
     _resolveAliases(aliases) {
         if (aliases[this.name]) {
             this.childrenNames.push('alias');
+            this.childrenNames = [...(new Set(this.childrenNames))];
```
IagoLast: Nice fix, IMO this is easier to read:

```js
const uniqueChildrenNames = new Set(this.childrenNames);
this.childrenNames = Array.from(uniqueChildrenNames);
```

PR #467, opened 2018-05-25
Guides update

arianaescobar:


arianaescobar:
It was just for highlight the code. And since that whole styling part is inside the ```const viz = new 
`);```, I felt it was the closer highlight we could put there. 

Thanks for the CR, although I'm planning to add more commits to this one 🙃

Jesus89:
OK. I'm also thinking that this Viz code will be always a string, therefore, no syntax highlight inside a string.

selection_168

arianaescobar:
Yeah, we should find a better way to highlight the code later on. Can you open an issue please?

Jesus89:
Since we are using strings all the chars would be the same.

However, I'm going to create an issue just to talk about that.

PR #471, opened 2018-05-28
Fix unnecessary reinstantiations

jgoizueta:
Instantiation was happening for small datasets even if no filters were used

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -350,9 +350,6 @@ export default class Windshaft {
         }
         const response = await fetch(endpoint(conf), this._getRequestConfig(mapConfigAgg));
         const layergroup = await response.json();
-        if (!response.ok) {
-            throw new Error(`Maps API error: ${JSON.stringify(layergroup)}`);
-        }
         this._subdomains = layergroup.cdn_url ? layergroup.cdn_url.templates.https.subdomains : [];
         return {
             url: getLayerUrl(layergroup, LAYER_INDEX, conf),
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -350,6 +350,9 @@ export default class Windshaft {
         }
         const response = await fetch(endpoint(conf), this._getRequestConfig(mapConfigAgg));
         const layergroup = await response.json();
+        if (!response.ok) {
+            throw new Error(`Maps API error: ${JSON.stringify(layergroup)}`);
+        }
         this._subdomains = layergroup.cdn_url ? layergroup.cdn_url.templates.https.subdomains : [];
         return {
             url: getLayerUrl(layergroup, LAYER_INDEX, conf),
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -129,7 +129,12 @@ export default class Windshaft {
      *  - When the filter conditions changed and the dataset should be server-filtered.
      */
     _needToInstantiate(MNS, resolution, filtering) {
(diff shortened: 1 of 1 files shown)


src/client/windshaft.js
```diff
@@ -129,7 +129,12 @@ export default class Windshaft {
      *  - When the filter conditions changed and the dataset should be server-filtered.
      */
     _needToInstantiate(MNS, resolution, filtering) {
-        return !R.schema.equals(this._MNS, MNS) || resolution != this.resolution || (JSON.stringify(filtering) != JSON.stringify(this.filtering) && this.metadata.featureCount > MIN_FILTERING);
+        return !R.schema.equals(this._MNS, MNS)
+            || resolution != this.resolution
+            || (
+                JSON.stringify(filtering) != JSON.stringify(this.filtering)
```
davidmanzanares: Hey, I guess I'm missing something, but I don't understand how the PR solves the problem. I was expecting some change in the conditions of this function.

src/client/windshaft.js
```diff
@@ -160,7 +165,7 @@ export default class Windshaft {
         };
         if (metadata) {
             if (metadata.featureCount >= 0) {
-                choices.backendFilters = metadata.featureCount > MIN_FILTERING;
+                choices.backendFilters = metadata.featureCount > MIN_FILTERING || !metadata.backendFiltersApplied;
```
davidmanzanares: I don't understand this. Why not just `choices.backendFilters = metadata.backendFiltersApplied;` ?

davidmanzanares:
Sorry, I pushed some changes accidentally to this branch, I already reverted the commit.

jgoizueta:
Hey @davidmanzanares I think this warrants some explanation, cause, as you noticed, this doesn't seem to make much sens at first sight.

It's messy, but not too easy to simplify, and since reinstantiation will probably go away soon, I'd prefer to leave it as is.

The (infamously named) `choices` affect instantiation; `backendFilters` in particular tells the instantiation to send backend filters or not.

On the other hand, only the instantiation knows if filters where actually sent to the backend (even if backendFilters choice was true, there could be no filters or no filters that could be executed in the server). I've added this information explicitly to the instantiation results.

So what we do is:
* The first time we instantiate (we don't if the dataset is large and we don't know if any filter could executable in the backend) se set backendFilters to true. (note that otherwise we won't even try to send filters and we won't ever know if filters could have been sent to backend).
* Then we reevaluate our choices and if they changed we'll need to reinstantiate. So, in partticular we'll need to reinstantiate if backendFilters changes to `false`.
 
That's why, if we notice filters are not needed we set `backendFilters` paradoxically to true: it will prevent reinstantiation only for filters shake, and it won't hurt if we do reinstantiate, because no filters will acctually be sent. On the other hand, if filters were used and the dataset is small we set `backendFilters` to false, which triggers reinstantiation without filters.

davidmanzanares:
I don't understand this part:
> The first time we instantiate (we don't if the dataset is large and we don't know if any filter could executable in the backend) we set backendFilters to true. 

In general, we don't want filters applied, right? Because most maps will have a feature count below the threshold, right? So... If we set it to true, wouldn't we reinstantiating most of the times?

jgoizueta:
I initially thought also that it's best to first not apply filters. Then I realised that would require more complex changes, and that reinstantiating for the large datasets, even if less frequent, could be more costly. 

Reconsidering this, I might be wrong in the last part, because metadata is not recomputed in the re-instantiation, and the other checks that Maps API peforms may not be that costly.

So, probably you're right... but let's merge this which is an improvement anyway 😛 

If we don't eliminate re-instantiation soon we can get back to that (choosing no filters in the first instantiation)

davidmanzanares:
I see, let's merge this, then, but we'll need to review that reinstantiation :smile:

Issue #210, opened 2018-03-22
Maps API error control


davidmanzanares:
Right now we are not checking that Maps API success on instantiation. This means that when there is a problem the client tries to download tiles from with an `undefined` layergroupID.

We should control this error condition.

davidmanzanares:
Situations:

- [ ] Invalid username. The server returns HTTP 302 redirecting to http://carto.com/noneuser.html . Should the API be changed? I would expect a 4xx not a 3XX. Moreover this finally fails with a `Response for preflight is invalid (redirect)`
- [ ] invalid apiKey with a public dataset: defaults to public one. Is this expected? or an error should be thrown?
- [ ] invalid apiKey with a private dataset. The server returns a 403 with an error in a JSON. I think we should check for a 403, and output an error with "invalid apikey..." in that case.
- [ ] styling with property names that don't exist. The server returns a 404 with an error like: `{"errors":["column \"wadus\" does not exist"]...`. I think we should use that error message and just retransmit it (wrapping it accordingly).

elenatorro:
After reading the different tasks we should start next, I think this one is highly related with:

* **Unify API Promises/Error management** https://github.com/CartoDB/carto-vl/issues/342
* **Use factory instead constructor to create layers** https://github.com/CartoDB/carto-vl/issues/321

In my opinion, we could solve this one along with #342.

#321 would be blocked until it gets solved. 
What do you think? cc @rochoa

davidmanzanares:
I've rechecked the previous situations since there have been some Maps API changes:
Situations:

- [ ] Invalid username. The server returns HTTP 302 redirecting to http://carto.com/noneuser.html . Should the API be changed? I would expect a 4xx not a 3XX. Moreover, this finally fails with a `Response for preflight is invalid (redirect)`
- [ ] invalid apiKey with a public dataset: Maps API returns an error that we should control
- [ ] invalid apiKey with a private dataset: Maps API returns an error that we should control
- [ ] styling with property names that don't exist. The server returns a 404 with an error like: `{"errors":["column \"wadus\" does not exist"]...`. I think we should use that error message and just retransmit it (wrapping it accordingly).

I'm going to start working on the last 3 situations, the first situation will require work from Maps API if we decide to change the public API (which I think we should).

rochoa:
[...]

PR #473, opened 2018-05-28
Windshaft error control

davidmanzanares:
I detected a problem when asking for invalid property names in the `viz`. As expected, things failed, however, it left the status in a place where the map couldn't recover.

This happened because, after the next blendToViz, the previous `viz` was marked as `ok`, so it tried to create a blending between the invalid property and the new, valid, one. This cause another invalid reinstantiation.

The root cause of this is that when Maps API returned some error, the client didn't check the HTTP status code, and the returned error was in a JSON format compatible with the expected response (in the sense that no error was thrown at instantiation time). In fact, this created an MVT tile rule with a layergroupID of `undefined`.

This problem is deeply related with https://github.com/CartoDB/carto-vl/issues/210 but, that issue is focused on managing the errors with good user message, while this PR just detects the HTTP status code error and throws a not perfectly formatted error.

I think that for now, we could just live with this patch, and in the future, we should take a closer look at that issue and go case by case, error by error.

Note: this PR starts from https://github.com/CartoDB/carto-vl/pull/471 , wait for that merge before merging this.

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -350,6 +350,9 @@ export default class Windshaft {
         }
         const response = await fetch(endpoint(conf), this._getRequestConfig(mapConfigAgg));
         const layergroup = await response.json();
+        if (!response.ok) {
+            throw new Error(`Maps API error: ${JSON.stringify(layergroup)}`);
+        }
         this._subdomains = layergroup.cdn_url ? layergroup.cdn_url.templates.https.subdomains : [];
         return {
             url: getLayerUrl(layergroup, LAYER_INDEX, conf),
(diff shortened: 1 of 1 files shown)


src/client/windshaft.js
```diff
@@ -338,6 +350,9 @@ export default class Windshaft {
         }
         const response = await fetch(endpoint(conf), this._getRequestConfig(mapConfigAgg));
         const layergroup = await response.json();
+        if (!response.ok) {
+            throw new Error(`Maps API error: ${JSON.stringify(layergroup)}`);
```
jgoizueta: I'm always a little confused when promises are involved... so in this case it's preferable to cause the exception here than returning a rejected promise?

davidmanzanares: Yes, here is safe because we are inside an `async` function, therefore, thrown exceptions will automatically reject the promise that the function returns.

PR #478, opened 2018-05-28
Sample without geometry

jgoizueta:
Geometry can get large easily (e.g. polygons) and we don't need it in the sample, so avoid it.

This requires [backend changes](https://github.com/CartoDB/Windshaft-cartodb/pull/968) to be deployed

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -200,7 +200,7 @@ export default class Windshaft {
             aggSQL = filteredSQL;
         }
 
-        let { url, metadata } = await this._getInstantiationPromise(query, conf, agg, aggSQL, overrideMetadata);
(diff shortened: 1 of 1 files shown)


src/client/windshaft.js
```diff
@@ -282,14 +282,14 @@ export default class Windshaft {
     }
 
     _buildSelectClause(MRS, dateFields = []) {
-        return MRS.columns.map(name => R.schema.column.getBase(name)).map(
+        const columns = MRS.columns.map(name => R.schema.column.getBase(name)).map(
```
davidmanzanares: Isn't this `MNS`?

jgoizueta: oops! but I consistently misnamed it 😛

jgoizueta: But... Minimum Required Schema sounds even better than Minimum Needed Schema!

PR #481, opened 2018-05-29
Minor performance improvement

davidmanzanares:
I tried to improve the performance in cases where there are many vertices.

The results are mediocre, to be honest. They are a 0.3%.

However, I think we should merge this anyway because the code is almost as simple as before and the improvement may be better in other circumstances (datasets, styles, and hardware).

This PR also includes a minor improvement to `benchmark.js` to show the standard error of the average, to be able to compare averages with some context. See http://www.stat.wmich.edu/s216/book/node81.html#zci2

Results (avg is in milliseconds, lower is better):
```
Before patch
avg: 21.851120473157415,
stderr: 0.07242037686506851

After patch
avg: 21.783982937012162,
stderr: 0.07237076023544146,
```

### DIFF
--- a/test/benchmark/benchmark.html
+++ b/test/benchmark/benchmark.html
@@ -36,7 +36,7 @@
     });
 
     carto.setDefaultAuth({
-      user: 'mamataakella',
+      user: 'cartovl',
       apiKey: 'default_public'
     });
 
--- a/test/benchmark/benchmark.html
+++ b/test/benchmark/benchmark.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html>
+
 <head>
   <title>Animation | CARTO</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
@@ -15,6 +16,7 @@
       margin: 0;
       padding: 0;
     }
+
     #map {
       position: absolute;
       height: 100%;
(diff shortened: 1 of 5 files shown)


test/benchmark/benchmark.html
```diff
@@ -15,40 +16,42 @@
...
       style: 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json',
-      center: [0, 30],
-      zoom: 1
+      center: [-104.93637263263889, 39.72562104533162],
+      zoom: 11
     });
 
     carto.setDefaultAuth({
-      user: 'cartovl',
+      user: 'mamataakella',
```
jgoizueta: Why not copying the dataset to cartovl? (we better have a single account dependency for all things CARTO VL)

davidmanzanares: Ups, I forgot

IagoLast:
maybe is a micro-optimization (dmanzanares loves them 😉) but if we use early-returns instead else conditions in the glsl code we avoid unnecessary comparisons (and is IMO easier to read)

davidmanzanares:
I disagree about being an optimization at all. The current code doesn't need to keep evaluating the `else` since the code already must have entered the `if`. You could say that there is an additional copy to a variable, but that doesn't really matter since internally, the compiler must do that anyway. In fact, early returns can hurt performance if the compiler doesn't optimize them in the best way. This happens because GPU hardware doesn't have real stacks nor the regular architecture of a CPU. You may want to know more about SIMT (https://en.wikipedia.org/wiki/Single_instruction,_multiple_threads) and how the execution model is based on executing groups of threads in lock-step, as opposed to CPU threads in which each thread is a completely independent thread with its own program counter.

Said that, and to be honest, I would prefer the early return in JS code :joy:

davidmanzanares:
I changed the user to `cartovl`, but the dataset is not there yet, DNM yet

davidmanzanares:
The dataset is already there and I checked it works, merging!

PR #484, opened 2018-05-30
Update to Webpack4 and enable tree shaking

davidmanzanares:
Update to Webpack 4 and enable tree-shaking.

The minified bundle has been reduced from 226KB to 211KB

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@carto/carto-vl",
   "version": "0.3.0",
+  "sideEffects": false,
   "description": "CARTO Vector library",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@carto/carto-vl",
   "version": "0.3.0",
+  "sideEffects": false,
```
IagoLast: what is this?

davidmanzanares: This marks the code as pure, to enable tree shaking, see https://webpack.js.org/guides/tree-shaking/

PR #487, opened 2018-05-31
MVT Source

ethervoid:
This is the MVT source I made to be able to consume MVT directly from an URL without using Windshaft

We use it to show the DO Tiler tiles and I have made some changes like:

- Expose Metadata as an object to be able to pass it with the expected columns schema coming from the MVT (we're not able to gather it prior to ask the first MVT)

#### TODO (After CR)

- [x] Add some unit tests


src/api/source/mvt.js
```diff
@@ -0,0 +1,452 @@
...
+        tiles.forEach(t => {
+            const { x, y, z } = t;
+            this._getDataframe(x, y, z).then(dataframe => {
+                if (dataframe.empty) {
+                    needToComplete--;
+                } else {
+                    completedTiles.push(dataframe);
+                }
+                if (completedTiles.length == needToComplete) {
+                    this._oldDataframes.map(d => d.active = false);
```
davidmanzanares: There is a `requestGroupID` logic in `windshaft.js` that has not been placed here. That logic is important since without it, if tiles from an early request come after the last tiles requested are processed, then the first ones replace the new ones, which may not cover the screen completely.

src/api/source/mvt.js
```diff
@@ -0,0 +1,452 @@
...
+                }
+                polygon = {
+                    flat: [],
+                    holes: [],
+                    clipped: [],
+                    clippedType: [], // Store a bitmask of the clipped half-planes
+                };
+            } else {
+                if (j == 0) {
+                    console.log('Invalid MVT tile: first polygon ring MUST be external');
```
davidmanzanares: If this is the case, we may want to assume that the orientation is reversed entirely. In that case a `continue` won't be enough, but this depends much on our expectations regarding the backend.

src/api/source/mvt.js
```diff
@@ -0,0 +1,452 @@
...
+                    e = f.properties[c.name];
+                }
+                properties[fieldMap[c.name]][i] = e;
+            });
+        }
+
+        return { properties, points, featureGeometries, geomType };
+    }
+
+    _computePointGeometry(data) {
```
davidmanzanares: I think all these `compute*` functions are dead code, right? @

src/api/source/mvt.js
jgoizueta: We probably should refactor and extract a `TiledSource` class to manage all this logic, then derive both `Windshaft` and `MVT` from it (or use it as a helper). But for now I'd just reproduce this exactly as in Windshaft and leave the refactor for after this is merged.

src/api/source/mvt.js
```diff
@@ -0,0 +1,452 @@
...
+
+    bindLayer(addDataframe, removeDataframe, dataLoadedCallback) {
+        this._addDataframe = addDataframe;
+        this._removeDataframe = removeDataframe;
+        this._dataLoadedCallback = dataLoadedCallback;
+    }
+
+    async requestMetadata(viz) {
+        this._MNS = viz.getMinimumNeededSchema();
+        this._resolution = viz.getResolution;
```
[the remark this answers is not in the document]
ethervoid: In this case, I just copied what Windshaft source does. What is the main purpose of the resolution? https://github.com/CartoDB/carto-vl/blob/master/src/client/windshaft.js#L83

src/api/source/mvt.js
```diff
@@ -0,0 +1,452 @@
...
+        return this.metadata;
+    }
+
+    requestData(viewport) {
+        const tiles = rsys.rTiles(viewport);
+        this._getTiles(tiles);
+    }
+
+    free() {
```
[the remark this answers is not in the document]
ethervoid: done

src/api/source/mvt.js
ethervoid: deleted, you're right

src/api/source/mvt.js
ethervoid: added

src/api/source/mvt.js
ethervoid: Here we have a situation. Some polygons are CW but because they have loops, for example, they are treated as CCW. Also, invalid polygons could fall into this category. Is something we have to work in. I'll give it a try after this is merged

src/api/source/mvt.js
```diff
@@ -0,0 +1,452 @@
...
+     * @fires CartoError
+     *
+     * @constructor MVT
+     * @extends carto.source.Base
+     * @memberof carto.source
+     * @api
+     */
+    constructor(serverURL, metadata) {
+        super();
+        // this._validateServerURL(serverURL);
```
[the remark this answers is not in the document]
ethervoid: Yep, adding...

src/api/source/mvt.js
```diff
@@ -0,0 +1,452 @@
...
+import * as Protobuf from 'pbf';
+import * as LRU from 'lru-cache';
+import { VectorTile, VectorTileFeature } from '@mapbox/vector-tile';
+
+export default class MVT extends Base {
+
+    /**
+     * Create a carto.source.MVT.
+     *
+     * @param {object} data - A MVT data object
```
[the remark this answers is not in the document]
ethervoid: Added

src/index.js
```diff
@@ -8,29 +8,34 @@
...
  * - {@link carto.source.SQL|carto.source.SQL}
  * - {@link carto.source.GeoJSON|carto.source.GeoJSON}
+ * - {@link carto.source.MVT|carto.source.MVT}
  * - {@link carto.expressions|carto.expressions}
  * - {@link carto.Layer|carto.Layer}
  * - {@link carto.Viz|carto.Viz}
  * - {@link carto.Interactivity|carto.Interactivity}
  * - {@link carto.setDefaultAuth|carto.setDefaultAuth}
  * - {@link carto.setDefaultConfig|carto.setDefaultConfig}
+ * - carto.data
```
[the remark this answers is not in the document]
ethervoid: removed (was a kind of separator)

ethervoid:
retest this please

src/api/source/mvt.js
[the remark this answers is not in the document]
davidmanzanares: `metadatadata`

src/api/url.js
```diff
@@ -0,0 +1,20 @@
+import CartoValidationError from './error-handling/carto-validation-error';
```
ethervoid: @davidmanzanares @jgoizueta if you see that this refactor is no necessary I can revert it

src/api/source/mvt.js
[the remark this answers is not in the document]
ethervoid: fixed

src/api/url.js
jgoizueta: Please keep it 😄

ethervoid:
@davidmanzanares @jgoizueta done, second round? :D

src/api/source/mvt.js
```diff
@@ -0,0 +1,269 @@
...
+                const jsonFields = [];
+                const stringFields = [];
+                this.metadata.columns.map(c => {
+                    const type = c.type;
+                    if (type == 'json') {
+                        jsonFields.push(c.name);
+                    } else if (type == 'number') {
+                        numFields.push(c.name);
+                    } else if (type == 'category') {
+                        stringFields.push(c.name);
```
ethervoid: Right now this is just plain strings, not categories because we can use string data coming from MVT. We should figure out how to use categories

src/api/source/mvt.js
```diff
@@ -0,0 +1,269 @@
...
+                var tile = new VectorTile(new Protobuf(response));
+                const mvtLayer = tile.layers[Object.keys(tile.layers)[0]];
+                var fieldMap = {};
+                const numFields = [];
+                const jsonFields = [];
+                const stringFields = [];
+                this.metadata.columns.map(c => {
+                    const type = c.type;
+                    if (type == 'json') {
+                        jsonFields.push(c.name);
```
ethervoid: Probably we should add arrays in the future too

davidmanzanares: Well... This has been discussed a lot, particularly a hook system that allows developers to do pre-processing of the MVTs before passing them to CARTO VL rendering system. But, we decided to don't do it in the near future unless we see that we really need it.

davidmanzanares:
Hi!

Sorry for the delay @ethervoid 

Last week we decided to give a huge push to the MVT source. Our plan is to do a few small changes to the PR (is it ok to you if I just push to this branch instead of creating a new one?) and then add some extra functionality (in other PRs, after this is merged) that we need, like giving support for 500-600 properties on the MVTs.

ethervoid:
@davidmanzanares perfect for me. This was just the beginning for this source. I started working on it to have something to show MVT without windshaft. So if you take it from here its perfect for me :)

davidmanzanares:
Great!

And thank you for making the PR!

davidmanzanares:
I made significant changes to the MVT and Windshaft sources. I extracted some common TileClient functionality from both, and I made MVT the internal source of Windshaft. Relieving Windshaft from the MVT decoding functionality.

Right now I created a few regressions that must be addressed before merging this:
- [ ] Fix Dates in windshaft
- [ ] Fix multiple subdomains
- [ ] Fix performance

davidmanzanares:
Hi,

I made quite a few changes and I think this is ready for another, hopefully final, CR. I left some things outside this PR, but I think it should be merged because of the importance of the windshaft refactor, the likelihood of problematic merge conflicts and because I think that this no longer introduces regressions (I hope, I guess).

Things I made:

- Extract the `decodeLines` function
- Add E2E test for MVT
- Extract the LRU cache for dataframes functionality
- Autodiscover the geometry type from the MVT tiles
- Extract the new `TileClient`
- Adapt Windshaft to use internally the new MVT source
- Add a way to override the default property decoder of the MVT source

Things that I think should be done in future PRs:
- Document the metadata object, check it and add tests to it
- Add unit and integration tests
- Refactor the inheritance of Maps API sources
- Refactor the paths of the files and the directory structure.

src/api/source/mvt.js
```diff
@@ -0,0 +1,175 @@
...
+        const rs = rsys.getRsysFromTile(x, y, z);
+        const dataframeGeometry = this._metadata.geomType == geometryTypes.POINT ? points : featureGeometries;
+        const dataframe = this._generateDataFrame(rs, dataframeGeometry, properties, mvtLayer.length, this._metadata.geomType);
+        return dataframe;
+    }
+
+
+    _decodeMVTLayer(mvtLayer, metadata, mvt_extent) {
+        let points;
+        if (metadata.geomType == geometryTypes.POINT) {
```
elenatorro: I think this early check is not needed, because it is also done again in the loop.

src/api/source/mvt.js
```diff
@@ -0,0 +1,175 @@
...
+     * new carto.source.MVT("https://{server}/{z}/{x}/{y}.mvt", metadata);
+     *
+     * @fires CartoError
+     *
+     * @constructor MVT
+     * @extends carto.source.Base
+     * @memberof carto.source
+     * @api
+     */
+    // FIXME, multi subdomain
```
elenatorro: What this FIXME means?

src/api/source/mvt.js
```diff
@@ -0,0 +1,175 @@
...
+        super();
+        this._templateURL = templateURL;
+        this._metadata = metadata;
+        this._tileClient = new TileClient(templateURL);
+        this._categoryStringToIDMap = {};
+        this._numCategories = 0;
+    }
+
+    _clone() {
+        return new MVT(this._templateURL, JSON.parse(JSON.stringify(this._metadata)));
```
elenatorro: Why do we need `JSON.parse(JSON.stringify(this._metadata))`?

src/api/source/mvt.js
```diff
@@ -0,0 +1,175 @@
...
+                featureGeometries.push(decodedLines);
+            } else {
+                throw new Error(`Unimplemented geometry type: '${metadata.geomType}'`);
+            }
+            Object.keys(f.properties).forEach(propertyName => {
+                const propertyValue = f.properties[propertyName];
+                const decodedPropertyValue = this.decodeProperty(propertyName, propertyValue);
+                if (decodedPropertyValue !== undefined) {
+                    if (decodedProperties[propertyName] === undefined) {
+                        decodedProperties[propertyName] = new Float32Array(mvtLayer.length + 1024);
```
elenatorro: Why `1024`?

src/api/url.js
```diff
@@ -0,0 +1,20 @@
...
+
+export function validateServerURL(serverURL) {
+    var urlregex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/;
```
elenatorro: It would be great to have regex constants outside. Also, as they are constants, it should be `const URL_REGEX = //`

src/api/source/mvt.js
```diff
@@ -0,0 +1,175 @@
...
+        }
+        let featureGeometries = [];
+        const decodedProperties = {};
+        for (let i = 0; i < mvtLayer.length; i++) {
+            const f = mvtLayer.feature(i);
+            const geom = f.loadGeometry();
+            const mvtGeomType = f.type;
+            if (metadata.geomType === undefined) {
+                switch (mvtGeomType) {
+                    case 1:
```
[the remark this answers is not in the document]
davidmanzanares: ahg... I hate bad libraries... I'll improve this... the best I can.

src/api/source/mvt.js

src/api/source/mvt.js
davidmanzanares: Ups, the FIXME is outdated

src/api/source/mvt.js
davidmanzanares: I wanted to play it safe, a clone should not share state with its "cloner"

src/api/source/mvt.js
davidmanzanares: Ups, I should have used the constant `RTT_WIDTH` defined in `renderer.js`. The reason for this addition is that when we use the property Float32Array in `dataframe.js` to upload it to WebGL there is a required padding of the width of the texture, which is equal to `RTT_WIDTH`. Adding `RTT_WIDTH` assures that the padding exists. I'll document this reason on the code.

src/api/source/mvt.js
davidmanzanares: I thought it was redundant, but, after changing it, it seems it's not. The reason is that the second `if` is executed only when the geometry type wasn't defined previously. This makes this first check necessary, of course, additional logic can be used to remove this `if` and make the second `if` work in both cases, but... I think it's basically the same.

elenatorro: ok!

src/api/layer.js
```diff
@@ -318,22 +319,20 @@ export default class Layer {
      * @param {Dataframe} dataframe
      */
     _onDataframeAdded(dataframe) {
+        dataframe.freeObserver = () => {
```
IagoLast:
```js
dataframe.freeObserver = this._onDataFrameRemoved.bind(this)
```
or

```js
dataframe.onRemove(this.onDataframeRemoved.bind(this)
```

src/api/source/mvt.js
```diff
@@ -0,0 +1,179 @@
...
+    POLYGON: 'polygon'
+};
+
+export default class MVT extends Base {
+
+    /**
+     * Create a carto.source.MVT.
+     *
+     * @param {object} data - A MVT data object
+     * @param {object} metadata - A carto.source.mvt.Metadata object
```
IagoLast: this object does not exist or is not available under the carto namespace.

I think we should update document and make public  [this object](https://github.com/CartoDB/carto-vl/blob/master/src/core/metadata.js)

src/api/source/mvt.js
IagoLast: should we create an `util.clone` method?

src/api/source/TileClient.js
```diff
@@ -0,0 +1,80 @@
...
+                        completedTiles.map(d => d.active = true);
+                        this._oldDataframes = completedTiles;
+                        this._dataLoadedCallback();
+                    }
+                });
+        });
+    }
+
+    _getDataframe(x, y, z) {
+        return this._cache.get(`${x},${y},${z}`, () => this._requestDataframe(x, y, z));
```
IagoLast: - Calling `_requestDataframe` without the 4º parameter  will make[this](https://github.com/CartoDB/carto-vl/pull/487/files#diff-f44fa8904729341dc09a5c44a14ed8c5R74) fail.

- This function is not used

davidmanzanares:
Hey, @IagoLast I fully agree that the current structure is a mess. But it is not a **new** mess.

In fact, the first thing I did when I entered into this codebase was creating https://github.com/CartoDB/carto-vl/issues/544 and https://github.com/CartoDB/carto-vl/issues/546

I don't agree much though on that this is within the scope of this PR.

And, also, I don't agree with your proposed new folder structure. In fact, I think it contains some of the same problems that the current one has:

- You separated `mvt` from `api` when `MVT` is part of the API.
- You're still using an `api/`. The API folder for me makes zero sense. In fact, we have 90% of the API (all viz expressions) outside the `api/` folder. Every class will have some public things and some private things, dividing the code between `api/` and the rest creates non-sense separations like having some source things in one root folder and having the real implementation in other folders. If you were consistent with separating between `api` and not-api, you will each expression in two places, the api part and the not api part.

src/api/layer.js
davidmanzanares: :+1:

src/api/source/mvt.js
davidmanzanares: As commented in my previous comment, I left the Metadata object undocumented on purpose. For me, if this PR is merged, we will have an "experimental" MVT source. Not more, since we'll need to address some issues and I don't feel comfortable right now to write the API for the metadata object, I prefer to wait for that. That means that we shouldn't release a version with MVT support until these details are sorted out.

IagoLast:
@davidmanzanares you are right, we should do a deep folder review since the client/api doesnt make much sense 

In any case my two cents:

- IMO we should try to do as much refactoring as we can in the issues.
- Agreed, much of the mess was already there, but the `TileClient` is a new file and IMO with the current folder structure should be in the `client` folder.
-

src/api/source/TileClient.js
```diff
@@ -0,0 +1,80 @@
...
+        this._cache = new DataframeCache();
+        this._oldDataframes = [];
+    }
+
+    _getTiles(tiles, responseToDataframeTransformer) {
+        this._requestGroupID++;
+        let completedTiles = [];
+        let needToComplete = tiles.length;
+        const requestGroupID = this._requestGroupID;
+        tiles.forEach(t => {
```
IagoLast: tile

src/api/source/TileClient.js
```diff
@@ -0,0 +1,80 @@
...
+            const { x, y, z } = t;
+            this._cache.get(`${x},${y},${z}`, () => this._requestDataframe(x, y, z, responseToDataframeTransformer)).then(
```
IagoLast: this is not working as a cache and IMO is very confusing.

IMO the fetch logic should remain in the MVT client and the cache must remain a simple and classic cache!

something like:
```js
if (this._cache.has(ID)){
   this._cache.get(ID).then(this._onDataframeCallback)
}
else {
   const dataFramePromise = this._requestDataframe(x, y, z, responseToDataframeTransformer);
   this.cache.set(ID, dataframePromise);
   dataframePromise.then(this._onDataframeCallback);
}


```

src/api/source/mvt.js
IagoLast: ok, so I think is better to remove all the metadata references from the doc

src/client/mvt/feature-decoder.js
```diff
@@ -8,6 +8,18 @@ export class Polygon {
     }
 }
 
+export function decodeLines(geom, mvt_extent) {
```
IagoLast:
```js
export function decodeLines(geom, mvt_extent) {
    return geom.map(l => l.map(point => (2 * point.x / mvt_extent - 1, 2 * (1 - point.y / mvt_extent) - 1)));
}
```

?

src/api/layer.js
```diff
@@ -331,22 +332,20 @@ export default class Layer {
...
+            this._renderLayer.removeDataframe(dataframe);
```
IagoLast: This function [is removed in this PR](https://github.com/CartoDB/carto-vl/pull/487/files#diff-f276b6fefb280b9dbe9c2b44aa3c035eL25) I think its gonna crash

src/api/source/TileClient.js
davidmanzanares: I think it's pretty clear that it refers to a tile, it is only used in the next line

src/api/source/TileClient.js
davidmanzanares: The current cache is uncoupled with the tile format, what you are proposing is to duplicate the cache code for each format. For a common cache class I already have LRU-cache, what I need is another layer that encapsulates the dataframe-related logic of the cache.

IagoLast: > The current cache is uncoupled with the tile format, what you are proposing is to duplicate the cache code for each format

What I am proposing is even less coupled since the cache simply stores promises.

> for each format

Which formats do we have? 

> What I need is another layer that encapsulates the dataframe-related logic of the cache.

I dont follow you

src/client/mvt/feature-decoder.js
davidmanzanares: Not exactly, that returns a different kind of structure, but I've simplified the function a little bit

src/api/source/TileClient.js
IagoLast:
```js
tiles.forEach(({x, y, z}) => {
```

src/api/source/mvt.js
davidmanzanares: I undocumented the MVT class, since it is experimental

src/api/source/mvt.js
davidmanzanares: The first question for a method like that will be? Is it shallow? Is it deep? Will it allow circular dependencies? With the current implementation, you don't need to look up those in our internal utils since it only use the built-in utils.

src/api/source/TileClient.js
davidmanzanares: Of course, the code you still leave at the cache is uncoupled, but you move code to the MVT file, that code is coupled.

Before following refactors the DataframeCache class was shared by the MVT source and the Windshaft source. After more refactoring, windshaft just use MVT, this means that the DataframeCache is used right now by only one class (MVT), but the abstraction worked, works, and if we add a new format in the future it will still work.

The DataframeCache encapsulates the logic of "how many dataframes to keep", "for how long", and "what to do in case we need to remove one from the cache". To do it, it needs to know about Dataframes, it doesn't need to know about MVT/GeoJSON/...

IagoLast: > but you move code to the MVT file, that code is coupled.

which code? all the code was already there 

> Before following refactors the DataframeCache class was shared by the MVT source and the Windshaft source. After more refactoring, windshaft just use MVT, this means that the DataframeCache is used right now by only one class (MVT), but the abstraction worked, works, and if we add a new format in the future it will still work.

In which way my proposal is not compatible with that?


> The DataframeCache encapsulates the logic of "how many dataframes to keep", "for how long", and "what to do in case we need to remove one from the cache". To do it, it needs to know about Dataframes, it doesn't need to know about MVT/GeoJSON/...

With the current approach the cache knows nothing about the dataframe, but needs a second parameter in the `get` function I never seen before in a cache

davidmanzanares:
@elenatorro @jgoizueta ping :slightly_smiling_face:

src/api/source/TileClient.js
elenatorro: I agree with @IagoLast that we should use descriptive names even if it is clear for us.

src/api/source/TileClient.js
davidmanzanares: eing?

src/api/source/base-windshaft.js
```diff
@@ -42,17 +42,10 @@ export default class BaseWindshaft extends Base {
     _generateURL(auth, config) {
         let url = (config && config.serverURL) || DEFAULT_SERVER_URL_TEMPLATE;
         url = url.replace(/{user}/, auth.username);
-        this._validateServerURL(url.replace(/{local}/, ''));
+        validateServerURL(url.replace(/{local}/, ''));
         return {
             maps: url.replace(/{local}/, ':8181'),
```
elenatorro: In my opinion, PORT numbers and any settings related should be in a config file. `/setup/config-service` could be a good place.

src/api/source/mvt.js
```diff
@@ -0,0 +1,179 @@
...
+    async requestMetadata() {
+        return this._metadata;
+    }
+
+    requestData(viewport) {
+        return this._tileClient.requestData(viewport, this.responseToDataframeTransformer.bind(this));
+    }
+
+    async responseToDataframeTransformer(response, x, y, z) {
+        const mvt_extent = 4096;
```
elenatorro: constant values, as numbers or strings, should be in capital letters

src/api/source/mvt.js
```diff
@@ -0,0 +1,179 @@
...
+        }
+        let featureGeometries = [];
+        const decodedProperties = {};
+        for (let i = 0; i < mvtLayer.length; i++) {
+            const f = mvtLayer.feature(i);
+            const geom = f.loadGeometry();
+            const mvtGeomType = f.type;
+            if (metadata.geomType === undefined) {
+                switch (mvtGeomType) {
+                    case mvtDecoderGeomTypes.point:
```
elenatorro: ✨

davidmanzanares:
After my last changes (see https://github.com/CartoDB/carto-vl/pull/580 for more detail) I think this is almost done. There are only two things I want to look at before merging:
- The colors are changed for ramps with categories that are not classified with `top` or `buckets`. I don't know the reason.
- When the viz mixes cluster aggregation with un-aggregated properties the behavior has changed. We should fix https://github.com/CartoDB/carto-vl/issues/583 before merging

src/core/renderer.js
```diff
@@ -150,7 +150,7 @@ class Renderer {
         const aspect = this.getAspect();
         let drawMetadata = {
             zoom: 1. / this._zoom,
-            columns: []
+            columns: [] // What is this?
```
jgoizueta: You're the only one who could have known 😆

davidmanzanares:
Ok, the reason for the changed colors is that the loop that iterates through the properties is now `Object.values` which is random :S IMO, this should be fixed

davidmanzanares:
https://github.com/CartoDB/carto-vl/issues/583 fixed

src/api/source/mvt.js
davidmanzanares: Changed

elenatorro:
LGTM

ethervoid:
BRAVOOOOOOO :_)

PR #488, opened 2018-05-31
Admit epoch numbers as argumet of time()

jgoizueta:
It can be useful to be able to treat epoch values (e.g. from a datasource column) as times, for example for filtering or formatting the result of `torque.getSimTime()`.

This will be important if we end up converting all time/date column values to an epoch in MVTs.

### DIFF
--- a/src/core/viz/expressions/time.js
+++ b/src/core/viz/expressions/time.js
@@ -25,7 +25,13 @@ import BaseExpression from './base';
 export default class Time extends BaseExpression {
     constructor(date) {
         if (!(date instanceof Date)) {
-            date = new Date(date);
+            if (typeof(date) === 'number') {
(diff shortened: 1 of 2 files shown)


src/core/viz/expressions/time.js
```diff
@@ -25,7 +25,13 @@ import BaseExpression from './base';
 export default class Time extends BaseExpression {
     constructor(date) {
         if (!(date instanceof Date)) {
-            date = new Date(date);
+            if (typeof(date) === 'number') {
+                const epoch = date;
+                date = new Date(0);
+                date.setUTCSeconds(epoch);
```
IagoLast: I asked pramsey and Postgres has milliseconds precision

Issue #392, opened 2018-05-10
top(): add unit tests and type checking


davidmanzanares:

davidmanzanares:
You just approved the PR two hours ago xD

PR #490, opened 2018-05-31
Feature independent check

davidmanzanares:
The parameters of some expressions don't support expressions that depend on feature properties. Like the duration of a `torque`.

This condition was uncontrolled.

This PR controls the condition and add unit tests to that control.

I fixed here https://github.com/CartoDB/carto-vl/issues/392 too since the problem was very related.

### DIFF
--- a/src/core/viz/expressions/basic/variable.js
+++ b/src/core/viz/expressions/basic/variable.js
@@ -36,6 +36,9 @@ export default class Variable extends BaseExpression {
         super({});
         this.name = name;
     }
+    _isFeatureDependent(){
+        return this.alias? this.alias._isFeatureDependent(): undefined;
+    }
     get value() {
         return this.eval();
     }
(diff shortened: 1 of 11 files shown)


src/core/viz/expressions/basic/variable.js
```diff
@@ -29,13 +29,16 @@ import { checkString } from '../utils';
...
-        checkString('name', 'name', 0, name);
+        checkString('variable', 'name', 0, name);
         if (name == '') {
             throw new Error('variable(): invalid parameter, zero-length string');
         }
         super({});
         this.name = name;
     }
+    _isFeatureDependent(){
+        return this.alias? this.alias._isFeatureDependent(): undefined;
```
IagoLast: why not returning false?

davidmanzanares: Because it is technically undefined, a variable may or may not be feature dependent until we know the aliased expression

davidmanzanares:
Ping :slightly_smiling_face:

src/core/viz/expressions/basic/variable.js
davidmanzanares: Done

davidmanzanares:
Renamed

Issue #130, opened 2018-03-06
Timestamps: proper support for temporal maps (server-side)


davidmanzanares:
Related to https://github.com/CartoDB/renderer-prototype/issues/16

We have support for temporal maps by using numeric properties. However, we don't support native timestamp properties in the Windshaft client.

I think that a good enough implementation would automatically transform a timestamp property into a numeric one (transparently to the user).

I think that a good approach would be to download the minimum and maximum timestamps (as metadata). To later download (in the MVT tiles) the timestamp properties as numbers in the [0,1] range, where 0 means minimum timestamp, and 1 means maximum timestamp.

Since timestamps can have very high resolutions, but most of the times we don't need those, we should have a mechanism for aggregations. After implementing the naive version (without aggregations, setting the timestamp in `dimensions`), we could add new aggregation functions that only works with timestamps, for example: `HOUR(), DAY(), WEEK(), MIN(), SEC(), MSEC())`

jgoizueta:
Interesting fact: Mapnik MVT don't support date/time columns: they're just omitted from the tile:

https://github.com/CartoDB/node-mapnik/blob/v3.6.2-carto/src/mapnik_vector_tile.cpp#L2561-L2593

As a work around we can cast to text `::text`, e.g. in the dimensions expressios

rochoa:
Not a surprise, see https://github.com/CartoDB/Windshaft/issues/103#issuecomment-31860378.

davidmanzanares:
What are the plans for this? https://github.com/CartoDB/renderer-prototype/issues/177 requires to solve this on Maps API or to apply the `::text` workaround

davidmanzanares:
[...]

PR #495, opened 2018-06-01
Add parameter to request dates as numbers

IagoLast:
https://github.com/CartoDB/carto-vl/issues/130
--- 

Maps API was patched to accept a new parameter that allows requesting dates as numbers,
this way we can send dates through `mvt` tiles and we get rid of the double-instantiation process
needed to accept dates.


IagoLast:
I think with this change we can simplify the windshaft file a lot, in fact @elenatorro is working on it ;)

IagoLast:
@davidmanzanares you are right.

A new parameter will be added to the `metadata` response so we can parse `date` fields and the behaviour will be the same

davidmanzanares:
Great! :+1:

Keep in mind that, on top of that, we should keep track of these `date-numeric` properties, since common arithmetic operations like subtraction won't work out of the box anymore. I think that leaving these properties as `date` and adapting the current code properly will solve it.

jgoizueta:
I added the do not merge label because we must wait for backend changes to be deployed

jgoizueta:
We should be aware and document what the epoch conversion implies regarding time zones:

* If the user data (in the database) has a time zone, it will be honoured to compute the correct time, but since the client receives it converted to UTC, the information of what the original time zone was is lost. The client can choose to show the data as UTC, the client's local zone or any other zone, but it will not now what the data specified.
* If the user data doesn't have a time zone we will interpret it as UTC (as was done before).

IagoLast:
IMO date representation in a `i18n` issue so our server and library should provide data in a standard format and is responsibility of the app to display the dates in the locale format so:

- When you save a date in our server the original timezone is lost, by default we assume UTC
- When you get dates from the server you receive a UTC epoch in milliseconds

jgoizueta:
@davidmanzanares take a final look please. Can we merge this?

examples/animation/demo.html
```diff
@@ -0,0 +1,99 @@
...
+      center: [0, 0],
+      zoom: 0,
+      dragRotate: false
+    });
+
+    const animStart = '2018-04-11T12:00:00Z';
+    const animEnd = '2018-04-12T12:00:00Z';
+
+    const source = new carto.source.GeoJSON(generateData(), { dateColumns: ['sim_time'] });
+    function generateData() {
```
davidmanzanares: I would move these examples to `debugging/` or something like that

IagoLast: +1 . `/internal` sounds like a good name to me

jgoizueta: moved to /debug which we can use for this kind of stuff, as MGL does

PR #496, opened 2018-06-01
Refactor Windshaft client - Polygon Decoder

elenatorro:
This is the first change in the process of refactoring [client/windshaft.js](https://github.com/CartoDB/carto-vl/blob/master/src/client/windshaft.js) in order to organize it in different files.

The main idea is to move all the methods related with the decode process to another directory. After reading the documentation, I think it could be something like:

```
|- / src
|-- / client
|--- / mvt
|---- feature-decoder.js
|---- layer-decoder.js
|---- value-decoder.js
```
What do you think?


src/utils/geometry.js
```diff
@@ -0,0 +1,75 @@
+module.exports = {
```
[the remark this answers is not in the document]
elenatorro: ouch!

src/utils/geometry.js
```diff
@@ -0,0 +1,75 @@
...
+    // Add polygon composed by preClippedVertices to the `polygon.flat` array
+    clipPolygon(preClippedVertices, polygon, isHole) {
```
IagoLast: maybe we could move this method to the `Polygon` class as `clip `and do

```js
clipped = polygon.clip(precliped, ishole);
```

maybe with some more refactor we can do only

```js
clipped = polygon.clip();
```

src/utils/geometry.js
```diff
@@ -0,0 +1,78 @@
+function clipPolygon(preClippedVertices, polygon, isHole) {
```
IagoLast: not sure about the need of having this method inside a utility file I think `clipPolygon` is only used in `feature-decoder.js`

IagoLast:
LGTM let @davidmanzanares say the final word

PR #500, opened 2018-06-05
Evaluation of now expression

jgoizueta:
I guess this affects the evaluation of torque in turn. And that could be end up being evaluated when finding features at a location. I'm not sure of the implications, but this works fine with my drum machine.


jgoizueta:
Oh well... some tests are probably now unpredictable. doh! 😧

jgoizueta:
OK, forget it, I see this is not how Now was meant to work

PR #501, opened 2018-06-05
Expose rendering simulation time

jgoizueta:
I'd like to have access to the rendering time like this


jgoizueta:
No need for this, torque.getSimTime is good enough

PR #343, opened 2018-04-25
Sprites

davidmanzanares:
Adds https://github.com/CartoDB/carto-vl/issues/266

Example: http://127.0.0.1:8080/examples/editor/#eyJhIjoidHJhZmZpY19hY2NpZGVudHNfY29weSIsImIiOiIiLCJjIjoibWFtYXRhYWtlbGxhIiwiZCI6Imh0dHBzOi8ve3VzZXJ9LmNhcnRvLmNvbSIsImUiOiJ3aWR0aDogICAkY291bnQvMlxuY29sb3I6IG9wYWNpdHkoIHJhbXAobGluZWFyKCRjb3VudCwgMCwxMjApLCBSZWRPciksICRjb3VudC8yMClcbnN5bWJvbFBsYWNlbWVudDogcGxhY2VtZW50KHNpbigkY291bnQqbm93KCkpICwgY29zKCRjb3VudCpub3coKSkpXG5zeW1ib2w6IHNwcml0ZSgnLi4vZXhwcmVzc2lvbnMvbWFya2VyLnN2ZycpXG5cbiAiLCJmIjp7ImxuZyI6LTEwNC45NTkxMjEwNzMzMzIzMSwibGF0IjozOS43MTA4OTc3NzM4NjE4fSwiZyI6MTAuODA1NTQxOTIyNzc4NzUzfQ==

To be done on this PR:
- [x] `symbol:`
- [x] `symbolPlacement`
- [x] Support multi-sprite selection based on category
- [x] complete the refactor of the shader compilation process to remove unnecessary classes and allow for the generation of one shader from multiple expressions
- [x] Add render tests
- [x] Fix the re-render issue on sprites
- [x] Fix the render tests scripts to avoid leaking the http-server
- [x] Review the entire PR (myself included)
- [x] Add reference docs

To be done on future PRs:
- Add a guide
- Discuss cross-origin problems and possible solutions/workarounds
- Discuss `color` and `strokeColor` semantics
- Maybe, provide some default sprites
- Improve interactivity for sprites

PR #503, opened 2018-06-05
Simplify ramp constructor

IagoLast:
This simplifies the logic, notice also that when a array of sprites is passed as palette
we got the following wrong error:

Palettes must be formed by constant expressions, they cannot depend on feature properties


davidmanzanares:
Check tests, pls

IagoLast:
blocked by https://github.com/CartoDB/carto-vl/pull/343

IagoLast:
outdated

PR #527, opened 2018-06-07
Torque clarification

jgoizueta:
This is my attempt to make some concepts more clear and accurate.

I've also eliminated mentions to the range of cyclic torque values being 0 to 1 because the user sees them (through `getSimTime()`)as the limit of the linear expression or the global min/max of a property.

@makella can you take a look and see if I succeeded? It's possible that I made things *less clear* 😛


makella:
@jgoizueta thank you so much for going through! Looks good and super appreciate it!!

PR #538, opened 2018-06-08
Optimize interactivity

davidmanzanares:
This PR optimizes interactivity in two ways:

## Don't do point in polygon computations if they are not needed.

Two tests are changed because they assumed that for example, a `featureEnter` event shouldn't be fired if the cursor is moved within the feature, even if the `.on` method was just called. The new semantics will fire the `featureEnter` if the user starts listening with cursor already on the feature.

## Results

For Ruben's map the performance increases from 1.7 fps to 23 fps: a 13.5x improvement

### DIFF
--- a/src/api/interactivity.js
+++ b/src/api/interactivity.js
@@ -101,6 +101,7 @@ export default class Interactivity {
         }
         preCheckLayerList(layerList);
         this._init(layerList);
+        this._numListeners = {};
     }
 
     /**
@@ -114,6 +115,7 @@ export default class Interactivity {
      */
     on(eventName, callback) {
         checkEvent(eventName);
+        this._numListeners[eventName] = (this._numListeners[eventName] || 0) + 1;
         return this._emitter.on(eventName, callback);
     }
 
@@ -128,6 +130,7 @@ export default class Interactivity {
      */
     off(eventName, callback) {
         checkEvent(eventName);
+        this._numListeners[eventName] = this._numListeners[eventName] - 1;
         return this._emitter.off(eventName, callback);
     }
 
@@ -148,6 +151,10 @@ export default class Interactivity {
     }
 
     _onMouseMove(event) {
+        if (!this._numListeners['featureEnter'] && !this._numListeners['featureHover'] && !this._numListeners['featureLeave']) {
+            return;
+        }
+
         const featureEvent = this._createFeatureEvent(event);
         const currentFeatures = featureEvent.features;
 
@@ -178,6 +185,10 @@ export default class Interactivity {
     }
 
     _onClick(event) {
+        if (!this._numListeners['featureClick'] && !this._numListeners['featureClickOut']) {
+            return;
+        }
+
         const featureEvent = this._createFeatureEvent(event);
         const currentFeatures = featureEvent.features;
 
--- a/src/api/interactivity.js
+++ b/src/api/interactivity.js
@@ -32,7 +32,7 @@ import { wToR } from '../client/rsys';
  */
 
 /**
- * featureEnter events are fired when the user moves the cursor and the movement implies that a non-previously hovered feature is now under the cursor.
+ * featureEnter events are fired when the user moves the cursor and the movement implies that a non-previously hovered feature (as reported by featureHover or featureLeave) is now under the cursor.
  * The list of features that are now behind the cursor and that weren't before is provided.
  *
  * @event featureEnter
@@ -50,7 +50,7 @@ import { wToR } from '../client/rsys';
  */
 
 /**
- * featureLeave events are fired when the user moves the cursor and the movement implies that a previously hovered feature is no longer behind the cursor.
+ * featureLeave events are fired when the user moves the cursor and the movement implies that a previously hovered feature (as reported by featureHover or featureEnter) is no longer behind the cursor.
  * The list of features that are no longer behind the cursor and that were before is provided.
  *
  * @event featureLeave
--- a/src/core/dataframe.js
+++ b/src/core/dataframe.js
@@ -189,26 +189,14 @@ export default class Dataframe {
         // The viewport is in the [-1,1] range (on Y axis), therefore a pixel is equal to the range size (2) divided by the viewport height in pixels
         const widthScale = (2 / this.renderer.gl.canvas.clientHeight) / this.scale * this.renderer._zoom;
         const columnNames = Object.keys(this.properties);
-        const vizWidth = viz.width;
+        let scale = this._computeScale(columnNames, featureIndex, viz.width, widthScale);
         // Linear search for all features
         // Tests triangles instead of polygons since we already have the triangulated form
         // Moreover, with an acceleration structure and triangle testing features can be subdivided easily
-        let scale;
-        let computeScale = ()=>{
-            const f = {};
-            columnNames.forEach(name => {
-                f[name] = this.properties[name][featureIndex];
-            });
-            // Line with is saturated at 336px
-            const lineWidth = Math.min(vizWidth.eval(f), 336);
-            // width is a diameter and scale is radius-like, we need to divide by 2
-            scale = lineWidth / 2 * widthScale;
-        };
-        computeScale();
         for (let i = 0; i < vertices.length; i += 6) {
             if (i >= breakpoints[featureIndex]) {
                 featureIndex++;
-                computeScale();
+                scale = this._computeScale(columnNames, featureIndex, viz.strokeWidth, widthScale);
             }
             const v1 = {
                 x: vertices[i + 0] + normals[i + 0] * scale,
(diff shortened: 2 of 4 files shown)


src/core/dataframe.js
```diff
@@ -193,18 +193,23 @@ export default class Dataframe {
         // Linear search for all features
         // Tests triangles instead of polygons since we already have the triangulated form
         // Moreover, with an acceleration structure and triangle testing features can be subdivided easily
-        for (let i = 0; i < vertices.length; i += 6) {
-            if (i >= breakpoints[featureIndex]) {
-                featureIndex++;
-            }
+        let scale;
+        let computeScale = ()=>{
```
IagoLast: I think If we extract this to a function with parameters instead inline defining it the code will look cleaner.


```js
 _computeScale(columnNames, featureIndex, vizWidth, widthScale) {
        const f = {};
        columnNames.forEach(name => {
            f[name] = this.properties[name][featureIndex];
        });
        // Line with is saturated at 336px
        const lineWidth = Math.min(vizWidth.eval(f), 336);
        // width is a diameter and scale is radius-like, we need to divide by 2
        return lineWidth / 2 * widthScale;
    }
```

davidmanzanares: You would need to have two `computeScale` functions because the code for lines is different than the code for polygons.

test/integration/user/api/interactivity.test.js
```diff
@@ -225,15 +225,15 @@ describe('Interactivity', () => {
                 });
 
                 it('should not fire a featureEnter event when the mouse is moved inside the same feature', done => {
```
IagoLast: Does this description need to be updated?

test/integration/user/api/interactivity.test.js
```diff
@@ -225,15 +225,15 @@ describe('Interactivity', () => {
...
+                        interactivity.on('featureEnter', featureEnterSpy);
                         interactivity.on('featureHover', () => {
-                            expect(featureClickOutSpy).not.toHaveBeenCalled();
+                            expect(featureEnterSpy).toHaveBeenCalledTimes(1);
                             done();
                         });
+                        // Move mouse inside a feature 1
+                        util.simulateMove({ lng: 5, lat: 5 });
                         // Move mouse inside the same feature 1
                         util.simulateMove({ lng: 5, lat: 15 });
```
IagoLast: we never get here since `done` is called after the first `simulateMove` I think we should call the first simulateMove  before setting the featureHover callback

test/integration/user/api/interactivity.test.js
```diff
@@ -256,13 +256,13 @@ describe('Interactivity', () => {
...
                     onLoaded(() => {
-                        // Move mouse inside a feature 1
-                        util.simulateMove({ lng: 5, lat: 5 });
                         interactivity.on('featureLeave', event => {
                             expect(event.features[0].id).toEqual(-0);
                             expect(event.features[0].layerId).toEqual('layer1');
                             done();
                         });
+                        // Move mouse inside a feature 1
+                        util.simulateMove({ lng: 5, lat: 5 });
```
IagoLast: same

test/integration/user/api/interactivity.test.js
davidmanzanares: No, the code changes just to start listening to the events before the user actions.

test/integration/user/api/interactivity.test.js
davidmanzanares: Yes :+1:

src/core/dataframe.js
IagoLast: is the same it only changes the name of `vizWidth/vizStrokeWidth` parameter. Rename it to `width` and we got a generic one :trollface:

test/integration/user/api/interactivity.test.js
davidmanzanares: fixed

src/api/interactivity.js
```diff
@@ -114,6 +115,7 @@ export default class Interactivity {
      */
     on(eventName, callback) {
         checkEvent(eventName);
+        this._numListeners[eventName] = (this._numListeners[eventName] || 0) + 1;
```
elenatorro: I would suggest to initialize all the listeners to 0 in the constructor (since we only have five events we are listening) to avoid this check.

src/core/dataframe.js
```diff
@@ -189,22 +189,15 @@ export default class Dataframe {
         // The viewport is in the [-1,1] range (on Y axis), therefore a pixel is equal to the range size (2) divided by the viewport height in pixels
```
elenatorro: Hi :) I think we can use this PR to do a quick refactor here. `computeScale` method is duplicated (the only difference is that for lines it uses `vizStrokeWidth` and for polygons `vizWidth`). Also, `_getLinesAtPosition` and `_getPolygonsAtPosition` are also similar, so I propose the following refactor:

1. Add a function to get the scale value instead of modifying it:

```js
    _getScale(lineWidth, featureIndex, columnNames, widthScale) {
        const LINE_SATURATION_PIXELS = 336;
        const features = {};
        
        columnNames.forEach(name => {
            features[name] = this.properties[name][featureIndex];
        });

        const widthDiameter = Math.min(lineWidth.eval(features), LINE_SATURATION_PIXELS);
        const widthRadius = widthDiameter / 2;
        
        return widthRadius * widthScale;
    }
```

2. Change `_addFeatureToArray` to `_getFeatureProperties` and return only the feature properties, in order to avoid modifying the array in this function.

```js
    _getFeatureProperties(featureIndex) {
        // same code
        
        return { id, properties };
    }
```

3. Add a function to get the geometry values that are being used in `_getPointsAtPosition`, `_getLinesAtPosition` and `_getPolygonsAtPosition`.

```js
    _getGeometryValues(position) {
        const pos = wToR(position.x, position.y, { center: this.center, scale: this.scale });
        const points = this.decodedGeom.vertices;

        // The viewport is in the [-1,1] range (on Y axis), therefore a pixel is equal to the range size (2) divided by the viewport height in pixels
        const widthScale = (2 / this.renderer.gl.canvas.clientHeight) / this.scale * this.renderer._zoom;
        const columnNames = Object.keys(this.properties);

        return { pos, points, widthScale, columnNames };
    }
```

3. Add a function that shares the behaviour of `_getLinesAtPosition` and `_getPolygonsAtPosition` to avoid having the same code duplicated

```js
    _getLinesAtPosition(position, viz) {
        const lineWidth = viz.width;

        return this._getGeometryAtPosition(position, lineWidth);
    }

    _getPolygonAtPosition(position, viz) {
        const lineWidth = viz.strokeWidth;

        return this._getGeometryAtPosition(position, lineWidth);
    }
```

Where `_getGeometryAtPosition` will include all the changes above:

```js
    _getGeometryAtPosition(position, lineWidth) {
        // Linear search for all features
        // Tests triangles instead of polygons since we already have the triangulated form
        // Moreover, with an acceleration structure and triangle testing features can be subdivided easily
        const features = [];
        const vertices = this.decodedGeom.vertices;
        const normals = this.decodedGeom.normals;
        const breakpoints = this.decodedGeom.breakpoints;
        const { pos, widthScale, columnNames } = this._getGeometryValues(position);
        
        let featureIndex = 0;
        let scale = this._getScale(lineWidth, featureIndex, columnNames, widthScale);

        const TRIANGLES_EDGES = 6;

        for (let i = 0; i < vertices.length; i += TRIANGLES_EDGES) {
            
            if (i >= breakpoints[featureIndex]) {
                featureIndex++;
                scale = this._getScale(lineWidth, featureIndex, columnNames, widthScale);
            }

            const v1 = {
                x: vertices[i + 0] + normals[i + 0] * scale,
                y: vertices[i + 1] + normals[i + 1] * scale
            };
            const v2 = {
                x: vertices[i + 2] + normals[i + 2] * scale,
                y: vertices[i + 3] + normals[i + 3] * scale
            };
            const v3 = {
                x: vertices[i + 4] + normals[i + 4] * scale,
                y: vertices[i + 5] + normals[i + 5] * scale
            };

            const inside = pointInTriangle(pos, v1, v2, v3);
            
            if (inside) {
                const feature = this._getFeatureProperties(featureIndex);
                features.push(feature);
                // Don't repeat a feature if we the point is on a shared (by two triangles) edge
                // Also, don't waste CPU cycles
                i = breakpoints[featureIndex] - TRIANGLES_EDGES;
            }
        }

        return features;
    }
```

What do you think? 😄

davidmanzanares: I already addressed number 3 in https://github.com/CartoDB/carto-vl/pull/539 . I addressed number 1 in this PR as requested by @IagoLast . But on the #539 PR I went back to the inline function, since the refactor number 3 removes that duplication too, and it can be confusing, since that scale is not applicable for points, therefore, it's better to be with the code for polygons and lines only. 

Regarding number 2, I agree, but, in any case, I would do it on #539

elenatorro: 👍

PR #542, opened 2018-06-11
Fix E2E tests

davidmanzanares:
E2E tests were broken due to the `sprite` PR, this PR fixes it.


elenatorro:
Reference images for `mnmappluto`, `pop_density_points` and `sf_stclines` are missing.

davidmanzanares:
But... https://github.com/CartoDB/carto-vl/blob/e176eb03f1ae28649885f9515488915da173a1bf/test/acceptance/e2e/mnmappluto/reference.png

:confused:

test/acceptance/e2e.html.tpl
```diff
@@ -4,7 +4,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <script>
       // This script hides the map attribution when served locally improving screenshot testing precision.
```
IagoLast: This script is not needed anymore.

This was used to conditionally hide the map attribution depending on if the map was displayed on the developer center as an example or we were on a testing env.

Since testing env and examples are separated we could update this to a `<style>` tag where the attribution is always hidden.

test/acceptance/e2e.test.js
```diff
@@ -4,11 +4,27 @@ const util = require('../common/util');
 
 chai.use(require('chai-as-promised'));
 
+const handler = require('serve-handler');
+const http = require('http');
+
 const files = util.loadFiles(path.join(__dirname, 'e2e'));
 const template = util.loadTemplate(path.join(__dirname, 'e2e.html.tpl'));
+const PORT = 5000;
```
IagoLast: maybe this should be defined and exported in the `util.js` file

test/common/util.js
```diff
@@ -43,7 +43,7 @@ function takeReference(file, template, asyncLoad) {
         console.log(`Taking reference from ${getName(file)}`);
         writeTemplate(file, template);
         let options = loadOptions();
-        options.url = `http://localhost:5000/test/integration/render/scenarios${getHTML(getName(file))}/scenario.html`;
+        options.url = `http://localhost:5000/test/${getLocalhostURL(file)}/scenario.html`;
```
IagoLast: ...and we could use `PORT` here

test/acceptance/e2e.html.tpl
davidmanzanares: :+1:

test/acceptance/e2e.test.js
davidmanzanares: :+1:

Issue #337, opened 2018-04-24
Non deterministic render test failures


davidmanzanares:
We've seen a few times now some MGL-integration tests fail randomly on `master` since this is very difficult to reproduce, we should improve the testing framework to show more information. Not just a pixel mismatch.

For example:

- Console errors, console warns
- Uncaught exceptions
- Time, did it timeout?

davidmanzanares:
The lack of good error messages is also a problem when modifying the testing code. For example, while fixing the `e2e` tests I get this error:
```
Failed to load resource: the server responded with a status of 404 (Not Found)
Failed to load resource: the server responded with a status of 404 (Not Found)
Error: Event
    at Page._handleException (/home/dv/github/renderer-prototype/node_modules/puppeteer/lib/Page.js:427:38)
    at CDPSession.Page.client.on.exception (/home/dv/github/renderer-prototype/node_modules/puppeteer/lib/Page.js:103:60)
    at CDPSession.emit (events.js:127:13)
    at CDPSession._onMessage (/home/dv/github/renderer-prototype/node_modules/puppeteer/lib/Connection.js:219:12)
    at Connection._onMessage (/home/dv/github/renderer-prototype/node_modules/puppeteer/lib/Connection.js:119:19)
    at WebSocket.emit (events.js:127:13)
    at Receiver._receiver.onmessage (/home/dv/github/renderer-prototype/node_modules/ws/lib/WebSocket.js:141:47)
    at Receiver.dataMessage (/home/dv/github/renderer-prototype/node_modules/ws/lib/Receiver.js:389:14)
    at Receiver.getData (/home/dv/github/renderer-prototype/node_modules/ws/lib/Receiver.js:330:12)
    at Receiver.startLoop (/home/dv/github/renderer-prototype/node_modules/ws/lib/Receiver.js:165:16)
    at Receiver.add (/home/dv/github/renderer-prototype/node_modules/ws/lib/Receiver.js:139:10)
    at Socket._ultron.on (/home/dv/github/renderer-prototype/node_modules/ws/lib/WebSocket.js:138:22)
    at Socket.emit (events.js:127:13)
    at addChunk (_stream_readable.js:269:12)
    at readableAddChunk (_stream_readable.js:256:11)
    at Socket.Readable.push (_stream_readable.js:213:10)
    at TCP.onread (net.js:598:20)
```
I know that some URLs must be incorrect, but I don't receive information about which one

davidmanzanares:
[...]

PR #543, opened 2018-06-11
337 improve logs

IagoLast:
#337 
---

- Add a favicon to remove 404 error in test console
- Use exquisite to log `error`, `pageerror` and `requestfailed` events


test/common/util.js
```diff
@@ -119,6 +124,12 @@ function handleBrowserConsole(consoleMessage) {
     }
 }
 
+function _onRequestFailed(request) {
+    if (request.failure()) {
+        console.error(`${request.url()} --> ${request.failure().errorText}`);
```
davidmanzanares: I don't have any idea about the API, but, is it ok to call `request.failure()` twice?

IagoLast: I think so https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#requestfailure

IagoLast: first time check if null, then get the text

davidmanzanares: :+1:

Issue #512, opened 2018-06-06
Add some syntax sugar over fade(X, 100000)


davidmanzanares:
Add some syntax sugar over fade(X, 100000), which is used to make accumulative torque

PR #547, opened 2018-06-12
Add fadeIn expression

IagoLast:
#512
--- 

This is a syntax sugar to a fade(x, INFINITY).
Fade expression was extracted to its own file from torque one


IagoLast:
🚫

Issue #531, opened 2018-06-07
Altering the duration of a `torque` filter results in weird time jumps


skgsergio:
In a visualization with a `torque` filter altering the duration create jums in time. I would expect just a decrease/increase of the "speed"/features displayed per second but it also jumps in time.

It feels like it is internally creating a timeline of `duration` seconds and then interpolating the point in that timeline in the range of your time data, when you change the duration it jumps because the point you are on now corresponds to another date of your data.

I'm not sure if this issue addressed or is too much of a change to continue from the date you are at but with the new duration.

![cartovl_time_travel_bug](https://user-images.githubusercontent.com/186349/41123181-1611136e-6a9e-11e8-9aa1-09c3ec0401d0.gif)

## Version
CARTO VL 0.5.0-beta.2

## Example
https://skgsergio.github.io/cartovl/lightning-lines.html

Code: https://github.com/skgsergio/skgsergio.github.io/blob/master/cartovl/lightning-lines.html

PR #549, opened 2018-06-12
Fix change animation duration

elenatorro:
This PR solves two problems:

* https://github.com/CartoDB/carto-vl/issues/531 When the duration of the torque animation changes, now it starts from the position when it has been changed, instead of "jumping" to another position. 

* Also, as @ramiroaznar noticed, the first time the map is instantiated, the animation starts at the initial position (as it is expected).

torque

### DIFF
--- a/src/core/viz/expressions/torque.js
+++ b/src/core/viz/expressions/torque.js
@@ -1,6 +1,6 @@
 import BaseExpression from './base';
 import { implicitCast, DEFAULT, clamp, checkType, checkLooseType, checkFeatureIndependent } from './utils';
-import { number, linear, sub, add, globalMin, globalMax } from '../functions';
+import { number, linear, globalMin, globalMax } from '../functions';
 import Property from './basic/property';
 import Variable from './basic/variable';
 
--- a/src/core/viz/expressions/torque.js
+++ b/src/core/viz/expressions/torque.js
@@ -149,12 +149,11 @@ export class Torque extends BaseExpression {
     }
 
     _setTimestamp(timestamp) {
-        let deltaTime;
-        const speed = 1 / this.duration.eval();
-        
-        if (this._lastTime) {
-            deltaTime = timestamp - this._lastTime;
-        }
+        let deltaTime = 0;
+        const speed = 1 / this.duration.value;
+        const lastTime = this._lastTime || 0;
+    
+        deltaTime = timestamp - lastTime;
 
         this._lastTime = timestamp;
         this.progress.expr = (this.progress.expr + speed * deltaTime) % 1;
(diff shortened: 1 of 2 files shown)


src/core/viz/expressions/torque.js
```diff
@@ -134,24 +135,45 @@ export class Torque extends BaseExpression {
...
-
-        const _cycle = div(mod(now(), duration), duration);
-        super({ _input: input, _cycle, fade, duration });
+        
+        const progress = number(0);
+        
+        super({ _input: input, progress, fade, duration });
         // TODO improve type check
         this.duration = duration;
+        this.progress = progress; // 0 to 1
```
davidmanzanares: This line and the 143 line are redundant with the constructor of the base class

src/core/viz/expressions/torque.js
```diff
@@ -134,24 +135,45 @@ export class Torque extends BaseExpression {
...
         this.type = 'number';
         this._originalInput = originalInput;
+        this._lastTime = 0;
```
davidmanzanares: This should go in `_setTimestamp` because the map still needs to load

src/core/viz/expressions/torque.js
```diff
@@ -163,55 +186,54 @@ export class Torque extends BaseExpression {
...
-            // Check for NaN
-            if (_input <= 0.0 || 0.0 <= _input){
-                x = 1.- clamp(abs(_input-cycle)*duration/(_input>cycle? fadeIn: fadeOut), 0.,1.);
+            #ifndef TORQUE
+            #define TORQUE
+            float torque(float _input, float progress, float duration, float fadeIn, float fadeOut){
+                float x = 0.;
+                // Check for NaN
+                if (_input <= 0.0 || 0.0 <= _input){
+                    x = 1.- clamp(abs(_input-progress)*duration/(_input>progress? fadeIn: fadeOut), 0.,1.);
```
IagoLast: can we add whitespace between operators here? just for readability

src/core/viz/expressions/torque.js
```diff
@@ -134,24 +135,46 @@ export class Torque extends BaseExpression {
...
+    _setTimestamp(timestamp) {
+        let deltaTime;
+        const speed = 1 / this.duration.eval();
+        
+        if (this._lastTime) {
+            deltaTime = timestamp - this._lastTime;
+        }
+
+        this._lastTime = timestamp;
+        this.progress.expr = (this.progress.expr + speed * deltaTime) % 1;
```
IagoLast: any special reason to only reasign the expr?

could we just do:

```js
this.progress = number(this.progress.expr + speed * deltaTime) % 1)
```

src/core/viz/expressions/torque.js
```diff
@@ -163,55 +186,58 @@ export class Torque extends BaseExpression {
...
         const max = this.input.max.eval();
 
-        if (!(this.input.min.eval() instanceof Date)) {
-            return min + c * (max - min);
+        if (!(min instanceof Date)) {
+            return progress * (max - min) + min;
         }
 
-
         const tmin = min.getTime();
```
IagoLast: @davidmanzanares @elenatorro is the .getTime() call ok? I think max is already a number

Issue #494, opened 2018-06-01
Add simulation time controls to torque


jgoizueta:
In addition to the existing getSimTime I propose adding several functions to control simulation time:

* `pause()` would stop the animation; getSimTime() would freeze, returning the same value while torque is paused.
* `resume()` resume the animation (from the paused sim. time)
* `setSimTime(t)` advance the cycle to the specified simulation time

That would allow the users to implement UI controls for the animations

davidmanzanares:
:+1:

I would add a `stop()` too.

PR #550, opened 2018-06-13
494 add torque controls

IagoLast:
#494
---

- Add play method
- Add pause method
- Add stop method
- Add `setSimTime` method
- Add example

[https://gyazo.com/70772af31bf24ef7116f7f46eeed81bd](https://gyazo.com/70772af31bf24ef7116f7f46eeed81bd)

> The components used in the example are not definitive. We should update them using `airship` components when ready.


IagoLast:
About setSimTime as A programmer I prefer to put the percentage instead the exact date but maybe we should provide both methods.

What do you think?

davidmanzanares:
I would have a `setSimProgress` with the [0,1] progress and a `setSimTime` with the time in date, or number, whatever the type of input is.

examples/advanced/torque-controls.html
```diff
@@ -0,0 +1,188 @@
...
+        });
+
+        const animStart = '2018-04-11T12:00:00Z';
+        const animEnd = '2018-04-12T12:00:00Z';
+
+        const source = new carto.source.GeoJSON(generateData(), {
+            dateColumns: ['sim_time']
+        });
+
+        function generateData() {
```
davidmanzanares: Can't we just link to one of our datasets on the Carto account?

examples/advanced/torque-controls.html
```diff
@@ -0,0 +1,188 @@
...
+            };
+        }
+
+
+        const viz = new carto.Viz(`
+            @duration: 10
+            @torque: torque($sim_time, @duration, fade(0.2, 0.2))
+            width: linear($sim_time) * 100 + 1
+            strokeWidth: 0
+            filter: @torque`
```
davidmanzanares: any special reason for so many variables?

examples/advanced/torque-controls.html
```diff
@@ -0,0 +1,188 @@
...
+
+        $timeRange.onchange = () => {
+            viz.variables.torque.setSimProgress($timeRange.value);
+        }
+
+        $playBtn.onclick = () => viz.variables.torque.play();
+        $pauseBtn.onclick = () => viz.variables.torque.pause();
+        $stopBtn.onclick = () => viz.variables.torque.stop();
+
+        setInterval(() => {
```
davidmanzanares: Can't we use `layer.on(''updated',...)`?

src/core/viz/expressions/torque.js
```diff
@@ -186,14 +194,13 @@ export class Torque extends BaseExpression {
...
      */
-    
     getSimTime() {
         const progress = this.progress.eval(); //from 0 to 1
         const min = this.input.min.eval();
         const max = this.input.max.eval();
 
         if (!(min instanceof Date)) {
-            return progress * (max - min) + min;
+            return new Date(progress * (max - min) + min);
```
davidmanzanares: This seems wrong to me. If `min` is not an instance of `Date` that should be because the input is numeric. Therefore, the output of `getSimTime` should be a number, not a Date. In particular, it is not possible to do this cast because input could be in seconds, weeks, or years, but we don't know which unit

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
         return new Date(tmix);
     }
 
+    /**
+     * Set the time stamp of the simulation
+     * @api
+     * @memberof carto.expressions.Torque
+     * @instance
+     * @name getSimTime
```
davidmanzanares: **set**

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+     * @param {Date} simulationTime - A javascript Date object with the new simulation time
```
davidmanzanares: `torque` works with numeric (not date) inputs too

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+     */
+    setSimTime(simulationTime) {
+        const tmin = this._input.min.eval();
+        const tmax = this._input.max.eval();
+        if (!isDate(simulationTime)) {
+            throw new TypeError(`torque.setSimTime requires a valid date as parameter but got ${simulationTime}`);
+        }
+        if (simulationTime.getTime() < tmin) {
+            throw new RangeError('torque.setSimTime requires the date parameter to be higher than the minimal date of the dataset');
```
davidmanzanares: It is not the minimum date of the dataset, that's the default behavior, but it can be overridden with `linear`

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+        if (simulationTime.getTime() > tmax) {
+            throw new RangeError('torque.setSimTime requires the date parameter to be lower than the maximun date of the dataset');
+        }
+        this.progress.expr = (simulationTime.getTime() - tmin) / (tmax - tmin);
+    }
+
+    /**
+     * Get the simulation progress.
+     * 
+     * @returns {Number} A number representing the progress. 0 when the animation just started and 1 when complete.
```
davidmanzanares: `complete` doesn't sound cyclic to me

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+     * @api
+     * @instance
+     * @memberof carto.expressions.Torque
+     * @name getSimTime
```
davidmanzanares: progress

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+        return this.progress.value;
+    }
+
+    /**
+     * Set the simulation progress from 0 to 1.
+     * @param {number} progress - A number in the [0-1] range setting the animation progress.
+     * @api
+     * @instance
+     * @memberof carto.expressions.Torque
+     * @name getSimTime
```
davidmanzanares: set...

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+     */
+    setSimProgress(progress) {
+        progress = Number.parseFloat(progress);
```
davidmanzanares: parseFloat? Why?

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+        if (progress < 0 || progress > 1) {
+            throw new TypeError(`torque.setSimTime requires a number between 0 and 100 as parameter but got: ${progress}`);
```
davidmanzanares: setSimProgress

davidmanzanares: `1`, not `100`

test/unit/core/viz/expressions/torque.test.js
```diff
@@ -1,32 +1,80 @@
 import * as s from '../../../../../src/core/viz/functions';
 import { validateTypeErrors, validateStaticType, validateFeatureDependentErrors } from './utils';
 
-describe('src/core/viz/expressions/torque', () => {
+fdescribe('src/core/viz/expressions/torque', () => {
```
davidmanzanares: `fdescribe`

test/unit/core/viz/expressions/torque.test.js
```diff
@@ -1,32 +1,80 @@
...
             t._setTimestamp(0.25);
             expect(t.eval()).toEqual(0.75);
         });
     });
+
+
+
+    describe('.pause', () => {
+        it('should pause the simulation when playing', () => {
+            const t = s.torque(1, 10, s.fade(0));
```
davidmanzanares: With a `0` fade is actually very difficult to don't pass the test. Please, check how the `should eval close to 0.75 when the input is 0 and we have wait a quarter of the animation` test does it

examples/advanced/torque-controls.html
jgoizueta: I prefer this approach for this kind of demo/test example, so it is self-contained (independent of any server data). @davidmanzanares you seem to think this kind of think doesn't fit in the examples folder... maybe we should have a demo/debug/xxx folder?

jgoizueta: (note that I've also added three similar animation demos in other branch)

davidmanzanares: Yes, the examples folder is for documentation purposes to help end-users, we could have a `debugging` folder though. However, the functionality of having play/pause/stop buttons is useful for end-users, so, I think this example, or a similar one, should be added to `examples/`. Just with a better map.

src/core/viz/expressions/torque.js
jgoizueta: when complete -> at the end of the cycle

test/unit/core/viz/expressions/torque.test.js
```diff
@@ -1,32 +1,80 @@
...
+        it('should reset the simulation time', () => {
+            const t = s.torque(1, 10, s.fade(0));
+            t._setTimestamp(0);
+            t._setTimestamp(1);
+            expect(t.getSimProgress()).toEqual(0.1);
+
+            t.stop();
+            expect(t.getSimProgress()).toEqual(0);
+        });
+    });
```
jgoizueta: we also need tests for setSimTime, setProgress, right?

examples/advanced/torque-controls.html
IagoLast: we can ask mamata and create a new issue ;)

examples/advanced/torque-controls.html
IagoLast: nope, the frequency is so damn high so the range-slider events are overridden

src/core/viz/expressions/torque.js
IagoLast: to ensure the parameter is a number and make the `progress < 0 || progress > 1` comparison safe.

we could also check the type and throw an error. what do you prefer?

test/unit/core/viz/expressions/torque.test.js
IagoLast: we are only testing the how the torque progress changes when paused/stopped. I think the fade has no effect on it

test/unit/core/viz/expressions/torque.test.js
IagoLast: fixed :)

test/unit/core/viz/expressions/torque.test.js
davidmanzanares: Having a zero fade is not good. Firstly, it is highly unrealistic since it basically means to not show anything, making torque() an alias to `0`. Secondly, you are checking that the `eval` returns 0 when using a fade that will eval to `0` in basically all circumstances, it is very easy to break the implementation and pass the tests.

examples/advanced/torque-controls.html
```diff
@@ -0,0 +1,188 @@
...
+        <h2>Controls</h2>
+        <div>
+            <input type="button" id="js-play-btn" value="PLAY">
+            <input type="button" id="js-stop-btn" value="STOP">
+            <input type="button" id="js-pause-btn" value="PAUSE">
+        </div>
+        <h2>Duration</h2>
+        <div>
+            <input type="range" id="js-duration-range" min="1" max="60" step="1">
+            <span id="js-duration-span"></span> 10 seconds
```
davidmanzanares: The `10` should be within the span or things like `40 10 seconds` happens

IagoLast: wops

test/unit/core/viz/expressions/torque.test.js
IagoLast: > you are checking that the eval returns 0

where?

examples/advanced/torque-controls.html
IagoLast: I dont know, just copied an existing map

examples/advanced/torque-controls.html
davidmanzanares: You can use

davidmanzanares: You can use:

davidmanzanares: You can use:
```
        let last = $timeRange.value;
        $timeRange.onchange = (ev) => {
            viz.variables.torque.setSimProgress($timeRange.value);
            last = $timeRange.value;
        }
        layer.on('updated', () => {
            if ($timeRange.value == last) {
                $timeRange.value = viz.variables.torque.getSimProgress();
                last = $timeRange.value;
            }
        });
```

examples/advanced/torque-controls.html
davidmanzanares: I would remove them, they are unnecessary complexity

src/core/viz/expressions/torque.js
davidmanzanares: Check the type, please, there is an util for that.

test/unit/core/viz/expressions/torque.test.js
davidmanzanares: https://github.com/CartoDB/carto-vl/pull/550/files/bd83d3d483e0df08f73d27d6bd09b7ce4d20049a#diff-e3c5a624af6a9edc311887d558660ca5R45

src/api/util.js
```diff
@@ -27,17 +27,31 @@ function isNumber(value) {
     return typeof value == 'number';
 }
 
+function isDate(value) {
```
elenatorro: We are not using this function anywhere. I do not know if it was intended to be used [here](https://github.com/CartoDB/carto-vl/pull/550/files#diff-603cdd3c5b0b5e1002450b2c4e338129R40), but if it is not going to be used I would remove it.

src/core/viz/expressions/torque.js
```diff
@@ -203,10 +210,99 @@ export class Torque extends BaseExpression {
...
+     * @name setSimTime
+     * @param {Date|number} simulationTime - A javascript Date object with the new simulation time
+     */
+    setSimTime(simulationTime) {
+        simulationTime = castDate(simulationTime);
+        
+        const tmin = this._input.min.eval();
+        const tmax = this._input.max.eval();
+
+        if (simulationTime.getTime() < tmin) {
```
elenatorro: I would assign `simulationTime.getTime()` to a variable to avoid calling this method three times.

test/unit/core/viz/expressions/torque.test.js
elenatorro: I think we need to cover the case when the animation stops and then it plays again. I am not 100% sure if it is needed, so this is just a suggestion.

src/core/viz/expressions/torque.js
davidmanzanares: No, please, no. "Premature optimization is the root of all evil"

arianaescobar:
I would suggest something like this: 

controls

test/unit/core/viz/expressions/torque.test.js
IagoLast: fixed

Issue #510, opened 2018-06-06
Rendering artifacts between tiles


skgsergio:
Rendering lines with opacity causes some regions with colors blended and, therefore, with a different alpha result.

As the regions are well defined areas seems that it is tiles olverlaping between them.

This is with my torque map:
<img width="760" alt="screen shot 2018-06-06 at 15 25 51" src="https://user-images.githubusercontent.com/186349/41041026-4a680038-699e-11e8-9d6a-f1a1277cd416.png">

And this is the same dataset without torque filter:
<img width="321" alt="screen shot 2018-06-06 at 15 27 59" src="https://user-images.githubusercontent.com/186349/41041029-4b77aeb0-699e-11e8-8f6a-eec6c8316496.png">

Source: https://gist.github.com/skgsergio/51d3527e9ebf4ee46bd7f6451cf7f825

davidmanzanares:
I just found the problem, we are using a small buffer size, but we are only clipping polygons, not lines.

PR #562, opened 2018-06-15
Set buffersize to 0

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/510

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -339,7 +339,7 @@ export default class Windshaft {
         const LAYER_INDEX = 0;
         const mapConfigAgg = {
             buffersize: {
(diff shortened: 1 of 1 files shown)


jgoizueta:
So... our current polygon outlining relies solely on removing segments aligned with tile borders, right? (and we must be very tile-grid-unfriendly)

davidmanzanares:
Yes, our current ~chapuza~ awesome work in progress relies on removing aligned segments. In the future, we can use the buffersize as the definitive solution, for that we'll need to clip lines and points (polygon clipping is already implemented) and to pass the information of clipped vertices to the next stage (dataframe.js).

Issue #518, opened 2018-06-06
Default cursor doesn't help user notice interactive features


pramsey:
As seen in the [interactivity example](https://carto.com/developers/carto-vl/examples/maps/guides/interactivity/popups.html) the default cursor is the "pan hand" which doesn’t change as it transits through the features. This (a) means there’s no affordance saying “hey, you could click here” and (b) the cursor doesn’t provide a good tight target showing exactly where the click will hit (the “pan hand” isn’t made for targeted clicking).

We could say "hey devs, this is a problem you all must fix", ala @jesusbotella 
```javascript
 const mapboxGLCanvas = document.querySelector('.mapboxgl-canvas');
    
    interactivity.on('featureEnter', featureEvent => {
      mapboxGLCanvas.style.cursor = 'pointer';
    });

    interactivity.on('featureLeave', featureEvent => {
      mapboxGLCanvas.style.cursor = 'default';
    });
```
Or maybe if there's a featureEnter event live the library could do it for the developer automagically.

davidmanzanares:
Yes, I think that the magic solution would be the best option. Good snippet!

@arianaescobar @makella Do you agree on adding this behavior?

IagoLast:
+1

jesusbotella:
I do vote for the magic solution, especially because the event has a huge delay when you have so many polygons (as we did in our hackarto map), so the cursor is set after a few seconds of waiting and is not responsive at all.

davidmanzanares:
Hmu... that delay shouldn't be of a few seconds. We know that the interactivity code is basically not optimized. But, we didn't run into performance problems. It would be good to have an example to test it, confirm it, and do the acceptance after a fix. Therefore, could you open a new ticket with that, including the code? Thank you @jesusbotella !

IagoLast:
I dont like the idea of multiple events changing the style from a performance point of view. Maybe a static css overriding the mapbox default one and then let the user manually update the cursor on feature hover would be our best option. 

We can also give some magic:

```js
// Changes the cursor when the user moves the mouse over a feature, default: `false`
carto.dynamicCursor = true;
````

IagoLast:
wops, I just realized we are going to need both events in every case... maybe it's time to think about some optimizations for the interactivity

arianaescobar:
I completely agree. I actually asked @Jesus89 about this months ago and he told me it was a Mapbox limitation. But if we can make the cursor change as it should, let's definitely do it.

[...]

PR #578, opened 2018-06-19
518 change default cursor

IagoLast:
#518
---

This PR adds a new parameter to the Interactivity object that allows controlling de cursor when the user points to a feature.

By default the cursor will be [pointer](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) and the user can override this by pasing an argument when creating the interactivity.

```js
// Cursor wont change on mouse hover
const interactivity = new carto.Interactivity(layer, {defaultCursor: true});

// Cursor will change on mouse hover
const interactivity = new carto.Interactivity(layer, {defaultCursor: false});

// Cursor will change on mouse hover (default)
const interactivity = new carto.Interactivity(layer);
```

### DIFF
--- a/src/api/interactivity.js
+++ b/src/api/interactivity.js
@@ -101,7 +101,6 @@ export default class Interactivity {
         }
         preCheckLayerList(layerList);
         this._init(layerList);
-        this._numListeners = {};
     }
 
     /**
@@ -139,9 +138,17 @@ export default class Interactivity {
         this._layerList = layerList;
         this._prevHoverFeatures = [];
         this._prevClickFeatures = [];
+        this._numListeners = {};
         return Promise.all(layerList.map(layer => layer._context)).then(() => {
             postCheckLayerList(layerList);
             this._subscribeToIntegratorEvents(layerList[0].getIntegrator());
+        }).then(this._setInteractiveCursor.bind(this));
+    }
+
+    _setInteractiveCursor() {
+        this.on('featureHover', event => {
+            // eslint-disable-next-line 
+            map.getCanvas().style.cursor = event.features.length ? 'pointer' : ''; // map is know at runtime
         });
     }
 
--- a/src/api/interactivity.js
+++ b/src/api/interactivity.js
@@ -76,7 +76,7 @@ export default class Interactivity {
     *
     * @param {carto.Layer|carto.Layer[]} layerList - {@link carto.Layer} or array of {@link carto.Layer}, events will be fired based on the features of these layers. The array cannot be empty, and all the layers must be attached to the same map.
     * @param {object} [options={}] - Object containing interactivity options
-    * @param {boolean} [options.defaultCursor=false] - A boolean flag indicating if the cursor should change when the mouse is over a feature.
+    * @param {boolean} [options.autoChangePointer=true] - A boolean flag indicating if the cursor should change when the mouse is over a feature.
     * 
     * @example
     * const interactivity = new carto.Interactivity(layer);
@@ -96,7 +96,7 @@ export default class Interactivity {
     * @memberof carto
     * @api
     */
-    constructor(layerList, options={}) {
+    constructor(layerList, options = { autoChangePointer: true }) {
         if (layerList instanceof Layer) {
             // Allow one layer as input
             layerList = [layerList];
@@ -145,7 +145,7 @@ export default class Interactivity {
             postCheckLayerList(layerList);
             this._subscribeToIntegratorEvents(layerList[0].getIntegrator());
         }).then(() => {
-            if (!options.defaultCursor) {
+            if (options.autoChangePointer) {
                 this._setInteractiveCursor();
             }
         });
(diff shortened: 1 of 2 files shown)


src/api/interactivity.js
```diff
@@ -75,7 +75,9 @@ export default class Interactivity {
     * Moreover, the order of the features in the events will be determined by the order of the layers in this list.
     *
     * @param {carto.Layer|carto.Layer[]} layerList - {@link carto.Layer} or array of {@link carto.Layer}, events will be fired based on the features of these layers. The array cannot be empty, and all the layers must be attached to the same map.
-    *
+    * @param {object} [options={}] - Object containing interactivity options
+    * @param {boolean} [options.defaultCursor=false] - A boolean flag indicating if the cursor should change when the mouse is over a feature.
```
davidmanzanares: I would rename the parameter and improve its documentation. It is not clear what is the meaning of true and what is the meaning of false. I would name something like `autoChangePointer`.

src/api/interactivity.js
```diff
@@ -134,14 +135,26 @@ export default class Interactivity {
...
+            if (!options.defaultCursor) {
+                this._setInteractiveCursor();
+            }
+        });
+    }
+
+    _setInteractiveCursor() {
+        this.on('featureHover', event => {
+            const map = this._layerList[0].getIntegrator().map; // All layers belong to the same map
+            map.getCanvas().style.cursor = event.features.length ? 'pointer' : '';
```
davidmanzanares: If there are multiple interactivity objects for the same map and more than one interactivity enables the feature, I think it's gonna misbehave. They will race between each other if there is a feature for one of the interactivities but there is no feature for the other one.

src/api/interactivity.js
elenatorro: I agree ☝️

src/api/interactivity.js
elenatorro: Instead of setting the cursor to `''` by default, I would set it to `'auto'`. 

About what @davidmanzanares said, I did not understand it very well. If there are multiple interactivity objects, at the end the only purpose of this function is to change the cursor pointer by checking if at least there is one feature, right? I am missing how it could fail :( 

Also, since all the layers belong to the same map, would it be possible to have a property 'map' so we can get the map when we need it?

IagoLast: > I think it's gonna misbehave

I think there is no easy way to manage this but advising the users to use the brain and don't create crazy mixes x) 

In any case:

- x interactivities with the same config: no problem
- x interactivities with different configs:
  - no overlapping features: no problem
  - overlapping features: Callbacks will collide and the last executed will win

IagoLast: > Instead of setting the cursor to '' by default, I would set it to 'auto'

We dont want the cursor to be auto, we want the cursor to be the one defined by mapbox (`grab`)

If we do `canvas.style.cursor = 'auto'` we are creating a css more specific css rule that overrides the one given by mapbox leading to a wrong behaviour.

When we do `canvas.style.cursor = ''` we are creating a empty specific rule so the browser applies the one given by mapbox.

elenatorro: 👍

IagoLast: @davidmanzanares I don't understand your proposal

davidmanzanares: 1. Create a per-map `Set` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set ) initialized to empty. This set is going to keep track of the Interactivity objects that want to change the cursor to `pointer`.
2. Each time the hover callback is invoked do:
2.1 Check if there is a feature behind the cursor for its layers.
2.2 If there is a `hit`, then add the Interactivity object to the set, if not, remove the interactivity object from the set
2.3 Set the style cursor to `pointer` if the set object is not empty, set it to the default otherwise.

davidmanzanares:
Now I don't see the problem, can you explain it in more depth? Which is the result?

IagoLast:
The "problem" is that pointer always have precedence so the user cant say:

`cities wont have pointer icon but countries will`

davidmanzanares:
Why not? If the interactivity for cities is configured without the auto-magic it won't trigger its featureHover, but the countries Interactivity should trigger it

IagoLast:
@davidmanzanares

Imagine I wanted the default cursor when the mouse is over Madrid:

 [https://gyazo.com/e139de1e6818c4f8255238ccfc723f7f](https://gyazo.com/e139de1e6818c4f8255238ccfc723f7f)

davidmanzanares:
Ah, yes, but I don't think that's a bug

IagoLast:
ok, so ready to CR 🏳️

src/api/interactivity.js
```diff
@@ -65,7 +65,16 @@ const EVENTS = [
     'featureHover',
     'featureLeave',
 ];
+
+// https://github.com/CartoDB/carto-vl/pull/578#discussion_r196356856
+const _interactivityMap = new Map();
+
 export default class Interactivity {
+    
+    static get _interactivityMap() {
```
davidmanzanares: Unused

Issue #402, opened 2018-05-11
top() bug


davidmanzanares:
top() assumes that the categoryNames variable of the metadata is sorted by count. I think this is true for windshaft sources, but not for Geojson.

davidmanzanares:
This was fixed by https://github.com/CartoDB/carto-vl/pull/580

Issue #576, opened 2018-06-18
MVT Source Metadata definition


davidmanzanares:
Related to https://github.com/CartoDB/carto-vl/pull/487

We need to define a Metadata API for the new MVT source.

## Metadata usage

GeomType:
    - used for selecting style defaults

Column types:
    - Type checking on expression's _compile method

Min, Avg, Max, Sum:
    - Global agg expressions

Sample:
    - Global agg (Percentile) expression

Category count:
    - top() expression

## Proposal

Geomtype:
    - autodiscovery to simplify MVT usage
    - for windshaft, it's on the metadata, so we could speed up start-up time by pre-compaling the shaders before the first tile comes

Column types:
    - autodiscovery to simplify MVT usage
    - for windshaft, it's on the metadata, so we could speed up start-up time by pre-compaling the shaders

Min,Avg,Max,Sum, Sample and Category count:
    - for windshaft, it's on the metadata
    - for MVT, it's optional in the metadata, if it is not present and the user tries to use these functions => Error!

### Required changes

The MVT source should return the metadata once the first tile comes, layer should prepare the source to be able to produce tiles before waiting for the metadata.

decodeProperty should return string, number or undefined. When it returns other than undefined => check type, if undefined => set type on metadata

API for Metadata:
```
{
    properties: {
        wadusProperty: {
            max: 10,
            min: 5,
        },
        wadusCatPropertyName: {
            categories: {
                red: {count: 100},
                blue: {count: 12},
                green: {count: 3},
            }
        },
    },
    sample: [
        { wadusProperty: 6, wadusCatPropertyName: 'blue'},
        { wadusProperty: 7, wadusCatPropertyName: 'red'},
        { wadusProperty: 6, wadusCatPropertyName: 'blue'},
    ]
}
```

jgoizueta:
On one hand I'd like to be able to manually define all the metadata when defining the MVT source (the user can know things we don't know, it would allow to have date properties as we have in GeoJSON, and we could optimize this case as you mention for Windshaft). On the other hand it's clear that the most common case is that the user has no clue/doesn't care so we need to support autodiscovery, and given that it would be a pain to support both (autodiscovery and manual metadata).

jgoizueta:
I think the API is OK (sample is optional, and most users won't provide it)

How could we acomodate in it the info about date properties? (as we do in the GeoJSON)

davidmanzanares:
[...]

PR #580, opened 2018-06-19
Mvt metadata

davidmanzanares:
- Improve the metadata format: https://github.com/CartoDB/carto-vl/issues/576
- Fix multiple MVT bugs
- extract categorizer function
- Fix https://github.com/CartoDB/carto-vl/issues/402
- Manage `null` categories in a good way
- Use metadata class
- Adapt GeoJSON
- Fix tests


src/client/windshaft.js
```diff
@@ -322,45 +328,40 @@ export default class Windshaft {
...
+        });
+        Object.values(properties).map(property => {
+            property.type = adaptColumnType(property.type);
+        });
+        const metadata = new Metadata({ properties, featureCount, sample: stats.sample, geomType, isAggregated: aggregation.mvt });
+        Object.keys(properties).forEach(propertyName => {
+            const property = properties[propertyName];
+            if (property.type === 'category' && property.categories) {
+                property.categories.forEach(category => {
+                    metadata.categorizeString(category.category);
```
jgoizueta: I find it confusing that we have to *categorize* strings here, but the Metadata constructor also does some of this...

Issue #595, opened 2018-06-21
Fix docs generation


Jesus89:
There are some consistency issues during the generation of the documentation and in the docs.

1. In [CartoDB/developers](https://github.com/CartoDB/developers) the tag should be used to generate the documentation instead of `master`. @csubira 
2. We should replace the [guide](https://github.com/CartoDB/carto-vl/blob/master/docs/guides/01-getting-started.md) to use a dynamic variable for the version instead of a fixed value, like in [carto.js's guide](https://github.com/CartoDB/carto.js/blob/master/docs/guides/01-quickstart.md) with `%VERSION%`.
3. Replace `https://libs.cartocdn.com/carto-vl/v0.4.0/carto-vl.js` by `../../../../dist/carto-vl.js` and add this rules in the docs generator.

After all these changes we should release the next `0.5.0` to follow the convention of using minor for new features in `0.x`.

Jesus89:
Yes. It is the same as with the examples.

PR #596, opened 2018-06-21
Fix guides

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/595

- Fix version in guides
- Fix bundle in examples' guides


Jesus89:
Yes. Before a release, we should check that docs, guides and examples are working OK.

> With great power comes great responsibility

PR #541, opened 2018-06-11
Avoid duplicated points between tiles

jgoizueta:
Fixes #90
See also #91 

The tiler may repeat points between tiles, see https://github.com/CartoDB/Windshaft-cartodb/issues/889 and fixing that in the tiler may have a performance penalty.

So this avoids the problem by excluding points with tile coordinates out of `[-1, +1)` when decoding the MVT.

PR #598, opened 2018-06-21
Fix clipping

jgoizueta:
Finally 🤞 polygon clipping seems really correct, removing only the segments clipped (on the tile border) and nothing else.

@dmanzanares's bitwise mask are reintroduced to do the trick.


Jesus89:
:pray:

jgoizueta:
Points should be also OK if we combine this with #541
As for lines... I guess not 😓

jgoizueta:
Hey, let's merge #607 and #609 into this branch first, then we should be able to use (a little) buffersize, so we also fix the dissapearing point problem

src/client/mvt/feature-decoder.js
```diff
@@ -161,15 +162,102 @@ function _getPreClippedVertices(geom, mvtExtent) {
...
+    }
+    let point0 = line[0];
+    let type0 = clipType(point0);
+    for (let i=1; i<line.length; ++i) {
+        let point1 = line[i];
+        let type1 = clipType(point1);
+        const nextType = type1;
+        const nextPoint = point1;
+
+        for(;;) {
```
IagoLast: ?

src/client/mvt/feature-decoder.js
```diff
@@ -85,24 +105,32 @@ export function clipPolygon(preClippedVertices, polygon, isHole) {
...
+            if (clippedTypes2[0]) {
+                clippedTypes2[preClippedVertices2.length] = clippedTypes2[0];
+            }
             preClippedVertices2.push(preClippedVertices2[0]);
         }
         preClippedVertices = preClippedVertices2;
+        clippedTypes = clippedTypes2;
     }
 
     if (preClippedVertices.length > 3) {
```
elenatorro: Why 3?

src/client/mvt/feature-decoder.js
```diff
@@ -130,15 +162,102 @@ function _getPreClippedVertices(geom, mvtExtent) {
...
+function clipLine(line) {
+    // linestring clipping based on the Cohen-Sutherland algorithm
+    // input is a single linestring [point0, point1, ...]
+    // output is an array of flat linestrings:
+    // [[p0x, p0y, p1x, p1y, ...], ...]
+    let clippedLine = [];
+    const clippedLines = [];
+    function clipType(point) {
+        let type = 0;
+        for (let i = 0; i < 4; i++) {
```
elenatorro: Why 4?

src/client/mvt/feature-decoder.js
elenatorro: What does `for(;;)` do 😳?

src/client/windshaft.js
```diff
@@ -335,7 +335,7 @@ export default class Windshaft {
         const LAYER_INDEX = 0;
         const mapConfigAgg = {
             buffersize: {
-                'mvt': 0
+                'mvt': 1
```
elenatorro: Quotes not needed in 'mvt'

src/client/mvt/feature-decoder.js
```diff
@@ -53,26 +53,46 @@ export function isClockWise(vertices) {
...
-        (a, b) => geometryUtils.intersect(a, b, [1, -10], [1, 10]),
-        (a, b) => geometryUtils.intersect(a, b, [-10, 1], [10, 1]),
-        (a, b) => geometryUtils.intersect(a, b, [-1, -10], [-1, 10]),
-        (a, b) => geometryUtils.intersect(a, b, [-10, -1], [10, -1]),
-    ];
+
+    let clippedTypes = {};
 
     // for each clipping edge
     for (let i = 0; i < 4; i++) {
```
elenatorro: Why 4?

src/client/mvt/feature-decoder.js
jgoizueta: It's an infinite loop 😁 I had used `white(true)` originally but lint didn't like it.
The loop repeats until the lines need not be clippend with any of the sides (right, top, left, bottom)

src/client/mvt/feature-decoder.js
jgoizueta: These are the edges 0=right, 1=top, 2=left, 3=bottom as defined by `clippingEdges` and `clippingEdgeIntersectFn`.

src/client/mvt/feature-decoder.js
jgoizueta: With less than 3 vertices we have a degenerate ring (a line or a point), not a proper one, so we ignore it.

src/client/windshaft.js
jgoizueta: true!

src/client/mvt/feature-decoder.js
jgoizueta: Same as before, we need to clip against the 4 edges of the tile

src/client/mvt/feature-decoder.js
jgoizueta: Hey, that was wrong! I think it shoud be >= 3

jgoizueta: ... unless we're accounting for the first vertex being repeated

jgoizueta: Yes, that's right 😅 when we get there we have duplicated the first vertex at the end to close the ring. But if we end up with 3 or less vertices (so we have at most 2 distinct vertices) that must be a degenerate case and we omit it

elenatorro: Ok, I would use then `const MAX_VERTICES = 3` to make it more understandable :)

src/client/mvt/feature-decoder.js
elenatorro: In this case, I would define the different edges to understand this:

`const RIGHT_EDGE = 0`
`const TOP_EDGE = 1`
`const LEFT_EDGE = 2`
`const BOTTOM_EDGE = 3`

jgoizueta: You're removing all the mystery from the classic geometry algorythms :joy:

src/client/mvt/feature-decoder.js
IagoLast: :O didn't know about that.

I any case I think `while(true)` is more legible

PR #600, opened 2018-06-21
Fix example JSON with new examples

davidmanzanares:


Jesus89:
Since we are adding files with a new extension (SVG) we should check that it works in the Developer Center using a tag for testing purposes in staging.

examples/examples.json
```diff
@@ -113,6 +113,21 @@
                 }
             ]
         },
+        {
+            "title": "Sprites",
```
arianaescobar: I'm not sure about creating a separate bucket just for sprites... in my opinion, it should be part of Styling.

examples/examples.json
```diff
@@ -113,6 +113,21 @@
...
+            "samples": [
+                {
+                    "title": "Single Sprite",
+                    "desc": "Show a sprite in the place of points",
```
arianaescobar: Show a sprite **instead** of points.

examples/examples.json
```diff
@@ -113,6 +113,21 @@
...
+                    "file": "sprites/sprite.html"
+                },
+                {
+                    "title": "Multiple Sprites",
+                    "desc": "Show different sprites based on a category",
```
arianaescobar: Show different sprites based on **category values**.

examples/examples.json
```diff
@@ -113,6 +113,21 @@
...
+                    "file": "sprites/sprite-ramp.html"
```
arianaescobar: I would rename it `multiple-sprites.html` so its the same as the title and to another example: `multiple-layers.html`

examples/examples.json
davidmanzanares: :+1:

examples/examples.json
davidmanzanares: :+1:

examples/examples.json
davidmanzanares: :+1:

examples/examples.json
davidmanzanares: :+1:

examples/examples.json
davidmanzanares: `based on a category property`?

arianaescobar: Sure, that works too.

davidmanzanares:
@arianaescobar Yesterday I talked with @csubira about the `desc` property in the JSON, it seems that it is not used in CARTO VL, it is there for historical reasons (CARTO.js). However, the `desc` property is intended to be used for the label on the html, so I'm gonna remove the `desc` property from the JSON, but keep those descriptions and your CR comments on the `.html`.

examples/examples.json
```diff
@@ -140,6 +155,11 @@
                     "title": "Hurricane Harvey",
                     "desc": "Rainfall totals from Hurricane Harvey aggregated to counties. The animation is based on total rainfall. Counties with more rain draw first to lower rainfall totals. The opacity is based off population density (total pop / area). The brighter polygons have higher populations.",
                     "file": "advanced/hurricane-harvey.html"
+                },
+                {
+                    "title": "Torque controls",
```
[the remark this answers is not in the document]
davidmanzanares: :+1:

examples/examples.json
```diff
@@ -140,6 +155,11 @@
...
+                    "desc": "Control torque animations",
+                    "file": "advanced/torque-controls.html"
```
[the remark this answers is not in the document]
davidmanzanares: I disagree, I understand this can be seen as a form of interaction, but I think "interactivity" is for reading back properties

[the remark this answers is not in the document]
arianaescobar: Oh, I just saw this one. I understand your point, but for users not knowing the technology behind as strongly as you do, this makes wayyy more sense in interactivity than here. Also because the examples in Advanced are truly advanced, in terms of story, or styling, or complexity... this map is super simple in comparison (so... not really advanced 😛, just a simple map with controls).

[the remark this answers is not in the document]
davidmanzanares: I'll buy that... :crying_cat_face:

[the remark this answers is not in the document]
arianaescobar: 🙂

PR #608, opened 2018-06-25
Add glsl loader [WIP]

elenatorro:
This PR comes from a leapfrog 🐸.

Instead of writing shader code in strings, when possible, it would be great that we use the `.glsl` file extension. With this, we can use code highlight & linters in our editor.

**Line Vertex Shader example:**

captura de pantalla 2018-06-25 a las 11 43 03

I used [webpack-glsl-loader](https://www.npmjs.com/package/webpack-glsl-loader) plugin after trying different plugins. However, IMO, we could create our own plugin based on this one to have more control (this one was updated the last time 3 years ago).


src/core/shaders/common/antialiasing/antialiasing-vertex-shader.glsl
```diff
@@ -0,0 +1,9 @@
+precision highp float;
+attribute vec2 vertex;
```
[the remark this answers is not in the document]
davidmanzanares: I thought it was `somethingSomething` for others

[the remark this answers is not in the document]
elenatorro: 👍 perfect

IagoLast:
@elenatorro User tests fails because they use a [custom webpack configuration](https://github.com/CartoDB/carto-vl/blob/master/test/integration/user/karma.conf.js#L16) you need to update this configuration with the glsl plugin or the build will fail when requiring those files :)

elenatorro:
Thanks @IagoLast 🎉tests are passing now.

src/core/viz/expressions/base.js
```diff
@@ -53,12 +53,14 @@ export default class Base {
...
             return '';
         }
+        
         return `
-        #ifndef DEF_${this._uid}
-        #define DEF_${this._uid}
-        ${glslCode}
-        #endif
-        `;
+            \n
```
IagoLast: do we need explicit whitespace with string literals?

elenatorro: Without them it does not compile :S

IagoLast: the gpu? or webpack?

davidmanzanares: What do you mean by "explicit whitespace"? :confused:

IagoLast: Javascript string literals capture whitespace chars (newline, tabs...) so:

```js
console.log(`

hello

david
`);
```

is equivalent to:

```js
console.log(`\nhello\n\ndavid`);
```

davidmanzanares: Yes, of course, but... that change must work xD

elenatorro: Without breaklines you end up with something similar to:

```
#endif#ifndef DEF_1
```
And the loader triggers an error. I would have to think a better solution, but if you have any suggestion, is more than welcome :)

IagoLast: My point is I think we dont need explicit `\n` we can just insert a newline

Issue #615, opened 2018-06-26
Red flickering with sprites


davidmanzanares:
Steps to reproduce:
- Open http://127.0.0.1:8080/examples/editor/#eyJhIjoiZmlyZV9wZXJpbWV0ZXJzX2NvcHkiLCJiIjoiIiwiYyI6Im1hbWF0YWFrZWxsYSIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6ICRnaXNfYWNyZXMvMzAwMCAgICAgXG5zeW1ib2w6IHJhbXAobGluZWFyKCRjYXVzZSwgMSwxNCksIFByaXNtKSAqIHNwcml0ZSgnLi4vc3R5bGluZy9mbG93ZXIuc3ZnJykiLCJmIjp7ImxuZyI6LTExOC43MDAxOTQxOTcwODc1MSwibGF0IjozNS42MTMxMTk3NzYyMDI4MX0sImciOjUuNzAyODQ2MDc2MzE1MTEyLCJoIjoiVm95YWdlciJ9
- Put a whitespace in the style
- The sprites will be colored red momentarily

PR #619, opened 2018-06-26
Fix sprite flickering

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/615

The problem is that the blending of two default colors is no longer considered a `default` anymore, therefore, until the blending ends, the color overrides the original sprite color.

The fix is based on avoiding blending between a default and another default, this should be a perf improvement too :smile: .


elenatorro:
👍

Issue #588, opened 2018-06-20
Viewport agg takes into account feature mulitple times


davidmanzanares:
When the same feature is more than one tile the feature will count twice (or N times) in viewport aggregations.

davidmanzanares:
Reopen due to buggy fix

PR #620, opened 2018-06-26
588 remove duplicated features

IagoLast:
#588
---

- Track feature IDS to prevent aggregating the same feature twice.


src/core/dataframe.js
```diff
@@ -126,7 +126,9 @@ export default class Dataframe {
         }
     }
 
-    inViewport(featureIndex, minx, miny, maxx, maxy) {
+    inViewport(featureIndex, scale, center, aspect) {
+        const {minx, miny, maxx, maxy} = this._getBounds(scale, center, aspect);
```
elenatorro: Add spaces between brackets.

src/core/dataframe.js
```diff
@@ -146,6 +148,17 @@ export default class Dataframe {
...
 
+    _getBounds(scale, center, aspect){
+        this.vertexScale = [(scale / aspect) * this.scale, scale * this.scale];
+        this.vertexOffset = [(scale / aspect) * (center.x - this.center.x), scale * (center.y - this.center.y)];
+        const minx = (-1 + this.vertexOffset[0]) / this.vertexScale[0];
+        const maxx = (1 + this.vertexOffset[0]) / this.vertexScale[0];
+        const miny = (-1 + this.vertexOffset[1]) / this.vertexScale[1];
+        const maxy = (1 + this.vertexOffset[1]) / this.vertexScale[1];
+
+        return {minx, maxx, miny, maxy};
```
elenatorro: Same here (*Add spaces between brackets*)

src/core/renderer.js
```diff
@@ -144,73 +144,89 @@ class Renderer {
...
+                }
+                // Ignore features outside viewport
+                if (!dataframe.inViewport(i, scale, this._center, aspect)) {
+                    continue;
+                }
+
+                const feature = this._featureFromDataFrame(dataframe, i);
+
+                // Ignore filtered features
+                if (viz.filter.eval(feature) < 0.5) {
```
elenatorro: What is `0.5`?

src/core/renderer.js
```diff
@@ -144,73 +144,89 @@ class Renderer {
...
+
+        viewportExpressions.forEach(expr => expr._resetViewportAgg());
+
+        // Keep track of features processed in viewPort expressions
+        const processedFeaturesIDs = new Set();
+
+        dataframes.forEach(dataframe => {
+            for (let i = 0; i < dataframe.numFeatures; i++) {
+                // If feature has been acumulated ignore it
+                if (processedFeaturesIDs.has(dataframe.properties.cartodb_id[i])) {
```
elenatorro: I think it is better to have a small function self-explanatory than writing a comment, but that is a personal opinion.

Example:

```js
_isFeatureAccumulated(processedFeaturesIDs, dataframe) {
   return processedFeaturesIDs.has(dataframe.properties.cartodb_id[i]);
} 
```

and 

```js
isFeatureInViewport(dataframe, index, scale, aspect) {
  return dataframe.inViewport(i, scale, this._center, aspect):
}
```

src/core/renderer.js
```diff
@@ -292,7 +304,7 @@ class Renderer {
 
         const s = 1. / this._zoom;
```
elenatorro: Since you have called it `scale` [here](https://github.com/CartoDB/carto-vl/pull/620/files#diff-ba2b399628ae0c09a9b06d41faff9b64R154) you can also refactor this one too :)

test/unit/core/viz/expressions/aggregation/viewportAggregation.test.js
```diff
@@ -19,10 +19,10 @@ describe('src/core/viz/expressions/viewportAggregation', () => {
                 }
             });
             expr._resetViewportAgg();
-            expr._accumViewportAgg({ price: 0, cat: 0, numeric_with_nulls: 0 });
-            expr._accumViewportAgg({ price: 0.5, cat: 1, numeric_with_nulls: 1 });
-            expr._accumViewportAgg({ price: 1.5, cat: 1, numeric_with_nulls: NaN });
-            expr._accumViewportAgg({ price: 2, cat: 2, numeric_with_nulls: 2 });
+            expr.accumViewportAgg({ price: 0, cat: 0, numeric_with_nulls: 0 });
```
elenatorro: great refactor 👍

src/core/renderer.js
davidmanzanares: I disagree for this case, this code is quite performant intensive, I prefer to avoid function calls

IagoLast: I don't see a proper way to handle `_isFeatureAccumulated` because the state lives in the scope...

I think the comment is ok to clarify the intention

elenatorro: Ok, I will buy you that.

Issue #614, opened 2018-06-26
Support multiple layers in MVT


davidmanzanares:
Right now we are using the first layer in an MVT source.

We should add an option to specify the layerId. I propose to add a third parameter to the MVT constructor:

`new carto.source.MVT('https:://wadus.com/tiles/{x}/{y}/{z}.mvt', metadata, 'myLayer')`.

davidmanzanares:
https://github.com/CartoDB/carto-vl/blob/mvt_source_poc/src/api/source/mvt.js#L70

PR #622, opened 2018-06-26
Support multiple layers in MVT

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/614


src/api/source/mvt.js
```diff
@@ -37,15 +38,16 @@ export default class MVT extends Base {
      * @memberof carto.source
      * @IGNOREapi
      */
-    constructor(templateURL, metadata = new Metadata()) {
+    constructor(templateURL, metadata = new Metadata(), layerId = undefined) {
```
IagoLast: layerId = undefined does nothing sine by default parameters are "undefined"

davidmanzanares: I know, but I  put it to be explicit about the "optionalness" of the parameter

IagoLast: ok, but in this case and since it`s a public method we should [update jsdoc too](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

Issue #604, opened 2018-06-22
(Ab)using isStyleLoaded


rochoa:
**ongoing PR**: https://github.com/CartoDB/carto-vl/pull/623

We use `isStyleLoaded` to decide when to load a layer. However, that might be `false` in some situations, e.g. using `fitBounds`, making the layer to not load, not even when the _flight_ finishes.

elenatorro:
Hi @rochoa, how can I reproduce this error? Thanks!

PR #623, opened 2018-06-26
Improve layer load

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/604
Related closed bugfix: https://github.com/CartoDB/carto-vl/commit/bef748208a5fa251f38322259da7b565e5caa6e9

---

**Current solution**

We have two ways to handle this issue. 

1. Use `try catch` to handle mapbox error when style is not loaded. We listen to `map.on('load')` to add the layer if the error has been triggered.

2. Check `map.style._loaded`.
If this property is false, then we listen to `map.on('load')` to add the layer.

Both of them have been tested and they work, but we have chosen the first one because the second one depends on a *private* variable that may change.

---

**Previous research:**

We can not relay on `map.isStyleLoaded()` to safely add a layer.

There is an interesting [Mapbox issue thread](https://github.com/mapbox/mapbox-gl-js/issues/6076) where this problem is being discussed. Highlights:

* From [comment](https://github.com/mapbox/mapbox-gl-js/issues/6076#issuecomment-365834569)
> `load` and `styledata` are fired when the style initially loads, but then never again - even though `isLoaded()` and `isStyleLoaded()` are reporting false as the camera zooms in.

* From [comment](https://github.com/mapbox/mapbox-gl-js/issues/6076#issuecomment-368303979)
> `.loaded()` is basically not useful for determining whether one can safely add a layer, and the user is responsible for adding a `.once("load", () => map.reallyLoaded = true` listener in order to be able to safely add layers later.

Another problem is that when trying to add a layer, mapbox-gl throws an error if the styles are not loaded. That is why we need to check _somehow_ that the styles have been loaded. Since there is not a _safe way_ to do this check, we would probably have to change our fork.


test/unit/api/layer.test.js
```diff
@@ -132,46 +132,46 @@ describe('api/layer', () => {
     describe('.addTo', () => {
         describe('._addToMGLMap', () => {
             let layer;
+            let eventId = 'data';            
+            const event = {
+                dataType: 'style'
+            };
+
+            const mapMock = {
+                isStyleLoaded: () => true,
```
IagoLast: ` jasmine.createSpy('isStyleLoaded').and.returnValue(true);` 🙄

src/api/layer.js
```diff
@@ -376,12 +381,25 @@ export default class Layer {
     }
 
     _addToMGLMap(map, beforeLayerID) {
-        if (map.isStyleLoaded()) {
-            this._onMapLoaded(map, beforeLayerID);
-        } else {
-            map.on('load', () => {
-                this._onMapLoaded(map, beforeLayerID);
-            });
+        map.once('data', (event) => {
```
davidmanzanares: I don't understand why does this work? Does it work? I thought that MGL required the style to be loaded before the `addLayer` method works

elenatorro: MGL requires the style to be loaded if the data event type is sytle. About the `data` event, after reading the documentation I think it is the only change that needs to be listened, and only the first time (that is why I am using `once`). 

https://www.mapbox.com/mapbox-gl-js/api/#map.event:data
> Fired when any map data loads or changes

The examples in the editor work, but I agree it needs further testing.

test/unit/api/layer.test.js
```diff
@@ -131,51 +131,104 @@ describe('api/layer', () => {
...
-            });
-
-            it('should not call onMapLoaded when the map is not loaded', () => {
-                const mapMock = { isStyleLoaded: () => false, on: () => { } };
-                layer._addToMGLMap(mapMock);
-                expect(layer._onMapLoaded).not.toHaveBeenCalled();
-            });
-
-            it('should call onMapLoaded when the map `load` event is triggered', () => {
+            describe('and dataType is style', () => {
```
IagoLast: when datatype is style ?

src/api/layer.js
```diff
@@ -376,12 +376,21 @@ export default class Layer {
     }
 
     _addToMGLMap(map, beforeLayerID) {
-        if (map.isStyleLoaded()) {
+        this._isSourceLoaded = false;
+        
+        map.on('sourcedata', this._onMapSourcedata);
+        map.on('load', this._onMapLoad.bind(this, map, beforeLayerID));
```
IagoLast: I think 

```js
map.on('load', this._onMapLoad.bind(this));
```
is more intuitive since we only bound the context but the parameters

src/api/layer.js
```diff
@@ -376,12 +376,21 @@ export default class Layer {
     }
 
     _addToMGLMap(map, beforeLayerID) {
-        if (map.isStyleLoaded()) {
+        this._isSourceLoaded = false;
+        
+        map.on('sourcedata', this._onMapSourcedata);
```
IagoLast: dont we need to bind `this` here?

src/api/layer.js
```diff
@@ -376,12 +376,21 @@ export default class Layer {
...
+        map.on('load', this._onMapLoad.bind(this, map, beforeLayerID));
+    }
+
+    _onMapSourcedata(event) {
+        this._isSourceLoaded = event.isSourceLoaded;   
+    }
+
+    _onMapLoad(map, beforeLayerID) {
+        if (map.isStyleLoaded() || this._isSourceLoaded) {
```
IagoLast: I think ` this._isSourceLoaded` is always `undefined` because [this](https://github.com/CartoDB/carto-vl/pull/623/files#r199458446) (pun intended) :trollface:

src/Layer.js
```diff
@@ -378,9 +378,9 @@ export default class Layer {
     }
 
     _addToMGLMap(map, beforeLayerID) {
-        if (map.isStyleLoaded()) {
+        try {
             this._onMapLoaded(map, beforeLayerID);
-        } else {
+        } catch (error) {
```
IagoLast: since multiple errors could happen here I'd look for the specific mapbox error and act accordingly

elenatorro: I agree with that, but I am not sure how to check it since `error` is a string and the message may vary also. This is how the error is triggered

```js 
    _checkLoaded() {
        if (!this._loaded) {
            throw new Error('Style is not done loading');
        }
    }
```

IagoLast:
```js
catch(err) {
// We can control the situation when the error is "style not done loading"
  if(err.message === 'Style is not done loading') {
    this._onMapLoaded(map, beforeLayerID);
    return;
  }

// Otherwise a "real" error happened, forward it
throw error;
}

elenatorro: done :)

Issue #534, opened 2018-06-08
Support interactivity for sprites


davidmanzanares:
- [ ] Use the sprite in the interactivity computations
- [ ] Add user test

PR #624, opened 2018-06-26
Add example for interactivity with sprites

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/534

- [x] Add example
- [x] Use the symbol placement offset in the interactivity computations


davidmanzanares:
@rochoa We may want to leave it like this in the short term, the experience is not bad when the markers are squarish or small.

IagoLast:
In my opinion it works ok 🚀

davidmanzanares:
Ok, I remove things from the scope of the PR

Issue #590, opened 2018-06-20
Features list from viewport


rochoa:
Create a method to retrieve the list of features present in the viewport.

danicarrion:
hey, just realized the PR has been merged. Is this something we can already use?

danicarrion:
thanks!!!

padawannn:
I test it with /sources/MVT.html example and It doesn't work.
<img width="1284" alt="captura de pantalla 2018-07-05 a las 16 28 30" src="https://user-images.githubusercontent.com/7164554/42329239-8377e490-8070-11e8-966f-0a3cfce9c9e7.png">

For other types of sources works well

rochoa:
What version/revision are you using?

jgoizueta:
It's a known issue: we cannot use viewportFeatures, viewportAggregation or some interactivity features with MVTs unless they have a cartodb_id property: https://github.com/CartoDB/carto-vl/issues/652

alasarr:
Is there any possibility to tell **viewportFeatures** to return all the properties? In Mastercard we need to get around 400 properties per feature.

padawannn:
@rochoa master branch

alasarr:
@jgoizueta is a possible workaround to include a cartodb_id in the external datasource?

jgoizueta:
Hey, as I've commented in https://github.com/CartoDB/carto-vl/issues/652#issuecomment-402754718 viewport functions should work in the current master, albeit with a bug (that was already there) and that can make features duplicated in the viewport aggregations and features list (if they appear in multiple tiles).

@alasarr: right now it makes no difference if you include a cartodb_id, because of the bug we have. As soon as we fix it we'll add support for having a cartodb_id in the MVT, so adding at the external source will be the right solution to avoid duplicated features

davidmanzanares:
Hey, I just commented the problematic code on `master` since it not only produces that crash with MVT sources, but it is buggy in general (it doesn't avoid repeating duplicated features on viewport aggregations when the features are in multiple tiles due to clipping).

With these lines commented you should be able to work, even though the duplication issue is still there. We'll work on that (https://github.com/CartoDB/carto-vl/issues/588), and on being able to pass or generate ids to MVT sources (https://github.com/CartoDB/carto-vl/issues/652).

jgoizueta:
[...]

PR #625, opened 2018-06-27
Add viewportList expression

jgoizueta:
See #590


jgoizueta:
Pending:
- [ ] validate for intended use cases
- [ ] tests
- [ ] document, add example

IagoLast:
LGTM but I think we can discuss the expression name,

`getViewportFeatures()` sounds good to me :)

elenatorro:
Hmm I do not fully agree with this name, @IagoLast, since we already have other viewport functions that do not start with `get` (`viewportAvg`, `viewportMax`,  `viewportAvgMin`...).

I agree that `list` is not accurate, I would go for `viewportFeatures` or `viewportFeatureList`. How does it sound?

IagoLast:
The thing is this expression has a quite different use case. I don't see any use case where the users will use this expression in the vizSpec, I think the use case is to get a featureList from javascript.

Maybe we should add it by default to the viz as a syntax sugar, something like:

```js
const features  = viz.getVisibleFeatures(); // Internally uses the expression.eval()...
```

debug/list.html
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
```
davidmanzanares: We should add a good example, but that is fine on another PR, @arianaescobar and @makella could help us with that

src/core/viz/expressions/viewportList.js
```diff
@@ -0,0 +1,30 @@
+import BaseExpression from './base';
+
+export class ViewportList extends BaseExpression {
+    constructor() {
+        super({});
+        this.expr = [];
+        // this.type = 'array';
```
davidmanzanares: It should have a type, but `array` is IMO to generic, maybe `featureList`?

davidmanzanares:
The current name is fine for me, but feel free to change it. However, I agree with @elenatorro about the `get` prefix.

Related to @IagoLast last idea, we forgot about Maps API with this PR (blame me :joy: ). This function won't work as expected right now since it won't return unreferenced columns. This is key. Could we just add a list of properties as arguments to this function?

debug/list.html
makella: ummmm yeah, if I could get a brief walkthrough of what this is, that would be great :)

davidmanzanares: Ah, sorry, yes. This a new feature to be able to get the list of the features that are on the viewport like: `[{id: 5, name: 'Madrid', pop: 500000}, {id: 6, name: 'Denver', pop:10000000}]`. Of course, the interesting part is to use that to create widgets

elenatorro:
I would go for the first option:

> only the variables for consistency with interactivity

davidmanzanares:
> only the variables for consistency with interactivity

This won't be very clean since it will require expressions accessing global (viz) state (variables) :confused:

rochoa:
After a quick offline convo, the proposal is to use `viewportFeatures` as any other viewport* expression. In other `viewport*` expressions you have to reference the property you need to compute the value for. E.g., `viewportMax($price)`. We are gonna do the same for `viewportFeatures`, you will reference all the properties you want to access, e.g. `viewportFeatures($price, $rooms, $etc)`.

```js
let viz = new carto.Viz(`
  @max: viewportMax($price)
  @list: viewportFeatures($rooms, $sqm)
  @dens: $dn
  width: 20
  filter: $dn > 150
`);
```

Although this looks unfamiliar, it's consistent, and when you use it with the JS API, it looks much closer to the _usual_ interface. Example:

```js
const list = s.viewportFeatures(s.prop('rooms'), s.prop('sqm'));
viz = new Viz({
    width: 20,
    variables: { list }
})
// ...
layer.on('updated', () => console.log(list.value))
```

This also allows you to have different lists for different properties, example:

```js
let viz = new carto.Viz(`
  @list_2017: viewportFeatures($price2017, rooms2017)
  @list_2018: viewportFeatures($price2018, rooms2018)
  width: 20
`);
```

You could declare new lists dynamically without having to add the required properties to the variables.

jgoizueta:
If added the properties as argument to `viewportFeatures` so we can experiment with it. I think we now need a way to specify *all properties* (maybe when no properties are specified?) But what should *all properties* be? All properties available in the source or all properties referenced in the viz?

IagoLast:
> I think we now need a way to specify all properties

I think this is a nice-to-have (maybe out of the scope of this issue), 

maybe some special token?

```sass
@list: viewportFeatures($*)
@list: viewportFeatures('*')
@list: viewportFeatures('all')
@list: viewportFeatures() // All by default?
```

> All properties available in the source or all properties referenced in the viz?

Good point. Makes sense to you to parametrize this?

```sass
@list: viewportFeatures('DATASET')
@list: viewportFeatures('VIZ')
@list: viewportFeatures($amount)
```

src/core/renderer.js
```diff
@@ -186,7 +182,14 @@ class Renderer {
                     continue;
                 }
 
-                viewportExpressions.forEach(viewportExpression => viewportExpression.accumViewportAgg(feature));
+                viewportExpressions.forEach(viewportExpression => {
+                    if (viewportExpression._requiredProperties) {
+                        const propNames = viewportExpression._requiredProperties.map(p => p.name);
```
davidmanzanares: I think this should go into the viewport agg instead of here, both for the filtering of properties and for the decoding of categories

src/core/viz/expressions/viewportFeatures.js
```diff
@@ -0,0 +1,75 @@
+import BaseExpression from './base';
+import * as schema from '../../../core/schema';
+import Property from './basic/property';
+
+/**
+ * Generates a list of features in the viewport
+ *
+ * For each feature, the properties specified as arguments to this expression will be available.
+ * Filtered features will not be present in the list.
+ * This expression cannot be used in the visualization, it can only be used in JavaScript code as in the example below.
```
davidmanzanares: It cannot be used to render, and type-checking should enforce that, I find the first sentence confusing

src/core/viz/expressions/viewportFeatures.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @param {...Property} properties - properties that will appear in the feature list
+ * @return {Array} list of features
```
davidmanzanares: Returns a ViewportFeatures, the eval returns the array of features.

src/core/viz/expressions/viewportFeatures.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @memberof carto.expressions
+ * @name viewportFeatures
+ * @function
+ * @api
+ */
+export class ViewportFeatures extends BaseExpression {
+    constructor(...properties) {
+        if (properties.some(p => !(p instanceof Property))) {
+            throw new Error('viewportFeatures arguments can only be properties');
```
davidmanzanares: If one property is a variable this check will fail, can we move this check to another entry point (method)?

debug/list2.html
```diff
@@ -0,0 +1,122 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
jgoizueta: Yeah, right, I'll condense them into one

src/core/renderer.js
jgoizueta: Agreed, specially since you're modifying the renderer to cache the features per dataframe in another PR. The problem I see is how to get the metadata in an expression that won't be compiled. I can do without the metadata to get the name of the required properties by using their MNS, but I need also the `IDToCategory` to decode categories... I'll need some help to decide how to do that.

src/core/viz/expressions/viewportFeatures.js
jgoizueta: Oops

jgoizueta:
There's a subtle point with `viewportFeatures` and aggregations: in that case all the properties passed to it will be a added to the GROUP BY of the aggregation, but still, each feature
in the list may actually represent multiple original features (_cdb_feature_count>1) if they all have the same property values and are clustered spatially.

So, the user could be interesting in passing not only properties, but cluster aggregation expressions like this:
```
@list: viewportFeatures($category, clusterSum($value), clusterMax($value))
```

(Perhaps another expression `viewportClusters` should be used for that, and this introduces the problem of how to name those properties)

I don't know if there's a real use case for this, but I guess the same reasons to use cluster aggregations in the viz would apply to  having them in the list.

jgoizueta:
You might still not like it, @davidmanzanares, but take a look at that last commit e487445

I used `_resetViewportAgg` to pass the metadata to the expressions. 
...and `renderLayer` doesn't really have the layer metadata, so I'm assuming all its dataframes have the same (layer) metadata.  

Do you buy it?

src/api/layer.js
```diff
@@ -308,8 +310,11 @@ export default class Layer {
         if (this._viz && this._viz.colorShader) {
             this._renderLayer.viz = this._viz;
             this._integrator.renderer.renderLayer(this._renderLayer);
-            if (this._viz.isAnimated() || this._isUpdated) {
-                this._isUpdated = false;
+            const isSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
```
IagoLast: this method will be nice located in some `utils.js`  file

davidmanzanares:
Seems good to me @jgoizueta

src/api/layer.js
```diff
@@ -342,7 +347,8 @@ export default class Layer {
             this._viz.setDefaultsIfRequired(dataframe.type);
         }
         this._integrator.needRefresh();
-        this._isUpdated = true;
+        this._fireUpdateOnNextRender = true;
+        console.log('to be fired');
```
IagoLast: 👀

src/renderer/viz/expressions/viewportFeatures.js
```diff
@@ -0,0 +1,90 @@
...
+ * });
+ *
+ * @memberof carto.expressions
+ * @name viewportFeatures
+ * @function
+ * @api
+ */
+export class ViewportFeatures extends BaseExpression {
+    constructor(...properties) {
+        if (properties.some(p => !(p instanceof Property))) {
```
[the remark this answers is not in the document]
jgoizueta: Right now it is not supported (we throw an error in the viewportFeatures if all parameters are not strictly properties).
I'm working on that case right now but I'm encountering some difficulties: if the variable is not used in the viz other than in the viewportFeatures it is not resolved (`_resolveAliases`not called). So I prefer to merge this as it is (so it can be tested) and deal with that in another ticket.

src/renderer/viz/expressions/viewportFeatures.js
```diff
@@ -0,0 +1,90 @@
...
+ *         console.log('value:', feature.value, 'category:', feature.category);
+ *     });
+ * });
+ *
+ * @memberof carto.expressions
+ * @name viewportFeatures
+ * @function
+ * @api
+ */
+export class ViewportFeatures extends BaseExpression {
```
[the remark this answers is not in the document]
jgoizueta: I'm going to have to learn about all that modern modules mumble jumble

src/renderer/viz/expressions/viewportFeatures.js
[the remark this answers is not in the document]
davidmanzanares: `_resolveAliasesnot ` is not being called because you didn't put the children in the call to the `super` constructor and the method is not overridden either :laughing:

src/core/viz/expressions/viewportFeatures.js
jgoizueta: See my [comment](https://github.com/CartoDB/carto-vl/pull/625#discussion_r199753163) below.

PR #626, opened 2018-06-27
Simplify decoder

IagoLast:
- Simplify decoder code
- Throw an error when MVT layer has mixed types


src/api/source/mvt.js
```diff
@@ -9,7 +9,7 @@ import { RTT_WIDTH } from '../../core/renderer';
 import Metadata from '../../core/metadata';
 
 // Constants for '@mapbox/vector-tile' geometry types, from https://github.com/mapbox/vector-tile-js/blob/v1.3.0/lib/vectortilefeature.js#L39
-const mvtDecoderGeomTypes = { point: 1, line: 2, polygon: 3 };
+// const mvtDecoderGeomTypes = { point: 1, line: 2, polygon: 3 };
```
davidmanzanares: Why?

src/api/source/mvt.js
```diff
@@ -68,70 +68,64 @@ export default class MVT extends Base {
...
+
+    _decode(mvtLayer, metadata, mvt_extent, decodeFn) {
+        const { properties, propertyNames } = this._getPropertyNames(metadata, mvtLayer.length);
+        const geometries = decodeFn ? [] : new Float32Array(mvtLayer.length * 2);
+
         for (let i = 0; i < mvtLayer.length; i++) {
             const f = mvtLayer.feature(i);
             const geom = f.loadGeometry();
-            const mvtGeomType = f.type;
-            if (metadata.geomType === undefined) {
```
davidmanzanares: The geometry type was being auto-discovered, I think you just removed that functionality.

src/api/source/mvt.js
```diff
@@ -68,70 +68,64 @@ export default class MVT extends Base {
...
-            filter(propertyName => metadata.properties[propertyName].type != 'geometry').
-            map(propertyName => metadata.properties[propertyName].sourceName || propertyName);
-        decodingPropertyNames.forEach(propertyName => {
-            decodedProperties[propertyName] = new Float32Array(mvtLayer.length + RTT_WIDTH);
-        });
+    }
+
+    _decode(mvtLayer, metadata, mvt_extent, decodeFn) {
+        const { properties, propertyNames } = this._getPropertyNames(metadata, mvtLayer.length);
+        const geometries = decodeFn ? [] : new Float32Array(mvtLayer.length * 2);
```
elenatorro: The only thing here, instead of initializing the geometry by checking if there is a `decodeFn`, I would pass it as a parameter:

```js
    _decodeMVTLayer(mvtLayer, metadata, mvt_extent) {
        let geometries;
        
        switch (metadata.geomType) {
            case geometryTypes.POINT:
                geometries = new Float32Array(mvtLayer.length * 2);
                return this._decode(mvtLayer, metadata, geometries, mvt_extent);
            case geometryTypes.LINE:
                geometries = [];
                return this._decode(mvtLayer, metadata, geometries, mvt_extent, decodeLines);
            case geometryTypes.POLYGON:
                geometries = [];
                return this._decode(mvtLayer, metadata, geometries, mvt_extent, decodePolygons);
            default:
                throw new Error('MVT: invalid geometry type');
        }
    }
```

IagoLast:
> This doesn't address the main problem you discussed off-channel, that MVT layer with mixed geometry types case is uncontrolled.

My intention is not to address mixed geometry in MVT, I see that as a new new issue out of the current cycle scope... my point was just to simplify this part.

src/api/source/mvt.js
```diff
@@ -68,70 +68,64 @@ export default class MVT extends Base {
...
-                throw new Error(`Unimplemented geometry type: '${metadata.geomType}'`);
+            else {
+                geometries[2 * i + 0] = 2 * (geom[0][0].x) / mvt_extent - 1.;
+                geometries[2 * i + 1] = 2 * (1. - (geom[0][0].y) / mvt_extent) - 1.;
             }
-            decodingPropertyNames.forEach(propertyName => {
-                const propertyValue = f.properties[propertyName];
-                decodedProperties[propertyName][i] = this.decodeProperty(propertyName, propertyValue);
-            });
+            this._decodeProperties(propertyNames, properties, f, i);
```
elenatorro: Also, IMO, this function should be pure and return a `properties` array instead of modifying it.

```js
properties = this._decodeProperties(propertyNames, properties, f, i);
```

src/api/source/mvt.js
IagoLast: Because the geometry comes in the metadata and currently is the same for all features, at least for windshaft.

For the MVT-Source I dont know the details but I see two options: auto-discover or force the user to indicate the geometry type in the metadata. In any case with the current implementation the geometry type would be the one given by the last feature and I'm not sure if this is fine.

src/api/source/mvt.js
IagoLast: Already given by metadata

src/api/source/mvt.js
davidmanzanares: Yes, it is the case for Windshaft, but it is not the case for general MVT, and yes, if mixed geometry types are present things will be wrong, that's why I said that that's the real problem

IagoLast: > The geometry type was being auto-discovered, I think you just removed that functionality.

Fixed :)

IagoLast: > If mixed geometry types are present things will be wrong

Are you ok if we throw an error (at least while not implement this right)

davidmanzanares: Yes, of course

src/api/source/mvt.js
```diff
@@ -88,6 +91,20 @@ export default class MVT extends Base {
                 throw new Error('MVT: invalid geometry type');
         }
     }
+    
+    _autoDiscoverType(mvtLayer) {
+        const type = mvtLayer.feature(0).type;
```
davidmanzanares: You are assuming that there is one feature, that may not be true.

IagoLast:
> This doesn't address the main problem you discussed off-channel, that MVT layer with mixed geometry types case is uncontrolled.

Now is controlled by throwing an error, do we have an issue to implement this?

davidmanzanares:
No, there is no issue

src/api/source/mvt.js
davidmanzanares: `properties` is a pre-allocated Float32Array, this cannot be pure without having an extra copy

IagoLast:
> No, there is no issue

Fixed: https://github.com/CartoDB/carto-vl/issues/627

src/api/source/mvt.js
```diff
@@ -68,70 +74,90 @@ export default class MVT extends Base {
...
+    _checkType(feature, expected) {
+        const type = feature.type;
+        const actual = MVT_TO_CARTO_TYPES[type];
+        if (actual !== expected) {
+            throw new Error(`MVT: mixed geometry types in the same layer. Layer has type: ${expected} but feature was ${actual}`);
         }
-        return { properties: decodedProperties, points, featureGeometries };
+    }
+
+    _getPropertyNames(metadata, length) {
```
davidmanzanares: Not a good name for a function that allocates the space for multiple arrays

src/api/source/mvt.js
```diff
@@ -68,70 +74,90 @@ export default class MVT extends Base {
...
+                return this._decode(mvtLayer, metadata, mvt_extent, [], decodeLines);
+            case geometryTypes.POLYGON:
+                return this._decode(mvtLayer, metadata, mvt_extent, [], decodePolygons);
+            default:
+                throw new Error('MVT: invalid geometry type');
+        }
+    }
+
+    _autoDiscoverType(mvtLayer) {
+        const type = mvtLayer.feature(0).type;
```
davidmanzanares: The layer could have zero features

IagoLast: wops, fixed [here](https://github.com/CartoDB/carto-vl/pull/626/commits/9031c8e3890714d1dea0a17b4b4d08b998842ca3)

src/api/source/mvt.js
```diff
@@ -68,79 +74,104 @@ export default class MVT extends Base {
...
+        switch (type) {
+            case mvtDecoderGeomTypes.point:
```
elenatorro: I think that now this is `MVT_TO_CARTO_TYPES`, right?

IagoLast: yep but a switch case reveals the intention better 👀

elenatorro:
LGTM

IagoLast:
@davidmanzanares ?

src/api/source/mvt.js
```diff
@@ -83,6 +83,9 @@ export default class MVT extends Base {
 
 
     _decodeMVTLayer(mvtLayer, metadata, mvt_extent) {
+        if (!mvtLayer.length) {
+            return { properties: [], geometries: {} };
```
davidmanzanares: Eing? How does that prevent the problem of accessing the first feature?

IagoLast: because if the layer comes empty the function returns and `_decode` is never called

IagoLast: sorry, I meant:

the funcion `_autoDiscoverType` is never called

PR #628, opened 2018-06-27
Accept non-compliant MVT tiles with wrong first ring orientations

davidmanzanares:


src/client/mvt/feature-decoder.js
```diff
@@ -28,12 +28,24 @@ export function decodeLines(geom, mvt_extent) {
...
+            return;
+        }
+        if (invertedOrientation === undefined) {
+            // According to the MVT spec this condition cannot happen for
+            // MVT spec compliant tiles, but many buggy implementations
+            // don't comply with this rule when generating tiles
+            // Also, other implementations accept this out-of-the-spec condition
+            invertedOrientation = area > 0;
+        }
+        const isExternalPolygon = invertedOrientation ? area > 0 : area < 0;
```
elenatorro: I am a bit confused by this code, I think I am missing something. I think this is the same than: 

```js
let area = signedPolygonArea(geom);

if (area === 0) {
  return;
}

const isExternalPolygon = invertedOrientation !== undefined
  ? area > 0 
  : area < 0;
```

Issue #591, opened 2018-06-20
MVT source with too many attributes


rochoa:
The MVT source—this might apply also to GeoJSON—should handle a lot (~500) of attributes. With our current approach, we don't take care of optimizing the attributes we use for the visualization, degrading the performance as the number of attributes grows.

We need to optimize the way we handle MVT tiles with more attributes than needed, being able to dynamically use just the ones required at any given moment.

davidmanzanares:
![perf-too](https://user-images.githubusercontent.com/1692175/41925840-07aa4f46-796e-11e8-98c1-ecdea40cfa62.png)
Seems our guess was right, the bottleneck is the GPU upload

PR #631, opened 2018-06-27
Lazy gpu properties

davidmanzanares:
This fixes https://github.com/CartoDB/carto-vl/issues/591 by improving the performance cost of the worst case.

However, I'm not very convinced about this, I may actually have decreased performance in the general case.

Includes https://github.com/CartoDB/carto-vl/pull/628


davidmanzanares:
The GPU upload process is no longer the bottleneck:
after-perf

PR #633, opened 2018-06-27
Multiple aggregations per property

### LINKED ISSUE #630: Support multiple cluster aggregations on the same column
In some metadata reorganization we have lost the ability to handle multiple cluster aggregation from the same column (e.g. clusterMax($value), custerMin($value)):

https://github.com/CartoDB/carto-vl/blob/3bef8ed47f0227e5fa30b47def0a6366ba85f34c/src/client/windshaft.js#L363-L366

Now the metadata can contain only a single aggregation per base column. If we try to use more we en up with undefined values for those not present in the metadata.

This breaks maps that use cluster aggregation by simply adding a variable definition with a different cluster aggregation for the same column

jgoizueta:
Fixes #630


src/core/metadata.js
```diff
@@ -34,4 +34,12 @@ export default class Metadata {
         this.numCategories++;
         return this.numCategories - 1;
     }
+    propertyNames(propertyName) {
+        const prop = this.properties[propertyName];
+        if (prop.aggregations) {
+            return Object.keys(prop.aggregations).map(fn => prop.aggregations[fn]);
+        } else {
```
IagoLast: unneeded else

Issue #536, opened 2018-06-08
Viewport agg precision


davidmanzanares:
**Ongoing PR:** https://github.com/CartoDB/carto-vl/pull/635

We are currently checking if a geometry is in the viewport by checking if the bounding boxes of the geometry and the viewport collide.

However, this fails when bounding boxes collide but the geometry does not.

**Example:**

1. It works because both the geometry and the geometry's bounding box collide with the viewport.

![collision](https://user-images.githubusercontent.com/3824953/41974242-b59f4a44-7a17-11e8-9b5a-d4012b5ee77b.png)

2. It does not work because here the geometry does not collide with the viewport.
![no collision](https://user-images.githubusercontent.com/3824953/41974241-b573fc54-7a17-11e8-9cb8-71dc86c5345d.png)

elenatorro:
Merged 🚀 great job, team!

Issue #400, opened 2018-05-11
Improve polygon stroke rendering


Jesus89:
Now line rendering for polygon stroke is very simple. It should be improved to add joins and fix the opacity issues (#738).

![selection_157](https://user-images.githubusercontent.com/2227572/39924236-4c0ac7f8-5527-11e8-8cd2-23d003e62dd2.png)

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/791

PR #635, opened 2018-06-28
Viewport agg precision

elenatorro:
Related issue: https://github.com/CartoDB/carto-vl/issues/536
Includes improvements for: https://github.com/CartoDB/carto-vl/issues/756

**Current status**:

This PR fixes the problem with polygon collision, but it still has to be better for lines. Since we have started to develop this https://github.com/CartoDB/carto-vl/issues/738 and this https://github.com/CartoDB/carto-vl/issues/400, I think we could merge this PR and open a new one to be solved after the lines rendering issues gets fixed. Otherwise, I think we would be duplicating effort.

### Master

captura de pantalla 2018-07-24 a las 10 22 28

### Current

captura de pantalla 2018-07-24 a las 10 17 25

## Checklist
* [x] Add collision algorithm
* [x] Use it for polygons
* [x] Use it for lines
* [x] Testing: Add specs for polygons
* [x] Testing: Add scenarios for 
* [x] ~~Testing: Add specs for lines~~
* [x] Testing: Add scenarios for lines
* [x] Add debug example
* [x] Apply normals to triangle vertices to fix stroke issue.
* [x] Test performance (array of objects vs array of arrays for points)
* [x] Refactor


src/renderer/Dataframe.js
```diff
@@ -135,27 +105,47 @@ export default class Dataframe {
...
+    _isPointInViewport(featureIndex, scale, center, aspect) {
+        const { minx, maxx, miny, maxy } = this._getBounds(scale, center, aspect);
+        const x = this.geom[2 * featureIndex + 0];
+        const y = this.geom[2 * featureIndex + 1];
+        return x > minx && x < maxx && y > miny && y < maxy;
+    }
+
+    _isPolygonInViewport(triangles, viewport) {
+        for (let i = 0; i < triangles.length; i++) {
+            if (triangles[i].collides(viewport)) {
```
davidmanzanares: :confused: The idea of using an AABB is to actually avoid this collision check most of the times, like this:
```
_isPolygonInViewport(triangles, viewport) {
        const result = checkAABB(polygonAABB, viewport);
        if (result !==undefined){
               return result;
       } 
        for (let i = 0; i < triangles.length; i++) {
               ....
        }
}
```

elenatorro: Yes, you are right!

IagoLast:
Maybe tomorrow I'll do some refactor but it's ready for CR

src/core/geometries/collision.js
```diff
@@ -0,0 +1,155 @@
+/**
+ * Determines if two bodies are colliding using the Separating Axis Theorem
+ * https://github.com/Prozi/detect-collisions/
+ */
+export default function SAT(a, b, aabb = true) {
```
IagoLast: since this is a function I think we should rename it to `sat`

test/integration/user/test/featureInViewport.test.js
```diff
@@ -0,0 +1,47 @@
...
+                div.style.height = '500px';
+                div.style.position = 'absolute';
+                document.body.style.margin = '0';
+                document.body.style.padding = '0';
+                document.body.appendChild(div);
+
+                map = new mapboxgl.Map({
+                    container: 'map',
+                    style: 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json',
+                    center: [-6.754536880813703, 35.30378023840868],
```
davidmanzanares: How can I check that this test is actually correct? Center, zoom and feature geometries seem magic

IagoLast: I created a dataset by hand with several squares. Given the map size in pixels the center and the zoom some of them are inside/outside/colliding with viewport so all cases are covered

suggestions to improve this are welcome :)

IagoLast: > How can I check that this test is actually correct?

open the `.html` file with the browser

davidmanzanares: Which `.hmtl`?

davidmanzanares: Regarding the suggestions, I would write a render test instead of a user test. Render tests are standalone and you can actually pan and zoom-in and zoom-out (if you use a Mapbox GL map instead of the simple standalone Map integrator).

IagoLast: strong disagree, render tests are intended to test rendering!

We are testing an expression loading a real scenario and [actually testing](https://github.com/CartoDB/carto-vl/pull/635/files#diff-c07b14a7498127fa2e3a3d9b3064a1d1R37) the value of the expression without dirty hacks. This is faster, more precise, more stable and doesn't get affected by browser-rendering issues.

IagoLast: > Which .hmtl?

1. Change [this line](https://github.com/CartoDB/carto-vl/pull/635/files#diff-c07b14a7498127fa2e3a3d9b3064a1d1R5) and put an `fdescribe` to run only those tests
2. Coment the `afterEach` block to prevent the map to be removed from the browser
3. Run tests in debug mode with: `test:user:watchc``
4. Youll see a karma browser, click on `debug` and you'll see the test case:

screen shot 2018-07-13 at 10 54 08

src/core/geometries/collision.js
```diff
@@ -0,0 +1,50 @@
+import { halfPlaneTest, pointInRectangle, pointInTriangle, checkSign } from '../../utils/geometry';
+
+export function triangleCollides (triangle, viewport, viewportAABB) {
+    if (_isFeatureGreaterThanViewport(triangle, viewport)) {
+        return true;
+    }
+
+    if (_isTriangleInViewport(triangle, viewportAABB)) {
```
jgoizueta: This seems to me a more frequent and more efficient check, so I'd placed it in first place (very minor optimization, but...)

elenatorro: True! ✨

src/renderer/Renderer.js
```diff
@@ -204,9 +205,9 @@ export default class Renderer {
      * Check if the feature at the "index" position of the given dataframe is in the renderer viewport.
      * NOTE: requires `this.aspect` to be set
      */
-    _isFeatureInViewport (dataframe, index, aspect) {
+    _isFeatureInViewport (dataframe, index, aspect, viz) {
         const scale = 1 / this._zoom;
-        return dataframe.inViewport(index, scale, this._center, aspect);
+        return dataframe.inViewport(index, scale, this._center, aspect, viz); // TODO viz
```
jgoizueta: Is that TODO relevant?

src/utils/geometry.js
```diff
@@ -28,9 +28,97 @@ export function perpendicular ([x, y]) {
...
+    const u = normalize([a[0] - b[0], a[1] - b[1]]);
+    const v = normalize([c[0] - b[0], c[1] - b[1]]);
+    const sin = -u[1] * v[0] + u[0] * v[1];
+    if (sin !== 0) {
+        return [(u[0] + v[0]) / sin, (u[1] + v[1]) / sin];
+    }
+}
+
+export function normalize (v) {
+    const s = Math.sqrt(v[0] * v[0] + v[1] * v[1]);
```
jgoizueta: Math.hypot can be used here

src/utils/geometry.js
```diff
@@ -28,9 +28,97 @@ export function perpendicular ([x, y]) {
...
+// Returns true if p is inside the triangle or on a triangle's edge, false otherwise
+// Parameters in {x: 0, y:0} form
+export function pointInTriangle (p, v1, v2, v3) {
+    // https://stackoverflow.com/questions/2049582/how-to-determine-if-a-point-is-in-a-2d-triangle
+    // contains an explanation of both this algorithm and one based on barycentric coordinates,
+    // which could be faster, but, nevertheless, it is quite similar in terms of required arithmetic operations
+
+    if (equalPoints(v1, v2) || equalPoints(v2, v3) || equalPoints(v3, v1)) {
+        // Avoid zero area triangle
+        return false;
```
jgoizueta: return p === v1 ?

src/utils/geometry.js
```diff
@@ -28,9 +28,97 @@ export function perpendicular ([x, y]) {
...
+    };
+    const lengthSquared = diff.x * diff.x + diff.y * diff.y;
+    return lengthSquared <= scale * scale;
+}
+
+export function pointInRectangle (point, bbox) {
+    return ((bbox.minx <= point.x) && (point.x <= bbox.maxx) && (bbox.miny <= point.y) && (point.y <= bbox.maxy));
+}
+
+export function checkSign (positions) {
```
jgoizueta: This can be optimized, but the cost is much uglier code:
```javascript
function checkSign (positions) {
    const n = positions.length;
    if (n === 0) {
        return false;
    }
    const s1 = Math.sign(positions[0]);
    if (s1 === 0) {
        return false; // true originally
    }
    if (n === 1) {
        return true;
    }
    for (let i=1; i < n; ++i) {
        const s2 = Math.sign(positions[i]);
        if (s2 !== s1) {
            return false;
        }
    }
    return true;
}
```

src/renderer/Dataframe.js
```diff
@@ -344,70 +442,80 @@ export default class Dataframe {
...
+                x: vertices[i + 0] + normals[i + 0] * scale,
+                y: vertices[i + 1] + normals[i + 1] * scale
+            }, {
+                x: vertices[i + 2] + normals[i + 2] * scale,
+                y: vertices[i + 3] + normals[i + 3] * scale
+            }, {
+                x: vertices[i + 4] + normals[i + 4] * scale,
+                y: vertices[i + 5] + normals[i + 5] * scale
+            }, {
+                x: vertices[i + 0] + normals[i + 0] * scale,
```
davidmanzanares: We planned to remove this "false" fourth vertex, right?

src/renderer/decoder.js
```diff
@@ -160,33 +165,16 @@ function decodeLine (geom) {
...
+
         breakpoints.push(vertices.length);
     });
+
     return {
         vertices: new Float32Array(vertices),
         breakpoints,
-        normals: new Float32Array(normals)
+        normals: new Float32Array(normals),
+        triangles: []
```
davidmanzanares: Not useful anymore, right?

src/utils/geometry.js
davidmanzanares: Could be, but, if that is true, then the point in triangle for another triangle of the feature should return true. I hope there are no features with just a  zero-sized "triangle"

src/utils/geometry.js
davidmanzanares: Yes, I think it's easier to refactor the caller than this function

src/renderer/Dataframe.js
```diff
@@ -134,28 +139,130 @@ export default class Dataframe {
...
 
-    inViewport (featureIndex, scale, center, aspect) {
-        const { minx, miny, maxx, maxy } = this._getBounds(scale, center, aspect);
+    inViewport (featureIndex, scale, center, aspect, viz) {
+        return this._geometryInViewport(featureIndex, scale, center, aspect, viz);
+    }
+
+    // Add new properties to the dataframe or overwrite previously stored ones.
+    // `properties` is of the form: {propertyName: Float32Array}
+    addProperties (properties) {
```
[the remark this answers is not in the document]
elenatorro: Public methods higher, private methods lower 😂

davidmanzanares:
`debug/test2.mvt` is changed, do you know why?

src/core/geometries/collision.js
```diff
@@ -0,0 +1,51 @@
+import { halfPlaneTest, pointInRectangle, pointInTriangle } from '../../utils/geometry';
+
+export function triangleCollides (triangle, viewport, viewportAABB) {
```
davidmanzanares: Why two viewports?

src/core/geometries/collision.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    if (_isViewportContainedByTriangle(triangle, viewport)) {
+        return true;
+    }
+
+    return _isViewportCollidingTriangleLines(triangle, viewport);
+}
+
+function _isViewportContainedByTriangle (triangle, viewport) {
```
davidmanzanares: `_isAnyViewportVertexInTriangle`

src/core/geometries/collision.js
```diff
@@ -0,0 +1,51 @@
...
+
+    return false;
+}
+
+function _isViewportCollidingTriangleLines (triangle, viewport) {
+    let sign = null;
+
+    for (let i = 0; i < 3; i++) {
+        for (let j = 0; j < viewport.length; j++) {
+            const position = halfPlaneTest(viewport[j], triangle[i], triangle[i + 1]);
```
davidmanzanares: `halfPlaneTest` > `signedDistanceFromPointToLine`?

src/core/geometries/collision.js
```diff
@@ -0,0 +1,51 @@
...
+
+            if (!sign) {
+                sign = Math.sign(position);
+            } else if (sign !== Math.sign(position)) {
+                return false;
```
davidmanzanares: We could add a comment here: `// This triangle line is a separating axis, therefore the triangle doesn't collide with the AABB`

src/renderer/Dataframe.js
```diff
@@ -21,59 +28,55 @@ export default class Dataframe {
...
-                        aabb.maxy = Math.max(aabb.maxy, vertices[2 * i + 1]);
-                    }
-                });
-                this._aabb.push(aabb);
-            });
+        this._aabb = this._computeAABB(geom, type);
+    }
+
+    get widthScale () {
+        return this.renderer
```
davidmanzanares: Didn't we plan to unify this logic into a single function?

src/renderer/Dataframe.js
```diff
@@ -21,59 +28,55 @@ export default class Dataframe {
...
+        switch (type) {
+            case 'point':
+                return [];
+            case 'line':
+            case 'polygon':
+                return this._computeFeatureAABBList(geometry, type);
         }
     }
 
+    _computeFeatureAABBList (geometry, type) {
```
davidmanzanares: I don't think there is a reason to separate this function and `_computeAABB`. The only difference is that the second one early returns if it is of `point` type, which could be done here.

src/renderer/Dataframe.js
```diff
@@ -21,59 +28,55 @@ export default class Dataframe {
...
+    get widthScale () {
+        return this.renderer
+            ? (2 / this.renderer.gl.canvas.clientHeight) / this.scale * this.renderer._zoom
+            : 1;
+    }
+
+    _computeAABB (geometry, type) {
+        switch (type) {
+            case 'point':
+                return [];
```
davidmanzanares: Do we need this `[]`? I expect a hard crash if we try to use an AABB list with points

src/renderer/Dataframe.js
```diff
@@ -134,28 +139,127 @@ export default class Dataframe {
         }
     }
 
-    inViewport (featureIndex, scale, center, aspect) {
-        const { minx, miny, maxx, maxy } = this._getBounds(scale, center, aspect);
+    inViewport (featureIndex, renderScale, center, aspect, viz) {
+        return this._geometryInViewport(featureIndex, renderScale, center, aspect, viz);
```
davidmanzanares: Couldn't we just put the body of `_geometryInViewport` here?

src/renderer/Dataframe.js
```diff
@@ -134,28 +139,127 @@ export default class Dataframe {
...
             default:
                 return false;
         }
     }
 
-    _getBounds (scale, center, aspect) {
-        this.vertexScale = [(scale / aspect) * this.scale, scale * this.scale];
-        this.vertexOffset = [(scale / aspect) * (center.x - this.center.x), scale * (center.y - this.center.y)];
+    _isPointInViewport (featureIndex, renderScale, center, aspect) {
+        const { minx, maxx, miny, maxy } = this._getBounds(renderScale, center, aspect);
```
davidmanzanares: calling `_getBounds` per point seems like a waste of CPU time in a hot spot

src/renderer/Dataframe.js
```diff
@@ -164,106 +268,129 @@ export default class Dataframe {
         return { minx, maxx, miny, maxy };
     }
 
-    _getPointsAtPosition (p, viz) {
-        p = wToR(p.x, p.y, { center: this.center, scale: this.scale });
+    _getViewportPoints (viewportAABB) {
+        return [
```
davidmanzanares: This seems like a smell to me, is there a reason to have this method?

src/renderer/Dataframe.js
```diff
@@ -164,106 +268,129 @@ export default class Dataframe {
...
                 features.push(this.getFeature(featureIndex));
             }
         }
+
         return features;
     }
 
     _getLinesAtPosition (pos, viz) {
-        return this._getFeaturesFromTriangles(pos, viz.width, viz.filter);
+        return this._getFeaturesFromTriangles('line', pos, viz);
```
davidmanzanares: Couldn't the caller just call directly to `_getFeaturesFromTriangles`, avoiding the need of having `_getLinesAtPosition` and `_getPolygonAtPosition` methods?

src/renderer/Dataframe.js
```diff
@@ -164,106 +268,129 @@ export default class Dataframe {
...
-        };
+        let strokeWidthScale;
+
         for (let i = 0; i < vertices.length; i += 6) {
             if (i === 0 || i >= breakpoints[featureIndex]) {
                 featureIndex++;
                 const feature = this.getFeature(featureIndex);
-                if (this._isFeatureFiltered(feature, filterExpression)) {
+
+                if (geometryType === 'polygon' && !pointInRectangle(p, this._aabb[featureIndex])) {
```
davidmanzanares: Why only polygons?

src/renderer/Dataframe.js
```diff
@@ -345,69 +441,93 @@ export default class Dataframe {
...
 
-    // A point is inside a triangle or in one of the triangles edges
-    // if the point is in the three half-plane defined by the 3 edges
-    const b1 = _halfPlaneTest(p, v1, v2) < 0;
-    const b2 = _halfPlaneTest(p, v2, v3) < 0;
-    const b3 = _halfPlaneTest(p, v3, v1) < 0;
+        for (let i = 0; i < numVertices; i += 4) {
+            aabb.minx = Math.min(aabb.minx, vertices[i + 0]);
+            aabb.miny = Math.min(aabb.miny, vertices[i + 1]);
+            aabb.maxx = Math.max(aabb.maxx, vertices[i + 2]);
```
davidmanzanares: This seems wrong to me if you have vertices [x1,y1, x2,y2] you are using the first vertex to compute the minimum and the second one to compute the maximum, but the first one should affect the maximum too, and the second one the minimum too. Go vertex by vertex with `i+=2` and the code will be much easier.

src/renderer/Dataframe.js
```diff
@@ -345,69 +441,93 @@ export default class Dataframe {
...
+            aabb.maxx = Math.max(aabb.maxx, vertices[2 * k + 0]);
+            aabb.maxy = Math.max(aabb.maxy, vertices[2 * k + 1]);
+        }
+
+        return aabb;
+    }
+};
 
-    return (b1 === b2) && (b2 === b3);
+function _getVerticesForGeometry (feature, aabb, geometryType) {
```
davidmanzanares: This naming is not ok, this doesn't get any vertices, it returns and modifies an AABB, the initialization code of the AABB could be inside these functions to simplify things

src/renderer/Dataframe.js
```diff
@@ -345,69 +441,93 @@ export default class Dataframe {
...
 
-// Tests if a point `p` is in the half plane defined by the line with points `a` and `b`
-// Returns a negative number if the result is INSIDE, returns 0 if the result is ON_LINE,
-// returns >0 if the point is OUTSIDE
-// Parameters in {x: 0, y:0} form
-function _halfPlaneTest (p, a, b) {
-    // We use the cross product of `PB x AB` to get `sin(angle(PB, AB))`
-    // The result's sign is the half plane test result
-    return (p.x - b.x) * (a.y - b.y) - (a.x - b.x) * (p.y - b.y);
+function _isFeatureInsideViewport (featureAABB, viewportAABB) {
```
davidmanzanares: _isFeatureAABBInsideViewport

src/renderer/Dataframe.js
```diff
@@ -345,69 +441,93 @@ export default class Dataframe {
...
+    return (featureAABB.minx >= viewportAABB.minx && featureAABB.maxx <= viewportAABB.maxx &&
+            featureAABB.miny >= viewportAABB.miny && featureAABB.maxy <= viewportAABB.maxy);
 }
 
-function _equal (a, b) {
-    return (a.x === b.x) && (a.y === b.y);
+function _isFeatureOutsideViewport (featureAABB, viewportAABB) {
```
davidmanzanares: _isFeatureAABBOutsideViewport

src/renderer/Dataframe.js
```diff
@@ -345,69 +441,93 @@ export default class Dataframe {
...
+                x: vertices[i + 0] + normals[i + 0] * strokeWidthScale,
+                y: vertices[i + 1] + normals[i + 1] * strokeWidthScale
+            }, {
+                x: vertices[i + 2] + normals[i + 2] * strokeWidthScale,
+                y: vertices[i + 3] + normals[i + 3] * strokeWidthScale
+            }, {
+                x: vertices[i + 4] + normals[i + 4] * strokeWidthScale,
+                y: vertices[i + 5] + normals[i + 5] * strokeWidthScale
+            }, {
+                x: vertices[i + 0] + normals[i + 0] * strokeWidthScale,
```
davidmanzanares: Could we remove this fourth vertex?

src/renderer/Dataframe.js
elenatorro: It was not calling `_getBounds` per point I think, but I have refactored this method by passing the viewportAABB (bounds) as a parameter.

src/renderer/Dataframe.js
elenatorro: I moved this logic to `collision`. The only reason to do that was that for some operations (`pointInTriangle` and `signedDistanceFromPointToLine`) we need the viewport coordinates instead of the AABB.

src/renderer/Dataframe.js
elenatorro: I have been checking this issue closer, because there was a test failing. It failed one checking if the line AABB collided with the viewport:

captura de pantalla 2018-07-24 a las 18 14 07

As we can see here, `y` point coordinates are _almost_ 0, so I think we can round the coordinates when calculating the point as follows:

```js
export function wToR (x, y, r) {
    return {
        x: ((x - r.center.x) / r.scale).toFixed(2),
        y: ((y - r.center.y) / r.scale).toFixed(2)
    };
}
```

what do you think?

src/utils/geometry.js
```diff
@@ -28,9 +28,90 @@ export function perpendicular ([x, y]) {
...
+    const b3 = signedDistanceFromPointToLine(p, v3, v1) < 0;
+
+    return (b1 === b2) && (b2 === b3);
+}
+
+// Tests if a point `p` is in the half plane defined by the line with points `a` and `b`
+// Returns a negative number if the result is INSIDE, returns 0 if the result is ON_LINE,
+// returns >0 if the point is OUTSIDE
+// Parameters in {x: 0, y:0} form
+export function signedDistanceFromPointToLine (p, a, b) {
```
jgoizueta: As discussed with @davidmanzanares, the original name of this function was probably better (this isn't  an actual distance)

src/utils/collision.js
```diff
@@ -83,7 +82,7 @@ function _viewportLineSeparatesTriangle (viewportAABB, triangle) {
 function _triangleLineSeparatesViewport (triangle, viewport) {
     for (let i = 0; i < 3; i++) {
         for (let j = 0; j < viewport.length; j++) {
-            const position = signedDistanceFromPointToLine(viewport[j], triangle[i], triangle[i + 1]);
+            const position = signedDistanceFromPointToLine(viewport[j], triangle[i], triangle[(i + 1) % 3]);
```
elenatorro: Good catch!

Issue #599, opened 2018-06-21
Desynchronizated torque layers


skgsergio:
With the last beta release, my map with two torque layers that were in sync now shows desynced.

Both datasets have a timestamp column that has the same range:
`MIN(timestamp) = 2018-06-03T22:15:18Z`
`MAX(timestamp) = 2018-06-03T22:38:09Z`

The first dataset is just points and the second one is a bunch of lines in a `n lines:1 point` relation, all n lines for a point has the same timestamp.

In the following GIF you can observe the issue (top image) and the expected result (bottom image) that is the behavior of the previous beta release.

![output](https://user-images.githubusercontent.com/186349/41732670-ea4df10e-7581-11e8-961d-d3ff5aeb0f6f.gif)

Sorry for the videoconsole quality like GIF, but GitHub doesn't let me add images bigger than 10MB

## Version
The issue showed in `0.5.0-beta.3`
Last known working version was `0.5.0-beta.2`

## Example
https://skgsergio.github.io/cartovl/lightning-lines.without-hack.html

Code: https://github.com/skgsergio/skgsergio.github.io/blob/master/cartovl/lightning-lines.without-hack.html

skgsergio:
I had a look the code and I discovered two fancy functions: `{get,set}SimProgress()`:

```
> console.log(lightnings_layer._viz.filter.getSimProgress() + '\n' + lines_layer._viz.filter.a.getSimProgress());
0.4117499999999316
0.40599806518554815
```

And setting the progress to 0 makes both torques to be in sync again 😃

```
> lightnings_layer._viz.filter.setSimProgress(0); lines_layer._viz.filter.a.setSimProgress(0);

> console.log(lightnings_layer._viz.filter.getSimProgress() + '\n' + lines_layer._viz.filter.a.getSimProgress());
0.030016666666667365
0.030033333333333634

> console.log(lightnings_layer._viz.filter.getSimProgress() + '\n' + lines_layer._viz.filter.a.getSimProgress());
0.8711999999999919
0.8712166666666591
```

IagoLast:
Interesting issue, I guess we should think on a standard way to sync animations.

- At viz object level: `viz0.syncAnimation([viz1, viz2])`
- At layer level: `layer0.syncAnimation([layer1, layer2]);`

maybe this should be the default behaviour

cc @rochoa @elenatorro @davidmanzanares @Jesus89

davidmanzanares:
[...]

PR #636, opened 2018-06-28
Synched animations

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/599


debug/multi-anim.html
```diff
@@ -0,0 +1,119 @@
...
+            center: [-3.589442037237177, 40.614912811223235],
+            zoom: 15,
+            dragRotate: false
+        });
+        carto.setDefaultAuth({
+            user: 'rochoa',
+            apiKey: 'default_public'
+        });
+        const s = carto.expressions;
+        const jaramaSource = new carto.source.Dataset('circuito_del_jarama');
```
jgoizueta: This dataset is very nice for examples, maybe we should ask @rochoa permission to put it in cartovl 😛

debug/multi-anim.html
```diff
@@ -0,0 +1,119 @@
...
+        const racetrackLayer = new carto.Layer('racetrack', jaramaSource, new carto.Viz(`
```
jgoizueta: I mean, this one too, of course

src/core/viz/expressions/Animation.js
```diff
@@ -93,7 +96,46 @@ export class Animation extends BaseExpression {
...
+                    return true;
+                }
+                return false;
+            }));
+            if (waitingForLayer.size > 0) {
+                return;
+            }
+            [...waitingForOthers.values()].map(anim => {
+                if (anim._paused === 'default') {
+                    anim._paused = false;
```
jgoizueta: should be anim.play() right?

src/core/viz/expressions/Animation.js
```diff
@@ -93,7 +96,46 @@ export class Animation extends BaseExpression {
         return !this.paused;
     }
 
+    _dataReady() {
```
jgoizueta: It takes some time to understand this logic 😅 and I'm afraid it'll take some time each time, but it seems right. I think I'd break it into some functions to try to make clear the intent of each piece, but maybe that could end up worse... so do as you please

src/core/viz/expressions/Animation.js
```diff
@@ -93,7 +96,46 @@ export class Animation extends BaseExpression {
...
+                    anim._paused = false;
+                }
+            });
+            waitingForOthers.clear();
+        }
+    }
+
+    _postShaderCompile(program, gl) {
+        waitingForLayer.add(this);
+        this._paused = 'default';
```
jgoizueta: I don't find `default`quite clear... maybe `'uninitialized'`? `'waiting-for-start'`? `'waiting'`?

src/core/viz/expressions/Animation.js
```diff
@@ -93,7 +96,46 @@ export class Animation extends BaseExpression {
...
+        super._postShaderCompile(program, gl);
+    }
+
     _setTimestamp(timestamp) {
+        super._setTimestamp(timestamp);
+
+        if (this._paused && this._lastTime === undefined) {
```
jgoizueta: This should warrant some comment for the future... although I understand anything you write could be potentially more puzzling

src/core/viz/expressions/Animation.js
elenatorro: Which are the values could have `paused`? `true`, `false` and 'default? Maybe, instead, I would use a different property `this._initialized` instead, but I have to take a deeper look.

elenatorro:
After reading [this comment](https://github.com/CartoDB/carto-vl/issues/599#issuecomment-400707023) where we discuss what happens if the user wants to synchronize animations, what if we add a method to do that? So if the user has

`viz1.animation`, `viz2.animation` and `viz3.animation` 

he can:

`viz2.animation.sync([viz2.animation, viz3.animation]);` 

What do you think?

src/core/viz/expressions/Animation.js
```diff
@@ -6,6 +6,9 @@ import Property from './basic/property';
 import Variable from './basic/variable';
 import { castDate } from '../../../api/util';
 
+let waitingForLayer = new Set();
```
IagoLast: const?

debug/multi-anim.html
davidmanzanares: Pinging @rochoa :smile:

src/core/viz/expressions/Animation.js
davidmanzanares: Nop, it is reset on line 105

davidmanzanares:
@elenatorro What would be the semantics of `viz1.animation.sync([viz2.animation, viz3.animation]);`?

At first, I had assumed that `viz1` would be set to the time/progress of the parameter, but then I saw that actually, it must be the other way since you have multiple animations in the array.

For me, it's strange to call a method with the other animations and to actually change the other two, while the first one is left as-is.

I think we could leave this to another PR.

davidmanzanares:
So... approved?

src/renderer/viz/expressions/Animation.js
```diff
@@ -93,7 +96,46 @@ export class Animation extends BaseExpression {
         return !this.paused;
     }
 
+    _dataReady() {
```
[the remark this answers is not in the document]
davidmanzanares: It is not API public, and this is consistent with other methods in the expressions, setting and complying with a new, to be discussed, rule is outside the scope of this

Issue #377, opened 2018-05-08
Viewport agg performance


davidmanzanares:
Viewport aggregation performance is not very good. It could be improved in different ways:
 - Use an acceleration structure like a quadtree or a BVH to speed up the intersection process
 - Use webworkers to avoid blocking the rendering thread.

davidmanzanares:
I think this should be de-prioritized since we prioritized it based on the wrong assumption that this was related to https://github.com/CartoDB/carto-vl/issues/526 But, actually, this is another thing.

davidmanzanares:
By doing some profiling with a point dataset I came with some hot spots:
 - The call to `_featureFromDataFrame`
 - The call to `getAspect()` (quite surprising, but easy to cache)
 - The call to `eval()`

At least for points, the intersection check is not the bottleneck.

I'll try some ideas:
 - Store the features as regular arrays of decoded features in auxiliar arrays, this will tradeoff memory for performance.
- Cache the aspect ratio
- Improve `linear.eval()` for the case of dates (seems difficult)

PR #644, opened 2018-07-02
Improve viewport agg performance

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/377

Improve performance on viewport aggregations with points in 1.9x

Optimizations:
- Remove the call to `_featureFromDataFrame` by caching its results. Tradeoff memory for CPU time.
- Remove the call to `getAspect()` in the _isFeatureInViewport since it is a expensive call in a loop
- `Little Things`

Previously on CARTO VL:
vperf-prev

On the next episode of CARTO VL:
vperf-fix


src/core/renderer.js
```diff
@@ -165,11 +165,35 @@ class Renderer {
 
         viewportExpressions.forEach(expr => expr._resetViewportAgg());
 
+        const viewportExpressionsLength = viewportExpressions.length;
+
         // Avoid acumulating the same feature multiple times keeping a set of processed features (same feature can belong to multiple dataframes).
         const processedFeaturesIDs = new Set();
 
+
+        this.aspect= this.gl.canvas.width / this.gl.canvas.height;
```
IagoLast: instead mutating this, I`d create a constant and i´d use this constant as a parameter in `_isFeatureInViewport``

```js
constant ASPECT_RATIO = this.gl.canvas.width / this.gl.canvas.height;

// ...

 if (!this._isFeatureInViewport(dataframe, i, ASPECT_RATIO)) {
// ...

File: src/core/renderer.js [hunk 2]
```diff
@@ -179,24 +203,25 @@ class Renderer {
...
 
-                const feature = this._featureFromDataFrame(dataframe, i);
-
                 // Ignore filtered features
                 if (viz.filter.eval(feature) < FILTERING_THRESHOLD) {
                     continue;
                 }
 
-                viewportExpressions.forEach(viewportExpression => viewportExpression.accumViewportAgg(feature));
+                for (let j = 0; j < viewportExpressionsLength; j++) {
```
IagoLast: what was the x% improvement of using a for loop?

davidmanzanares: I didn't check it with just that change, sorry, anyway, the for loop is quite hot

src/core/renderer.js
```diff
@@ -165,11 +165,35 @@ class Renderer {
...
         const processedFeaturesIDs = new Set();
 
+
+        this.aspect= this.gl.canvas.width / this.gl.canvas.height;
         dataframes.forEach(dataframe => {
+            const propertyNames = Object.keys(dataframe.properties);
+            const propertyNamesLength = propertyNames.length;
+
             for (let i = 0; i < dataframe.numFeatures; i++) {
+                let feature;
```
jgoizueta: Shouldn't all this be encapsulated in a Dataframe method?

IagoLast: In fact this logic were in the `_featureFromDataFrame` method

davidmanzanares:
:+1: Done, pls, re CR

src/renderer/Renderer.js
```diff
@@ -184,17 +188,20 @@ export default class Renderer {
...
+                }
             }
         });
     }
 
     /**
      * Check if the feature at the "index" position of the given dataframe is in the renderer viewport.
+     * NOTE: requires `this.aspect` to be set
      */
     _isFeatureInViewport(dataframe, index) {
```
IagoLast: what about passing the aspect as a parameter?

davidmanzanares: ok...

src/renderer/Renderer.js
```diff
@@ -219,11 +226,20 @@ export default class Renderer {
...
-        const propertyNames = Object.keys(dataframe.properties);
-        const feature = {};
-        for (let i = 0; i < propertyNames.length; i++) {
-            const name = propertyNames[i];
-            feature[name] = dataframe.properties[name][index];
+        let feature;
+        if (!dataframe.cachedFeatures){
+            dataframe.cachedFeatures = [];
+        }
+        if (!dataframe.cachedFeatures[index]) {
```
IagoLast: Little refactor 🌈 
```js
if (dataframe.cachedFeatures[index]) {
    return dataframe.cachedFeatures[index];
}

const feature = {};
const propertyNames = Object.keys(dataframe.properties);
for (let i = 0; i < propertyNames.length; i++) {
    const name = propertyNames[i];
    feature[name] = dataframe.properties[name][index];
}
dataframe.cachedFeatures.push(feature);
return feature;
```

IagoLast:
> That is not correct since the dataframe.cachedFeatures is uninitialized the first time.
 
are you sure?

if the feature is cached --> return feature from cache
otherwise --> generate feature, save it on cache and return it


I mean... the full method would be like:


```js

if (!dataframe.cachedFeatures){
    dataframe.cachedFeatures = [];
}

if (dataframe.cachedFeatures[index]) {
    return dataframe.cachedFeatures[index];
}

const feature = {};
const propertyNames = Object.keys(dataframe.properties);
for (let i = 0; i < propertyNames.length; i++) {
    const name = propertyNames[i];
    feature[name] = dataframe.properties[name][index];
}
dataframe.cachedFeatures.push(feature);
return feature;
```

davidmanzanares:
Ah, that's another code snippet xD

debug/viewportagg-perf.html
```diff
@@ -0,0 +1,73 @@
...
+      apiKey: 'default_public'
+    });
+
+    // Define layer
+    const source = new carto.source.Dataset('taxi_50k');
+    const viz = new carto.Viz(`
+      width: ($fare_amount/15000)*zoom()
+      filter: animation($tpep_pickup_datetime,10,fade(0.2,0.2))
+      color: rgba(255,255,255,0.2)
+      strokeColor: ramp(linear($tpep_pickup_datetime,time('2016-05-01T00:08:39Z'),time('2016-05-06T10:29:11Z')),[#CF1C90,#F8CA00,#9BBD4D,#07A5AB,#B100EB])
```
IagoLast: maybe some space here :trollface:

davidmanzanares: That's a debug example

Issue #469, opened 2018-05-25
Support using variable inside all expressions


Jesus89:
Check **all expressions** to improve dynamic type checking for variables an replace properly the aliased expression in the variable. For example.

The following codes should be equivalent:

```
color: hsv($category, 0.7, 1.)
```

```
@cat: $category
color: hsv(@cat, 0.7, 1.)
```

rochoa:
I just found another instance of this failing.

Use MVT with metadata, like:

```js
const source = new carto.source.MVT('https://example.com/{z}/{x}/{y}.mvt', {
  score: {
    type: 'number',
    min: 0,
    max: 100
  }
});
const viz = new carto.Viz(`
  @score: $score
  strokeWidth: 0.8
  color: blend(ramp(linear(@score), temps), grey, isnan(@score))
`);
```

It will complain with:

```js
globalAggregation.js:173 Uncaught (in promise) TypeError: Cannot read property 'min' of undefined
    at GlobalAggregattion._compile (globalAggregation.js:173)
    at _getChildren.map.child (base.js:75)
    at Array.map (<anonymous>)
    at Linear._compile (base.js:75)
    at Linear._compile (linear.js:78)
    at _getChildren.map.child (base.js:75)
    at Array.map (<anonymous>)
    at Ramp._compile (base.js:75)
    at Ramp._compile (ramp.js:116)
    at _getChildren.map.child (base.js:75)
```

PR #650, opened 2018-07-03
Support variables in all places

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/469


davidmanzanares:
I think so @elenatorro but I haven't tested it, the idea of this PR is that every variable should work everywhere out of the box since a proxy ensures that no special code is needed for treating the special case.

elenatorro:
Yes, I am asking that in order to see if we can solve this issue with this PR and close both issues :)

PR #656, opened 2018-07-04
Add support for variables to viewport Expressions

### LINKED ISSUE #647: Support variables in viewportFeatures
Currently all arguments to `viewportFeatures` can only be properties ($xxx). Variables that reference properties should also be supported.

We could also support cluster aggregations as arguments, as commented [here](https://github.com/CartoDB/carto-vl/pull/625#issuecomment-402045227).

jgoizueta:
Fixes #647


debug/list.html
```diff
@@ -53,10 +53,11 @@
 
     const source = new carto.source.Dataset('walmarts');
     let viz = new carto.Viz(`
-      @list: viewportFeatures($zip, $sttext)
+      @number: $zip
+      @list: viewportFeatures(@number, $sttext)
       @anotheList: viewportFeatures($zip)
-      width: 20
-      filter: $zip > 50000
+      width: 20 // + 0*@number
```
davidmanzanares: Why these comments?

PR #659, opened 2018-07-04
Replace use of instanceof to check expression type

### LINKED ISSUE #657: Use valueOf when checking with instanceof
When checking types with `instanceof`, we should apply `valueOf` to the expression to support variables, like here:

https://github.com/CartoDB/carto-vl/pull/656/files#diff-70dfd5511c05de98bb8c59e31215cf55R66

jgoizueta:
To support variables we need to use a function so that it is proxied to the underlying expression.
Fixes #657

### DIFF
--- a/src/renderer/viz/expressions/base.js
+++ b/src/renderer/viz/expressions/base.js
@@ -207,4 +207,8 @@ export default class Base {
     eval(feature) {
         throw new Error('Unimplemented');
     }
+
+    isA(expressionClass) {
+        return this instanceof expressionClass;
+    }
 }
--- a/src/client/windshaft-filtering.js
+++ b/src/client/windshaft-filtering.js
@@ -54,14 +54,14 @@ class AggregationFiltering {
     }
 
     _and(f) {
-        if (f instanceof And) {
+        if (f.isA(And)) {
             return this._and(f.a).concat(this._and(f.b)).filter(Boolean);
         }
         return [this._or(f)].filter(Boolean);
     }
 
     _or(f) {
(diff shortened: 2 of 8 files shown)


src/renderer/viz/expressions/base.js
```diff
@@ -207,4 +207,8 @@ export default class Base {
     eval(feature) {
         throw new Error('Unimplemented');
     }
+
+    isA(expressionClass) {
```
elenatorro: This PR looks ✨ The only thing, I would name this function `isExpression` instead of `isA`, because `isA` is not very self explanatory 😂 I would be great to add a test as well.

jgoizueta: I'm sorry you don't like my `IsA` 😢  It has a [long tradition](https://en.wikipedia.org/wiki/Is-a) in OO.

Anyway, let's see 🤔 I'm not sure what naming I prefer...
* `expr.isA(Property)`
* `expr.isExpression(Property)`
* `expr.kindOf(Property)`
* `expr.instanceOf(Property)`
* `expr.isExpressionOfType(Property)`

I'm not sure! 😕

elenatorro: If it is a tradition/convention, then is ok to use `isA` :)

davidmanzanares: hahahaha, just saying... one file - multiple classes is also a tradition in CARTO VL :smile_cat:

elenatorro: :trollface:

PR #665, opened 2018-07-06
Maps API requests improvements

rochoa:
Let's use GET requests to be able to cache some of the Maps API instantiations.


rochoa:
Being that guy. So, now we use Airbnb code style?

IagoLast:
No, we don't have style-guide that's why I said `"things that IMO are issues."` and I tried to explain why I think they are issues.

rochoa:
Could we focus on the functional changes instead of the code style? Because your PR doesn't even follow the proposed style from Airbnb guidelines.

IagoLast:
> Because your PR doesn't even follow the proposed style from Airbnb guidelines.

Again, I never said  we follow airbnb guidelines


> Could we focus on the functional changes instead of the code style

Functional changes LGTM

rochoa:
Then, I prefer not to change the style just for this file. The project uses `const` for function declaration quite heavily. Let's review that once we agree on the code style.

About this PR in particular
 - All the functions declared with `const` are used later in the code.
 - All the functions used in the Windshaft class are declared with `function`, I could even use `const` as they are not used as top functions.
 - I could buy the top-to-bottom argument, but then we need to review the whole Windshaft class itself.

IagoLast:
> All the functions declared with const are used later in the code.

This breaks the stepdown rule which IMO makes the code harder to understand

> All the functions used in the Windshaft class are declared with function, I could even use const as they are not used as top functions.

which functions do you mean? the older ones were declared using function and this PR adds functions declared with const.

> I could buy the top-to-bottom argument, but then we need to review the whole Windshaft class itself.

Sure, but I think small continuous improvements are better than big breaking changes



Finally, function names given by this PR are not verbs which again IMO is an issue, I think function/method names should be verbs to improve legibility.

rochoa:
Feel free to merge your PR.

IagoLast:
Now we have ended the style discussion, we can talk about the PR itself.

As I've said LGTM but I have one suggestion: 

As long as I know the only limitation for the request size is [IE](https://support.microsoft.com/en-ie/help/208427/maximum-url-length-is-2-083-characters-in-internet-explorer).  Reading [webgl compatibility](https://caniuse.com/#search=webgl) this only affects IE11 so I think we can improve the logic and only perform the check when the browser is IE11 so we can benefit of caching bigger requests in the rest of browsers.

Issue #652, opened 2018-07-04
Avoid requiring cartodb_id for some operations


jgoizueta:
Windshaft sources always have a `cartodb_id` unique feature identifier, but other sources may not.

Some operations fail without a `cartodb_id`, for example all viewport aggregations (& feature list) which use the cartodb_id to avoid duplicated features (same feature could appear in several dataframes):

https://github.com/CartoDB/renderer-prototype/blob/025f9d267234609f138755391205d4929eabe08d/src/renderer/Renderer.js#L179-L180

Other places where are using it:
* https://github.com/CartoDB/renderer-prototype/blob/025f9d267234609f138755391205d4929eabe08d/src/interactivity/featureVizProperty.js#L35
* https://github.com/CartoDB/renderer-prototype/blob/025f9d267234609f138755391205d4929eabe08d/src/interactivity/featureVizProperty.js#L57-L58

Interestingly, I see GeoJSON is already generating `cartodb_id` automatically, so the only problem now is with MVT sources.

jgoizueta:
This problem is making two features fail with MVT sources:
* viewport aggregations
* interactivity blends & resets

jgoizueta:
I propose:

Allow users to define a PK property in the MVT source metadata. If passed, it will be used for uniqueness checks, etc (we could even renamed/clone it as `cartodb_id`)

If that's is no specified, I wouldn't bother to try to identify duplicated features among tiles (that would be costly and potentially ambiguous). We could either avoid checks when an id is not present or devise a schema to generate unique ids without collisions between tiles.

jgoizueta:
We have notice the code to avoid dup features was innefective, so we have removed it in
that way, MVTs can be used with viewport functions while we fix this properly. So right now, some features can be duplicated in the viewport aggregations and the viewportFeatures list, which is a bug we already had. 
with reference to #590

davidmanzanares:
Fixed by https://github.com/CartoDB/carto-vl/pull/669

PR #669, opened 2018-07-09
IDs for MVT

jgoizueta:
See #652 

The intent of this is to
* Restore the unique feature filtering for viewport aggregations and features, temporarily removed in https://github.com/CartoDB/carto-vl/commit/3b8f2f905e830547d1f2ed89421408c0d5c5dc1a
* Add cartodb_id to MVT so we have always source-relative feature ids (we already have internal dataframe-relative feature ids).

In the WIP I have added a  `idProperty` to the metadata. It can be used in MVT sources to specify an existing property as the source id. It defaults to `cartodb_id`. The idea is that during MVT decoding, if the property specified in the metadata is not present in the MVT it will be generated automatically in the resulting dataframe (assuming features are not repeated among tiles).

The decisions I had taken:
* Allowing the ids to have arbitrary names, as defined in the metadata. The alternative considered was to fix it to `cartodb_id`.
  The reason for the choice was that it seems more natural for users to use their existing PKs, and to avoid renaming properties.
  On the othe hand, the resulting code for Feature/FeatureVizProperty is now cluttered with `idProperty` parameters, which is now nice.
* To generate the unique-by-source automatic ids I had planned to just use a sequence counter in the source. (shouldn't be a problem, since we have a single JS thread).


davidmanzanares:
I made some small changes:
- Avoid removing the ID property from `properties`. This is expected for the users.
- Fix unit tests
- Remove the autoincremental ID TODO since, as we talked off-channel, it will be highly problematic in combination with per-feature styling
- Support strings as IDs
- Removing one minor constraint from the undocumented Metadata

davidmanzanares:
The error when the id defaults to `cartodb_id` and it is not present it's very unfriendly, do you think we should enforce the presence of an ID in the metadata and then check that every feature has the ID?

elenatorro:
I am thinking about it. What if we use `idProperty='cartodb_id'` as default in the constructor?

davidmanzanares:
Yes, that's the default: https://github.com/CartoDB/carto-vl/pull/669/files#diff-33639dc88d296eb8e2c693c0785b6713R14

I'll add the check for presence in the decoding (if the ID (default or not) is not present for a feature, then throw)

Issue #676, opened 2018-07-10
viewPortFeatures() return empty list after applying filters


Josmorsot:
I'm using **viewPortFeatures** to count the number of elements in the ViewPort. Everything works fine but when I apply any filter and update the bounding box, this function returns an empty list.

I attached an example:
![filters](https://user-images.githubusercontent.com/7818205/42496744-50f646be-8427-11e8-9c52-160abd0a85a5.gif)

davidmanzanares:
Hi @Josmorsot !

Thank you for the report, I'll look into it.

davidmanzanares:
I found a bug regarding blending to variables while trying to debug this. I didn't found other problems, though.

That problem should be fixed in the `fix-var-blending` branch, could you check if that fixes it?

If not, we would need more details or code to help.

davidmanzanares:
Hi, I think we fixed this with https://github.com/CartoDB/carto-vl/pull/688 but it still requires more testing and work.

davidmanzanares:
Hi @Josmorsot , this should be fixed on `master` already.

PR #678, opened 2018-07-10
Bugfix for blending to variables multiple times

davidmanzanares:
Steps to reproduce:
1. Open debug/list.html with these changes
2. Run viz.filter.blendTo(carto.expressions.var('c')) on the console after map loading
3. Run viz.filter.blendTo(carto.expressions.TRUE)
4. Run viz.filter.blendTo(carto.expressions.var('c')) again, it will crash without the changes on src/

- [x] Fix
- [x] Regression test

~Could be fixing https://github.com/CartoDB/carto-vl/issues/676~ No, this doesn't fix that issue.

### DIFF
--- a/src/renderer/viz/expressions/basic/variable.js
+++ b/src/renderer/viz/expressions/basic/variable.js
@@ -56,8 +56,8 @@ export default function variable(name) {
             if (prop == 'parent') {
                 obj[prop] = value;
             } else if (prop == 'notify') {
-                return obj[prop];
-            } else if(alias && alias[prop]) {
+                obj[prop] = value;
+            } else if (alias && alias[prop]) {
                 alias[prop] = value;
             } else {
                 return false;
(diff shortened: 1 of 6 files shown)


src/renderer/viz/expressions/basic/variable.js
```diff
@@ -66,12 +66,16 @@ export default function variable(name) {
...
-            if (prop == '_resolveAliases') {
+            if (prop == 'parent') {
+                return obj[prop];
+            } else if (prop == '_resolveAliases') {
                 return resolve;
             } else if (prop == '_getDependencies') {
                 return _getDependencies;
             } else if (prop == 'notify') {
                 return obj[prop];
+            } else if (prop == 'blendTo') {
```
elenatorro: Question: should we not add the `blendTo` check also in the `set` method?

davidmanzanares: Unless we want to override the method manually by assigning to it, and I don't know why we would want to do that, no, it is not needed

davidmanzanares:
Btw @elenatorro I made some changes after your CR, you may want to look at those too

PR #679, opened 2018-07-10
Overzooming

### LINKED ISSUE #673: Over zooming
tl;dr version: we should be able to use different zoom levels from the source to render different zoom levels in the layer/visualization.

The most basic example being rendering zoom level 15 with data from tiles at zoom level 14. Obviously, the tiles at the source should have enough resolution/detail.

This requires changes in the Source API to be able to say what zoom levels are available. The API could support two mechanisms:
1. Basic: as mentioned before, specifying the max zoom level, from there the visualization will use tiles from that zoom level to render higher zoom levels.
1. Advance: being able to map a source zoom level to a visualisation zoom level. This allows to just use some zoom levels in the source. E.g., your source only has data at zooms 0, 8, and 16, you can map those source zoom levels to visualisation zoom levels: `visZoom => max(visZoom//8 * 8, 16)`.

The _Advance_ mechanism will be useful for implementing https://github.com/CartoDB/carto-vl/issues/17. This requires changes in the backend and the client requesting the tiles with the right resolution, though. This comes from an offline chat I had with @davidmanzanares.

Related issues:
 - https://github.com/CartoDB/carto-vl/issues/17
 - https://github.com/CartoDB/carto-vl/issues/475

davidmanzanares:
Fix #673 

- [x] Advanced API
- [x] Advanced example
- [x] Basic API
- [x] Docs
- [x] Tests

The advanced API allows user to control how to map between viewport zoom levels and MVT URL zoom levels. The default mapping function is `Math.ceil`.

It's easy to fix the zoom level so that all requests are to the same zoom level (regardless of the actual map zoom level):
fixedzoomlevel

It's also easy to make more advanced cases like `only use zoom levels 0,3,6,9...`:
`          viewportZoomToSourceZoom: z => Math.round(z/3)*3`

To be honest, I don't know if we need a basic API anymore, since limitting just the zoom is as easy as:
`          viewportZoomToSourceZoom: z => Math.min(Math.ceil(z), 9)`

I also found that it's easy to get crazy and start requesting thousands of tiles. If you fix the zoom level in the function to a "large" value like 8, and you zoom out, you will be requesting 4^8 (65536) tiles.

cc: @rochoa


rochoa:
VAMOS! 💪

rochoa:
The basic API would be nice to have something like: `maxSourceZoom: 14`.

src/sources/Mvt.js
```diff
@@ -24,14 +24,41 @@ const MVT_TO_CARTO_TYPES = {
     3: geometryTypes.POLYGON
 };
 
+/**
+ * A MVTOptions object declares a MVT configuration
+ * @typedef {object} MVTOptions
+ * @property {string} layerID - layerID on the MVT tiles to decode, the parameter is optional if the MVT tiles only contains one layer
```
elenatorro: `the MVT tiles only contain one layer`

davidmanzanares: :+1:

src/client/rsys.js
```diff
@@ -76,13 +76,8 @@ function tileRsys(x, y, z) {
...
 
-/**
- * Minimum zoom level for which tiles are no larger than the RSys rectangle
- * @param {RSys} rsys
- * @return {number}
- */
-function rZoom(zoom) {
-    return Math.ceil(Math.log2(1. / zoom));
+function rZoom(zoom, viewportZoomToSourceZoom) {
+    return viewportZoomToSourceZoom(Math.log2(1. / zoom));
```
IagoLast: not sure but I think in javascript 1 and 1. is the same right?

src/sources/Mvt.js
```diff
@@ -24,14 +24,41 @@ const MVT_TO_CARTO_TYPES = {
...
+ * @property {function} viewportZoomToSourceZoom - function to transform the viewport zoom into a zoom value to replace `{z}` in the MVT URL template, undefined defaults to `Math.ceil`
```
IagoLast: @property {function} [viewportZoomToSourceZoom=Math.ceil]  [jsDoc default value](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

src/sources/Mvt.js
```diff
@@ -24,14 +24,41 @@ const MVT_TO_CARTO_TYPES = {
...
+
 export default class MVT extends Base {
 
     /**
      * Create a carto.source.MVT.
      *
      * @param {object} data - A MVT data object
      * @param {object} [metadata] - A carto.source.mvt.Metadata object
-     * @param {string} [layerId] - layerID on the MVT tiles to decode, the parameter is optional if the MVT tiles only contains one layer
+     * @param {MVTOptions} [options] - MVT source configuration, the default value will be valid for regular URL templates if the tiles are composed of only one layer
```
IagoLast: document the options pick one of the following options:

- create and define `MVTOptions`
- define options as an object and document every property

src/sources/Mvt.js
```diff
@@ -68,7 +94,11 @@ export default class MVT extends Base {
     }
 
     requestData(viewport) {
-        return this._tileClient.requestData(viewport, this.responseToDataframeTransformer.bind(this));
+        return this._tileClient.requestData(viewport, this.responseToDataframeTransformer.bind(this),
+            zoom => this._options.maxZoom == undefined ?
```
IagoLast: I'd create a function in the MVT object like `responseToDataframeTransformer`

src/sources/Mvt.js
davidmanzanares: Hey, that was already done

src/sources/Mvt.js
davidmanzanares: I don't find that clearer

src/client/rsys.js
davidmanzanares: Yes, it is the same, I changed it

src/sources/Mvt.js
davidmanzanares: changed

PR #688, opened 2018-07-11
Fix category filter bug by caching decoded categories

davidmanzanares:


src/renderer/viz/expressions/ViewportFeatures.js
```diff
@@ -78,6 +78,7 @@ export default class ViewportFeatures extends BaseExpression {
 
     accumViewportAgg(feature) {
         const properties = this._getMinimumNeededSchema().columns;
+        console.log('!!!', { feature, properties });
```
davidmanzanares: hey!

elenatorro: 🤦‍♂️

src/renderer/viz/expressions/ViewportFeatures.js
```diff
@@ -62,7 +62,7 @@ export default class ViewportFeatures extends BaseExpression {
         return this.expr;
     }
 
-    resetViewportAgg() {
+    _resetViewportAgg() {
```
IagoLast: _resetViewportAgg is called from the outside so I think it should be public

davidmanzanares:
Hey, as we discussed off-channel there are some important issues with this PR. Since I think we should push the fix as soon as possible I'm gonna create a new PR with the initial commit and a single regression test.

If you think that some of the changes here can be re-used, feel free to create a new PR, but please, take the mentioned issues (API breaking changes, performance regression) into account.

PR #689, opened 2018-07-11
viewportZoomToSourceZoom should always be defined

rochoa:


src/sources/Mvt.js
```diff
@@ -79,6 +79,7 @@ export default class MVT extends Base {
         this._metadata = metadata;
         this._tileClient = new TileClient(templateURL);
         this._options = options;
+        this._options.viewportZoomToSourceZoom = this._options.viewportZoomToSourceZoom || Math.ceil;
```
IagoLast: eing? I think the constructor already defines a default options values right?  

```js
constructor(templateURL, metadata = new Metadata(), options = { layerId: undefined, viewportZoomToSourceZoom: Math.ceil, maxZoom: undefined }) 
```

rochoa: I know but look at the tests. I actually did two commits to show the incorrect behaviour before fixing it.

If an `options` object is provided, it ignores the default constructor parameter, thus, missing the not-provided keys from it. As the `viewportZoomToSourceZoom` is not optional, we need to guarantee its existence.

Issue #672, opened 2018-07-09
Bundle size 


rochoa:
Review what are the components increasing the size of the bundled code and decide what is the best approach for improving the situation.

```
$ yarn build
...
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets:
  carto-vl.min.js (253 KiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (253 KiB)
      carto-vl.min.js


WARNING in webpack performance recommendations:
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/
```

PR #693, opened 2018-07-12
Optimize bundle size

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/672

* Add Uglify plugin for Webpack 4.0.
* Move SVG to external files and optimize them.


webpack/webpack.min.config.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path');
 const webpack = require('webpack');
 const banner = require('./banner');
+const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
```
IagoLast: since we are using webpack 4 I think this is included by default in the [mode](https://webpack.js.org/concepts/mode/) flag

elenatorro: Yes, you are right. However, by using this plugin we include more options. With this plugin we go from 244k to 241k, which is not a lot so I think we can discuss whether to use it or not.

webpack/webpack.min.config.js
```diff
@@ -13,7 +14,43 @@ module.exports = {
...
+                    output: {
+                        comments: false
+                    },
+                    compress: {
+                        unsafe_comps: true,
+                        properties: true,
+                        keep_fargs: false,
+                        pure_getters: true,
+                        collapse_vars: true,
+                        unsafe: true,
```
davidmanzanares: `unsafe` doesn't feel very safe to me :slightly_smiling_face:

elenatorro: Actually, by setting `unsafe:true` is not necessary to use `unsafe_comps:true`, it sets al `unsafe_` properties to `true`. We can discuss which ones we can set to true, or directly remove it (since it is false base default).

Here is the list https://github.com/mishoo/UglifyJS2/tree/harmony#compress-options

davidmanzanares: I would remove all `unsafe` properties unless there is clear understanding of their behavior and if they bring an important measured improvement in size.

elenatorro: I totally agree 👍 By removing them it goes from 241kb to 242kb, which for me is not very significant.

davidmanzanares:
Isn't this ready to be merged? @elenatorro

Issue #677, opened 2018-07-10
Interactivity not working fine with Line's Layers


Josmorsot:
When I have any layer of lines, the map becomes fully clickable. It's looks like the action area is not calculated correctly in some zoom levels. 

I have pushed an example to: https://github.com/Josmorsot/featureClickCartoVL (Carto-VL has been compiled from master)

GIF: 
![popup](https://user-images.githubusercontent.com/7818205/42501412-fe8f2f64-8433-11e8-8430-f9a26b19adcd.gif)

davidmanzanares:
I think this will be fixed by https://github.com/CartoDB/carto-vl/pull/705

davidmanzanares:
The problem was that some of the points of the line segments were repeated and we didn't manage well that case.

davidmanzanares:
Fixed by https://github.com/CartoDB/carto-vl/pull/705

PR #705, opened 2018-07-13
Fix Line interactivity in corner case in MVT

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/677

### DIFF
--- a/src/client/mvt/feature-decoder.js
+++ b/src/client/mvt/feature-decoder.js
@@ -172,7 +172,7 @@ export function clipPolygon(preClippedVertices, polygon, isHole) {
     const MIN_VALID_NUM_VERTICES = 3;
 
     // preClippedVertices is closed by repeating the first vertex
-    if (preClippedVertices.length >= MIN_VALID_NUM_VERTICES+1) {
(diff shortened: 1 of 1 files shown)


src/client/mvt/feature-decoder.js
```diff
@@ -266,9 +266,27 @@ function clipLine(line) {
...
+
+    clippedLine = removeDuplicatedVerticesOnLine(clippedLine);
     if (clippedLine.length > 0) {
         clippedLines.push(clippedLine);
     }
 
     return clippedLines;
 }
+
+function removeDuplicatedVerticesOnLine(line) {
```
elenatorro: Minor thing: private functions start with `_` 😬The rest looks ✨ 👍

davidmanzanares: Since when? :slightly_smiling_face:

elenatorro: We have discussed this before many times. Private functions (both a method in a class and a function) starts with `_`. If it is going to be used outside, then it does not.

davidmanzanares: Ok...

Issue #709, opened 2018-07-16
Dynamically adding multiple layers in sequence causes rendering issues


det-jscott116:
Hi There,

I am building an application in which I would like to enable users to dynamically add/remove multiple layers. There seems to be an rendering issue when attempting to add a second layer to the map. For some reason this issue seems to be potentially related to type of geometry which is being added to the map.

I have created a JSFiddle accessible via https://jsfiddle.net/na5u8owq/40/ to demonstrate this issue. To reproduce do the following:

1. Click the "Add Layer 1" button to add the first layer - the pink points should be added to the map.
2. Click the "Add Layer 2" button to add the second layer - the purple polygons should be added to the map, everything works as expected.
3. Refresh / rerun the JSFiddle.
4. Now instead add the second layer first - "Add Layer 2" - the purple polygons should be added to the map.
5. Add the first layer - "Add Layer 1" - the red points should be added and visible on the map. However the old layer is now hidden and the entire basemap is now black. The only way to fix this is to manually resize the window. `map.resize()` or `window.dispatchEvent(new Event('resize'));` does not seem to work.

![image](https://user-images.githubusercontent.com/40873630/42742908-84e073d4-8902-11e8-9999-9f61c6494017.png)
The above image demonstrates the issue that I am facing.

Tested in both Chrome v67 and Firefox 61.0.1. 

Console log shows the following errors (Chrome):

```
[.Offscreen-For-WebGL-00000000058F60B0]GL ERROR :GL_INVALID_FRAMEBUFFER_OPERATION : glClear: framebuffer incomplete
[.Offscreen-For-WebGL-00000000058F60B0]RENDER WARNING: texture bound to texture unit 0 is not renderable. It maybe non-power-of-2 and have incompatible texture filtering.
```

Issue #710, opened 2018-07-16
Carto-VL does not work as expected when imported as a NPM package


det-jscott116:
I have run into a strange issue when using the Carto VL library as an NPM package. When trying to add multiple layers to a mapbox map, I have the issue that if there is a delay in adding multiple layers, the subsequent layers are never fetched (by inspecting the network tab) or rendered. The ``layer.on('loaded')`` callback also never fires for any of these subsequent layers.

I have created a quick and dirty CodeSandbox to demonstrate this issue accessible at https://codesandbox.io/s/rj2om7nw14.

**Scenario 1**
1. Click the "Add Layer 1" button. The first layer will be added and appear as pink dots on the map. An alert window will show to confirm that the ``on('loaded')`` callback is fired.
2. Click the "Add Layer 2" button. The second layer will NOT load. The ``on('loaded')`` callback does not fire. The network tab shows that an OPTIONS request is made to Carto, followed by a POST request to get the number of features. However it can be seen that none of the initial vector tiles are being fetched (which appears to be the case when a layer is added). 

**Scenario 2**
1. Click the "Add Layer 2" button. This layer will be added and appear as purple polygons on the map. As before, an alert window will open to indicate the ``on('loaded')`` callback has been fired.
2. Click the "Add Layer 1" button. As with scenario 2, this layer will not be added to the map and nor will the on load callback fire.

**Scenario 3**
1. Click the "Add Both Layers" button. Both layers should be added to the map. Two alerts should appear in sequence to confirm both of the ``on('loaded')`` callbacks have been fired.

So to me this seems like race condition/timing issue - perhaps a promise is failing to resolve, not sure.

If I swap out using the NPM package with using the pre-built library available via the CDN (e.g. https://libs.cartocdn.com/carto-vl/v0.5.1/carto-vl.min.js) this issue does not occur. Also pointing out I have double checked the version I am using via NPM and its the same version 0.5.1.

I'm not sure what could be different between the two, but could this please be investigated.

Thanks.

elenatorro:
Hi @det-jscott116, this should be fixed now in master branch :) I am closing the issue, if there is any other problem please open a new one. Thank you very much.

det-jscott116:
@elenatorro Awesome work! So impressed with the super fast turn around 👍

PR #714, opened 2018-07-16
Add initialized flag to Layer class

elenatorro:
This fixes https://github.com/CartoDB/carto-vl/issues/709 and may be related with https://github.com/CartoDB/carto-vl/issues/710


src/integrator/mapbox-gl.js
```diff
@@ -75,9 +75,9 @@ class MGLIntegrator {
                 this.invalidateWebGLState = invalidate;
                 this.notifyObservers();
                 this.renderer._initGL(gl);
-                this._layers.map(layer => layer.initCallback());
             }
 
+            layer.initCallback();
```
Jesus89: I would rename this to `layer.initialize()`

PR #715, opened 2018-07-16
Add semistandard rules

elenatorro:
This is a proposal to add `semistandard` linter to this repo in order to have similar rules in all our frontend projects. I have left the rules we had in this project (as indent spaces), and we can discuss which ones we want to add / remove.


Jesus89:
I'm OK running lint --fix with the current rules.

davidmanzanares:
:+1: But I would override the indentation to `4` spaces. Every line is going to change without that.

Jesus89:
OK for me. And let's install a linter plugin for eslint in our IDEs (to avoid linter failing in CI)

Jesus89:
UPDATE:
- Tests fixed
- Ignore only `no-tabs` and `import/no-duplicates` rules
- Rename SQL, MVT and GeoJSON since they are acronyms

elenatorro:
Before merging, since there are A LOT of changes, I think it is better to have more eyes on this CR 👀 cc @jgoizueta @IagoLast

IagoLast:
Why leaving the `eslint:recommended` rules ? Can we just extend semistandard?

.eslintrc.json
```diff
@@ -30,6 +34,8 @@
         ],
         "no-console": "off",
         "no-useless-escape": "off",
-        "no-var": "error"
+        "no-var": "error",
+        "no-new": "off",
+        "import/no-duplicates": "off"
```
davidmanzanares: Why this?

Jesus89: This is not required anymore since we have separated SVG files with proper spacing.

elenatorro:
🚀

test/common/util.js
```diff
@@ -39,7 +39,11 @@ function getName (file) {
     );
 }
 
+<<<<<<< HEAD
```
IagoLast: !!!

test/common/util.js
```diff
@@ -51,7 +55,11 @@ function takeReference (file, template, asyncLoad) {
     }
 }
 
+<<<<<<< HEAD
```
IagoLast: !!!

test/common/util.js
elenatorro: There was a problem with `code` but it was fixed in a commit before the merge.

PR #717, opened 2018-07-16
Change npm package

elenatorro:
We want to have a similar approach to `carto.js` way to publish the npm package. The idea is to have all the builds in the main directory when downloading from `npm`, and have `carto-vl.js` as the main module file instead of `src/index.js`.


package.json
```diff
@@ -64,10 +64,14 @@
...
   "main": "src/index.js",
-  "module": "src/index.js",
-  "jsnext:main": "src/index.js",
+  "module": "carto-vl.js",
+  "jsnext:main": "carto-vl.js",
   "scripts": {
+    "prepublish": "yarn build:npm",
     "build": "yarn build:dev && yarn build:min",
+    "build:npm":"yarn build:npm:dev && yarn build:npm:min",
+    "build:npm:dev": "webpack --config webpack/webpack.config.js --output-path='./'",
```
Jesus89: Maybe we should change the following commands:
- `build:npm:dev` to `build:npm:src`
- `build:dev` to `build:src`

elenatorro: I did that in order to keep the way we import the library from our `dist` folder in the examples. Do you think we should remove `dist`?

Jesus89: I was talking only about naming. But, thinking about `dist`... I think I prefer `dist` rather than having the files spread in the main directory. It is a [common practice](https://github.com/vuejs/vue/blob/dev/package.json#L5) also to distribute your `main` from the `dist/your-lib.js` so I think we could follow that path, and also I think we can get rid of `jsnext:main`.

PR #721, opened 2018-07-17
Publish assets to S3

Jesus89:


elenatorro:
I am concerned of what @davidmanzanares says and I think he is right, we can easily break previous maps :S 

Question: which large files are we uploading to git?

Jesus89:
The issue is that S3 SVG files have disappeared for no reason and we don't know why, which files have been erased, etc. So we have no control of the assets.

Since those SVG files will be used in our examples and we won't have tons of SVG files I would keep track of those in the repository.

Regarding the script I can move it out of the release so it can be executed in a different script when required.

davidmanzanares:
If we add GeoJSON (and this PR adds a GeoJSON) in the future we will probably add big GeoJSON files

davidmanzanares:
> The issue is that S3 SVG files have disappeared for no reason

Don't you think is suspicious they disappear after you upload (I don't know how) the GeoJSON to the S3?

How we don't have control of the assets without this, we have access to S3?

Jesus89:
OK, so that seems to be the reason yes. Sorry for that :disappointed:

About having the control I mean keep track of them to know what is uploaded and upload it again if someone (like me) screws up :P

davidmanzanares:
This happened particularly because of a script was miss-configured. Adding more automatization feels like going in the same direction.

Jesus89:
OK, if it is not convincing, I will close the proposal.

I will keep a copy of the assets, btw.

PR #724, opened 2018-07-17
Do not apply "eqeqeq" rule to glsl code

elenatorro:


elenatorro:
When running `yarn lint` it shows there is not an issue with this code, so I think it is fine to leave it like this. Maybe not a unit tests, but an integration test. What do you think?

Jesus89:
Yes. I think a render test should be perfect for that.

Issue #728, opened 2018-07-17
Interactivy stops working when 'prop' filter is applied


Josmorsot:
Hi,

I've just found a bug using filters. Given a layer with an 'Interactivity' assigned, when a filter is applied to this layer using any property, 'interactivity' is never launched again.

I've noted that this happens when using filters like 'in' or 'eq', but not using 'nin' or 'neq'.

I've just uploaded an example using the latest CARTO VL to this [branch](https://github.com/Josmorsot/featureClickCartoVL/tree/ivan/tooltip-not-working-after-filters)

In the example 'interactivity' should work until 'Turn off the some lines' is clicked.

Thanks!

davidmanzanares:
Hi,

Sorry for the delay, feel free to ping me in the future with a "@davidmanzanares" to raise awareness.

This seems like a serious bug, I'll take a look ASAP.

Thank you!

davidmanzanares:
Hello again, 

I was able to reproduce the issue with your provided `carto-vl.js` file, but I couldn't with the current `master` branch.

I think the root cause is the same as the one behind https://github.com/CartoDB/carto-vl/issues/676 and the fix for that already fixed this too.

Could you recheck with `master`?

davidmanzanares:
This should fix the problem: https://github.com/CartoDB/carto-vl/pull/736

Josmorsot:
Thank's @davidmanzanares!, I've just updated to that branch and now it's working fine.

Issue #698, opened 2018-07-13
Dataframe._getFeature() doesn't use the cached feature list generated by Renderer


davidmanzanares:

davidmanzanares:
Fixed by #736

PR #736, opened 2018-07-19
Fix category filtering interactivity

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/728 and https://github.com/CartoDB/carto-vl/issues/698

### DIFF
--- a/src/interactivity/feature.js
+++ b/src/interactivity/feature.js
@@ -21,7 +21,7 @@ export default class Feature {
             variables[varName] = new FeatureVizProperty(`__cartovl_variable_${varName}`, rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty);
         });
 
-        this.id = rawFeature.id;
+        this.id = rawFeature[idProperty];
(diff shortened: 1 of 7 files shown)


src/interactivity/feature.js
```diff
@@ -21,7 +21,7 @@ export default class Feature {
             variables[varName] = new FeatureVizProperty(`__cartovl_variable_${varName}`, rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty);
         });
 
-        this.id = rawFeature.id;
+        this.id = rawFeature[idProperty];
```
elenatorro: Could we name it `propertyName` instead of `idProperty` to be consistent with https://github.com/CartoDB/carto-vl/pull/736/files#diff-3ef34d2cff5a70a4c3874360d727d466R281 and https://github.com/CartoDB/carto-vl/pull/736/files#diff-015175475be99f8a24e412bd4790db19R16?

elenatorro: After talking with @davidmanzanares this is actually the name used as id, so the name is ok 👌

PR #737, opened 2018-07-19
506 spiky lines 1

### LINKED ISSUE #506: Humongous line renders on top of the map
cartovl bug

While zooming, a godzilla line crosses the screen. You can use the branch `line-bug` in [https://github.com/ivanmalagon/hackarto-vl/tree/line-bug](https://github.com/ivanmalagon/hackarto-vl/tree/line-bug) to reproduce it.

The data is private so be careful and nice.

Related: https://github.com/CartoDB/carto-vl/issues/185

jgoizueta:
This fixes most cases of *spiky* lines by adjusting the joint corner when it moves too far away.
Hopely this fixes #506


Jesus89:
I think this is not a proper fix for the issue. It simply does not render the join for a <60º situation. Equivalent to `Math.abs(sin) > 0.866`.

| selection_220 | selection_219 |
|---|---|

elenatorro:
Hi @jgoizueta,  could you add some screenshots for the cases it solves?

We could add a checklist of fixes in this https://github.com/CartoDB/carto-vl/issues/506 (since it is a big one) and maybe it can be useful to start solving them and track the progress.

What do you think? cc @davidmanzanares @Jesus89

Jesus89:
I think it is difficult to split the issue into separated fixes because it requires a change in the render. I would address this by adding render and e2e tests with simple and complete scenarios to check that issues are fixed for real datasets and also nothing is broken.

Regarding this particular PR, I would close it for the reasons I mentioned before.

jgoizueta:
Hey, I agree this is not a *proper* fix, rather a stop gag measure. That's why I made the PR while in RT. The idea is: if this looks generally better, let push this while we work on a better solution.

Jesus89:
Hi @jgoizueta. In case we want a hotfix for not until line rendering is improved I would move forward the previous fix: https://github.com/CartoDB/carto-vl/pull/219, that tackles the same issue from the decoder but adding a `bevel` join instead of rendering no join.

I attach also [this render test](https://github.com/CartoDB/carto-vl/blob/a55b129a2b7c831c363751eef6556784d94edad7/test/integration/render/scenarios/geometries/multi-line-string/joins/reference.png) to check that behaviour.

davidmanzanares:
The #219 PR was blocked since it displayed problematic artifacts on SF Lines, it also contained an important performance regression.

I don't know if this PR shows the same or similar artifacts in that case, and this should be definitely be tested before merging this.

Jesus89:
This PR shows equivalent artefacts when line opacity is less than 1, also in the same case, for angles <60º.

Regarding performance, that PR does not introduce a performance regression, even more, it includes an improvement in the decoder algorithm to reduce the number of operations required in the decoding.

davidmanzanares:
If this PR shows an equivalent regression, then I'm not so sure we should merge this.

We already talked about performance in that PR: https://github.com/CartoDB/carto-vl/pull/219/files/d5db510a31d4598c7370935254cd8c80d4743f40#diff-e589e7c3f424b3bf41e030d60ed0698dR116 The triangle count goes between 1.5x to 2x more.

Also, the #219 had many changes and an additional complexity that I don't think should be added without being a proper fix, this PR is simpler though.

Jesus89:
Adding bevel join requires inevitably adding triangles. Think about how many should be added for round corners. And this does not imply performance since they are new features.

PR #219 also adds necessary tests for line rendering and an improved decoder for lines as I said before, but no more apart from that.

The point is if we want to have a fix for that now or wait until future renderer improvements regarding this issue: https://github.com/CartoDB/carto-vl/issues/738.

davidmanzanares:
> Adding bevel join requires inevitably adding triangles.

Adding triangles on joints that are ok with mitter is not needed.

> how many should be added for round corners

Round corners can be achieved by using the fragment shader which may or may not have a greater performance impact.

> but no more apart from that

The `decodeLine` function is rewritten, which may be ok, but IMO, for a complete fix, not for a workaround.

Jesus89:
We can close this in favor of the final implementation in https://github.com/CartoDB/carto-vl/pull/219

Issue #734, opened 2018-07-19
Separate Dataset / Query field in editor


elenatorro:
## Description

Dataset/Query fields only generate `carto.source.Dataset` sources. This issue was discovered by @jimenamramos when testing the editor.

## Current Result
We currently have a unique field for both Dataset and Query settings.

![captura de pantalla 2018-07-19 a las 9 49 12](https://user-images.githubusercontent.com/3824953/42929014-21aefb8a-8b39-11e8-8aee-82f44778ebf5.png)

However, when generating the HTML file, it always creates the source with `carto.source.Dataset`, even when it is a Query:

![image](https://user-images.githubusercontent.com/3824953/42929112-66aa0982-8b39-11e8-9043-a309aa2dcadc.png)

## Expected Result

There should be two fields: one to create a Dataset source and another one to create an SQL source.

## Steps to reproduce

1. Open the editor
2. Fill Dataset/Query field with a valid query
3. Click on Generate HTML

PR #739, opened 2018-07-19
Save fields on change

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/734 and https://github.com/CartoDB/carto-vl/pull/735

### DIFF
--- a/examples/editor/index.js
+++ b/examples/editor/index.js
@@ -238,32 +238,28 @@ function setConfig (input) {
 }
 
 const superRefresh = (opts) => {
+    const sourceType = document.querySelector('input[name="source"]:checked').value;
+    const sourceAuth = {
+        user: document.getElementById('user').value,
+        apiKey: 'default_public'
+    };
+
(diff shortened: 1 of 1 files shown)


examples/editor/index.js
```diff
@@ -255,6 +263,8 @@ const superRefresh = (opts) => {
     const vizStr = document.getElementById('styleEntry').value;
     const viz = new carto.Viz(vizStr);
 
+    saveConfig();
```
Jesus89: Why is this added?

elenatorro: I added it because currently, every time there is an error, if you change the fields after that those changes are not saved. There may be a better way to do so, actually. Any suggestion?

Jesus89: OK. I would move the function to the top to save after Source or Viz errors. What do you think?

elenatorro: 👍 Agree!

PR #729, opened 2018-07-17
E2E with dockerized Windshaft PoC

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/632

- [x] Dockerized Windshaft environment (postgres, redis, node) (already done by Engine): https://github.com/CartoDB/Windshaft-cartodb/blob/master/docker/Dockerfile-nodejs6-xenial-pg101
- [x] Dockerized Testing envrionment (with prepared redis and postgres DBs) (scripts adapted from https://github.com/CartoDB/Windshaft-cartodb) new Dockerfile here
- [x] CI
- [x] Repeatable testing on local
- [x] Modify prepare tests to push our own datasets

Issue #632, opened 2018-06-27
Improve E2E testing


davidmanzanares:
Current E2E testing is not good:
- Tests are often broken in `master` since they are not run in CI
- They are not run in CI to improve times and stability

Proposal:
- Use a Docker image of Windshaft (the Windshaft team already has one) as a base image
- Create a new docker image by combining that image with some fixed datasets for testing
- Use that image to run a local Windshaft server and use that server as the target of E2E tests
- Enable E2E tests on CI

cc: @jgoizueta @dgaubert @rochoa

davidmanzanares:
Fixed by #744

PR #744, opened 2018-07-19
Enable E2E on CI with dockerized Windshaft

davidmanzanares:
Squashed commit of https://github.com/CartoDB/carto-vl/pull/729

Squashed because there were many small useless commits and because one commit introduced 24 MB of data that later were removed (and I don't want to have 24MB on `master`)

Fix for #632


elenatorro:
After reading what @Jesus89 said we should also include `.docker.run` in `.npmignore` file.

Issue #690, opened 2018-07-12
zoom and scale


davidmanzanares:
After finding some problems when styling by `zoom()`, particularly, when multiplying the width value by `zoom()`, we solve that with https://github.com/CartoDB/carto-vl/pull/653 which basically convert `zoom()` into a new similar function `scale()`.

### Scale

The scale is the ratio between a pixel and a real meter in the Earth. The scale is very useful, since, if you multiply a constant by it in `width:`, you'll get features with constant widths in meters (that means, variable widths in pixels).

`gl.drawingBufferHeight / (this._zoom * 512 * (window.devicePixelRatio || 1))` is one way to define the scale. This code returns `1` when you are viewing the entire Earth vertically (from the North Pole to Antartica) with a map size of 512 pixels.

### Zoom

Zoom, on the other hand, is usually viewed as a logarithmic transformation of the scale, as used by tiling. This is useful for tiling and for end-users, but the property that the scale has (being able to multiply a constant and producing constant (in meters) widths is lost.

## Mapbox GL zoom and CARTO VL zoom

Right now the equation that relates these two "zooms" is `mglZoom = log2(cartoZoom) +1`.

The `+1` is due to the usage of `1024` instead of `512`. Both values are relatively arbitrary (MGL stuck to the common MVT tile size), but since we used a `2x` we get that `+1`.

The `log2` is because our "zoom" expression is actually a scale.

## Proposal (breaking changes ahead)

- Create a new scale expression that returns `gl.drawingBufferHeight / (this._zoom * 512 * (window.devicePixelRatio || 1))`
- Change `zoom()` to return logarithmic zoom levels

Breaking changes:
- All previous usage of `zoom()` will generate completely broken maps. Replacing it by `scale` will fix it generally, but there will be a difference since the new scale would be `2x` larger. In other cases, a correct usage of zoom() will be prefered but it won't be as easy as a simple replacement.

jgoizueta:
Great explanation. I like the idea of having scale and zoom and will make things clearer for users. So I vote for doing this: we still can afford this kind of breaking changes at this stage.

Now, on the minor details, I like the option B idea. What's more if we make the argument to scale default to 1 we support both A and B!

makella:
@davidmanzanares nice work here!!

I agree, option B is more user-friendly.

davidmanzanares:
[...]

Issue #240, opened 2018-04-02
Points width limited to 126px


rochoa:
summary
---

Due a hardware limitation we wont draw points with a total diameter (stroke + radius) higher than 126px.

davidmanzanares:
As a mitigation, I'm gonna increase the limit, since the technical limitation of WebGL is not so strict, see https://webglstats.com/webgl/parameter/ALIASED_POINT_SIZE_RANGE

### Technical implementation to increase the limit from 64 pixels to 255 pixels
Since we are using just 8 bits for the size I'm gonna change the step/interval between sizes, previously it was at a fixed rate of 0.25 pixels, now I'll change it to:

0.25 pixels from [0, 15] covered by [0, 64]
1 pixel from (15, 96] covered by [64, 128]
2 pixels from [96, 255] covered by [128, 208]
[208, 255] region would be uncovered.

Note: 64+32=96

cc: @rochoa 

### Further work

WebGL limits are not going to disappear soon, and saturating the size can be very confusing to the user. I think that some other thing should be done, like changing the color or the form (it doesn't need to be a circle) when the limit is reached.

We can render anything centered at any point within 255x255 pixels in a square. However, it would be technically complicated to animate it or to just throw a JS error to the user.

I think this isn't a priority, but, probably for the Next milestone, we could do something like this, it would require work from design and cartography though.

cc: @makella @arianaescobar

arianaescobar:
Yup, that last part is something that @makella and I can discuss after CARTOLocations if that's ok with you, @rochoa.

davidmanzanares:
I just merged https://github.com/CartoDB/renderer-prototype/pull/251 This should increase the limit to 255, although I'm not confident in that it will work for many Macs with Retina displays. I think it's quite possible that these devices only work fine until 128 pixels.

It will be nice if someone could give me feedback about it.

If that's the case, then we may want to force a maximum of 128 pixels for everyone. :slightly_frowning_face:

makella:
@davidmanzanares on Mac retina, you are right! 255 (left) and 128 (right) look exactly the same:

<img width="1276" alt="screen shot 2018-04-03 at 1 51 11 pm" src="https://user-images.githubusercontent.com/1566273/38272364-3c216404-3746-11e8-8a44-ce676425dd4f.png">

davidmanzanares:
I see, sadly as expected :joy:

[...]

PR #745, opened 2018-07-19
Add scale()

davidmanzanares:
Related with: https://github.com/CartoDB/carto-vl/issues/690

Usage file:///home/dmanzanares/github/renderer-prototype/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IHNjYWxlKDIxKVxuY29sb3I6IHJnYigyMDQsMCwwKSIsImYiOnsibG5nIjo3Ljk4NDI5MjEyODM5MDg3NjUsImxhdCI6LTEuMTM2ODY4Mzc3MjE2MTYwM2UtMTN9LCJnIjoxLjA2MzM5NDk2NzkyMTMyMzUsImgiOiJEYXJrTWF0dGVyIiwiaSI6ImRhdGFzZXQifQ==


makella:
hi @davidmanzanares!

I hope I understand all of this properly!  I totally get `zoom` vs. `scale` and the problems we were encountering. 

What I tested out today are some use cases for `scale()` and I'm not sure that I am using it properly I also think that I am running into the issue of 126px max symbol size in many cases. 

Some examples below:

What if I wanted my symbols to automatically hold "scale" through zoom without multiplying `scale()` by anything? 

For example:
`width: ramp($amount,[1,20])*scale()` 
- doesn't work because I haven't multiplied by anything

On the other hand, if I do
`width: ramp($amount,[1,20])*scale(0.001)` 
- I get scaled symbols, but around zoom 16, all of the symbol sizes are homogeneous because of the scale factor I'm assuming that the size of `20` by that point, is too large for our [126px max symbol size](https://github.com/CartoDB/carto-vl/issues/240)? 
- As a user, how do I know what value to put for my map?
- If you zoom into the largest symbol in the northeast, you'll see what I am talking about:

http://127.0.0.1:8080/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IHJhbXAoJGFtb3VudCxbMSwyMF0pKnNjYWxlKDAuMDAxKVxuY29sb3I6IHJnYigyMDQsMCwwKVxuc3Ryb2tlV2lkdGg6IDAuNSIsImYiOnsibG5nIjoyLjE4ODM3NzIxMjIyNjQ0MTgsImxhdCI6NDEuMzkyMDExNjY5MzI3MjJ9LCJnIjoxMS44MDM3NjkyODExNjQzNzMsImgiOiJEYXJrTWF0dGVyIiwiaSI6ImRhdGFzZXQifQ==

Similarly, I can do this, 
`width: ramp($amount,[0.1,5])*scale(0.0005)`
- but the symbology breaks down at very small zooms 
- and then becomes homogeneous again around z19+ 
- In addition, this is super arbitrary and I don't think a user will go to this level to figure out the exact right values
http://127.0.0.1:8080/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IHJhbXAoJGFtb3VudCxbMC4xLDVdKSpzY2FsZSgwLjAwMDUpXG5jb2xvcjogcmdiKDIwNCwwLDApXG5zdHJva2VXaWR0aDogMCIsImYiOnsibG5nIjoyLjE3MTkyNTk0NDIxNDg3NywibGF0Ijo0MS4zOTY4MjM1NTI5NTI1MjZ9LCJnIjoxMS43NDk2NDY2NDE0NTU5NywiaCI6IkRhcmtNYXR0ZXIiLCJpIjoiZGF0YXNldCJ9

I also tried
`width: ramp($amount,[scale(1),scale(20)])`
http://127.0.0.1:8080/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IHJhbXAoJGFtb3VudCxbc2NhbGUoMSksc2NhbGUoMjApXSlcbmNvbG9yOiByZ2IoMjA0LDAsMClcbnN0cm9rZVdpZHRoOiAwLjUiLCJmIjp7ImxuZyI6Mi4xOTQzNTI4OTQwNDMxNjM0LCJsYXQiOjQxLjM4Mzc1ODgxNTA3NDM3fSwiZyI6MTEuNjAxMzMzNDEwOTE3ODIzLCJoIjoiRGFya01hdHRlciIsImkiOiJkYXRhc2V0In0=

and
`width: blend(scale(0.1),scale(0.5),linear($amount,1,100))`

and
`width: scale(sqrt($amount))` << that totally broke/froze the map

I also tried 
`scale(ramp($amount,[1,20]))` << broke/froze the map

I definitely want to talk this through!

Because really, I'm trying to figure out the logical use of this in my head. What am I trying to achieve in each of the examples? Mostly to have consistent, proportionately scaled symbols through zoom. But is that the right use of `scale()`?

davidmanzanares:
> What if I wanted my symbols to automatically hold "scale" through zoom without multiplying scale() by anything?

Yes, you can! Just use `width: scale (0.1)` for example.

Regarding the 126 size limit, yes, this doesn't override that limit.

> As a user, how do I know what value to put for my map?

`scale(x)` will output a width to make the width at zoom level 0 equal to `x` pixels. If you are at zoom `10` you're viewing things at 1024x, so, an `x` equal to 0.05 will output a width of 51 pixels at zoom level `10`, 102 pixels at zoom 11 and 25 pixels at zoom 9.

We could add a second parameter to `scale`, so that `scale(25, 7)` meant, `25` pixels at zoom level `7`.

> If you zoom into the largest symbol in the northeast, you'll see what I am talking about:

Yes, that's the size limit, but that's another issue.

> but the symbology breaks down at very small zooms

Consider that for that case you are like trying to view a single tree from the space, simply you can't.

> and then becomes homogeneous again around z19+

That's the size limit issue.

> width: ramp($amount,[scale(1),scale(20)])
http://127.0.0.1:8080/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IHJhbXAoJGFtb3VudCxbc2NhbGUoMSksc2NhbGUoMjApXSlcbmNvbG9yOiByZ2IoMjA0LDAsMClcbnN0cm9rZVdpZHRoOiAwLjUiLCJmIjp7ImxuZyI6Mi4xOTQzNTI4OTQwNDMxNjM0LCJsYXQiOjQxLjM4Mzc1ODgxNTA3NDM3fSwiZyI6MTEuNjAxMzMzNDEwOTE3ODIzLCJoIjoiRGFya01hdHRlciIsImkiOiJkYXRhc2V0In0=

That's a bug, good catch!

> scale(sqrt($amount))

Yes, that freezes your map because that will try to make each little point in Barcelona the size of Europe :joy:

makella:
> Yes, you can! Just use width: scale (0.1) for example.
But could something like `scale()` by itself default to that (where `scale()` = scale(0.1)` or whatever default number we decide)? And if you want to override the default, you can put in your own number?

> We could add a second parameter to scale, so that scale(25, 7) meant, 25 pixels at zoom level 7.
Yes! I think that would be great! And would 25 scale up and down between the remaining zooms?

> Yes, that's the size limit, but that's another issue.
This could get tricky/messy for users. Can we implement a hard max so people don't get the homogeneous symbols? where the smallest value width can't go above x and the max above our limit of 126?

I'll try things out again w/ your advice :)

> Yes, that freezes your map because that will try to make each little point in Barcelona the size of Europe 
UMMMM I want all my points the size of Europe!! 🇪🇺 BIG MAPS haha...
But, what if my `$amount` value was small? That should work, right?

makella:
> Consider that for that case you are like trying to view a single tree from the space, simply you can't.

This is going to be a hard balance... I can't seem to find the right one myself. If I want the points visible at small zooms, they are way too big at the intermediate zooms... I wonder if the second parameter `scale(zoom,width)` would fix that.

davidmanzanares:
>  I wonder if the second parameter scale(width,zoom) would fix that.

No, I don't think that's going to fix it. It's just another way to specify the scale input.

The problem you described is expected. If you make things constant in meters, then the range at which they appear at an OK size will be small.

Of course, you can limit the zoom levels and you can also mix `scale` with other expressions.

src/renderer/viz/expressions/scale.js
```diff
@@ -0,0 +1,48 @@
...
+ * const viz = new carto.Viz(`
+ *   width: s.scale(0.1)
+ * `);
+ *
+ * @memberof carto.expressions
+ * @name scale
+ * @function
+ * @api
+ */
+export default class Scale extends BaseExpression {
```
elenatorro: Does it make sense to set a default scale, so if the user writes `* scale()` everything remains the same?

src/renderer/Renderer.js
```diff
@@ -240,7 +240,8 @@ export default class Renderer {
         const gl = this.gl;
         const aspect = this.getAspect();
         const drawMetadata = {
-            zoom: gl.drawingBufferHeight / (this._zoom * 1024 * (window.devicePixelRatio || 1)) // Used by zoom expression
+            zoom: gl.drawingBufferHeight / (this._zoom * 1024 * (window.devicePixelRatio || 1)), // Used by zoom expression
```
elenatorro: I think we could remove this comment.

src/renderer/viz/expressions/scale.js
```diff
@@ -0,0 +1,48 @@
+import BaseExpression from './base';
```
elenatorro: The expression class LGTM ✨. I would add:

1. specs
2. rename the file to `Scale.js`.

elenatorro:
This bug still needs to be solved, right?

`width: ramp($amount,[scale(1),scale(20)])`

src/renderer/viz/expressions/scale.js
davidmanzanares: Another option will be to change the signature of `scale` so that it could be used as: `width: sqrt($amount) * scale( 8)`, meaning `sqrt($amount)` pixels at zoom 8. But this breaks the original approach (implementation-wise I don't care, both are very easy to implement, this feature is clearly about user-friendliness, not about implementaiton). @makella what do you think?

src/renderer/Renderer.js

src/renderer/viz/expressions/scale.js
davidmanzanares: width: scale(sqrt($amount)) * 0.0001 BAD, UGLY
width: scale(sqrt($amount), 7)
width: sqrt($amount) * scale(7)
width: ramp(zoomranges(7,10,12),
[blablabla, blablabla, scale(sqrt($amount), 7)]

davidmanzanares:
I added the discussed second optional parameter, @elenatorro you may want to CR it again

src/renderer/viz/expressions/Scale.js
```diff
@@ -0,0 +1,51 @@
...
+ * @name scale
+ * @function
+ * @api
+ */
+export default class Scale extends BaseExpression {
+    constructor (width, zoomlevel = 0) {
+        width = implicitCast(width);
+        zoomlevel = implicitCast(zoomlevel);
+        super({
+            scale: div(mul(width, 0), pow(2, zoomlevel))
```
elenatorro: Why do we multiply width by 0? (I do not understand this part 😓 )

test/unit/renderer/viz/expressions/Scale.test.js
```diff
@@ -0,0 +1,10 @@
+import { validateStaticType } from './utils';
+
+describe('src/renderer/viz/expressions/zoom', () => {
```
elenatorro: Some tests would be great ✨ and use `src/renderer/viz/expressions/scale` instead of `src/renderer/viz/expressions/zoom`

src/renderer/viz/expressions/Scale.js
```diff
@@ -0,0 +1,51 @@
...
+    eval () {
+        return this.scale.eval();
+    }
+    _compile (metadata) {
+        checkType('scale', 'width', 0, 'number', this.scale.a.a);
+        checkType('scale', 'zoomlevel', 1, 'number', this.scale.b);
+        super._compile(metadata);
+    }
+    _preDraw (program, drawMetadata, gl) {
+        this.scale.a.b.expr = drawMetadata.scale;
```
davidmanzanares: You need to multiply by the scale, the `0` is a placeholder that is set to the proper scale here

test/unit/renderer/viz/expressions/Scale.test.js
davidmanzanares: Yes, I'll fix the typo, regarding tests, this is lacking render tests actually (this cannot be tested with unit tests without coupling to the rendering implementation)

davidmanzanares:
@elenatorro since I based https://github.com/CartoDB/carto-vl/pull/860 in this branch and I don't want to keep working on two branches to fix the issues you pointed out here I will close this PR and address your comments in the linked PR

PR #748, opened 2018-07-20
Fix editor examples adding source option

elenatorro:
## Description

We have added an option to select the source type between Dataset and SQL. In this PR the examples have been updated so now they have configured which source type are using.

## Result

editor


elenatorro:
Awesome @Jesus89, thanks ✨

PR #752, opened 2018-07-23
Lazy property decoding

davidmanzanares:
Experimental


davidmanzanares:
After some real-use cases testing, it seems this actually makes things worse

Issue #594, opened 2018-06-21
Add an easy way to wait for all layers `loaded` event


davidmanzanares:

davidmanzanares:
I think we could add a `carto.on` function that worked like this:
```
carto.on([layer1, layer2], 'loaded', ()=>{

});
```

elenatorro:
I **love** this! 💖

I would say `carto.on('loaded', [...], () => ...`

davidmanzanares:
Fixed by #755

PR #755, opened 2018-07-23
Add `carto.on` function

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/594

### DIFF
--- a/src/apiUtils.js
+++ b/src/apiUtils.js
@@ -1,0 +1,42 @@
+
+/**
+ * Register an event handler for the given event name and for the given list of layers. Valid names are: `loaded`, `updated`.
+ *
+ * The 'loaded' event will be fired when all the layers are loaded (and their 'loaded' events are fired).
+ *
+ * The 'updated' event will be fired whenever one of the layers fired an 'updated event',
+ * but throttled by requestAnimationFrame to return a maximum of one event per frame.
+ *
+ * @param {string} eventName - Supported event
+ * @param {Layer[]} layerList
+ *
+ * @memberof carto
+ * @api
+ */
+export function on (eventName, layerList, callback) {
+    if (eventName === 'loaded') {
+        const waitingGroup = new Set(layerList);
+        layerList.forEach(layer => {
+            layer.on(eventName, () => {
+                waitingGroup.remove(layer);
+                if (waitingGroup.size === 0) {
+                    callback();
+                }
+            });
+        });
+    } else if (eventName === 'updated') {
+        let scheduledRAF = false;
+        layerList.forEach(layer => {
+            layer.on(eventName, () => {
+                if (!scheduledRAF) {
+                    window.requestAnimationFrame(() => {
+                        scheduledRAF = false;
+                        callback();
+                    });
+                }
+            });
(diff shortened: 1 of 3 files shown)


src/apiUtils.js
```diff
@@ -0,0 +1,42 @@
...
+ * The 'updated' event will be fired whenever one of the layers fired an 'updated event',
+ * but throttled by requestAnimationFrame to return a maximum of one event per frame.
+ *
+ * @param {string} eventName - Supported event
+ * @param {Layer[]} layerList
+ *
+ * @memberof carto
+ * @api
+ */
+export function on (eventName, layerList, callback) {
```
elenatorro: The callback is called `n` times, being `n` the layerList length, and it should be called only once.

elenatorro: Should we show an error / warning if the user wants to listen on a layer that is not added to the map?

elenatorro: After talking about this, this is working fine: the 'updated' event is called n times because it is the expected behaviour 👍

elenatorro: Solved: we should not show the error because the layer may be added later.

test/integration/user/test/apiUtils.test.js
```diff
@@ -0,0 +1,81 @@
...
+        });
+    });
+
+    describe('.off', () => {
+        it('should stop firing events', () => {
+            const layerLoaded = jasmine.createSpy('layerLoaded1');
+            layer1.on('loaded', layerLoaded);
+            layer2.on('loaded', layerLoaded);
+            carto.on('loaded', [layer1, layer2], layerLoaded);
+            carto.on('loaded', [layer1, layer2], () => {
```
Jesus89: off?

davidmanzanares: Yes, I forgot about that... the problem I see here is that the test is passed without calling `off`, I'll look into it

src/index.js
```diff
@@ -45,4 +46,4 @@ import { version } from '../package.json';
 
 const source = { Dataset, SQL, GeoJSON, MVT };
 
-export { version, setDefaultAuth, setDefaultConfig, source, expressions, Layer, Viz, Map, Interactivity };
+export { version, on, setDefaultAuth, setDefaultConfig, source, expressions, Layer, Viz, Map, Interactivity };
```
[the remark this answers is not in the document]
davidmanzanares: :+1: Thank you! The test seems broken too :(

davidmanzanares:
I hope to have solved all the issues

src/apiUtils.js
```diff
@@ -0,0 +1,86 @@
...
+        layerList.forEach(layer => {
+            const internalCallback = () => {
+                if (!scheduledRAF) {
+                    window.requestAnimationFrame(() => {
+                        scheduledRAF = false;
+                        callback();
+                    });
+                }
+            };
+            layer.on(eventName, internalCallback);
```
elenatorro: Instead of `layer.on(eventName...` I would say `layer.on('updated'...` to be consisted with the loaded event.

Issue #757, opened 2018-07-23
Add layer hide/show methods


davidmanzanares:
Since https://github.com/CartoDB/carto-vl/issues/428 is difficult to implement and it will imply unneeded performance implications, we have come up with the idea of adding `hide`/`show` methods to `Layer`.

cc: @padawannn

elenatorro:
🚀

PR #758, opened 2018-07-24
Add hide / show layer methods

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/757

## Checklist

* [x] Add show method
* [x] Add hide method
* [x] Disable interactivity when layer is hidden
* [x] Enable interactivity when layer is hidden
* [x] Add debug example
* [x] Add layer integration tests
* [x] Add layer tests
* [x] Add documentation

### DIFF
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -189,9 +189,11 @@ export default class Layer {
         this.metadata = metadata;
 
         source.bindLayer(this._onDataframeAdded.bind(this), this._onDataLoaded.bind(this));
+        
         if (this._source !== source) {
             this._freeSource();
         }
+
         this._source = source;
         this.requestData();
 
@@ -295,6 +297,11 @@ export default class Layer {
         if (!this.metadata) {
             return;
         }
+
+        if (!this._visible) {
+            return;
+        }
+
         this._source.requestData(this._getZoom(), this._getViewport());
         this._fireUpdateOnNextRender = true;
     }
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -189,7 +189,7 @@ export default class Layer {
         this.metadata = metadata;
 
         source.bindLayer(this._onDataframeAdded.bind(this), this._onDataLoaded.bind(this));
-        
+
         if (this._source !== source) {
             this._freeSource();
         }
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -294,11 +294,7 @@ export default class Layer {
     }
 
     async requestData () {
-        if (!this.metadata) {
-            return;
-        }
-
-        if (!this._visible) {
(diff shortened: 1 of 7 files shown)


src/Layer.js
```diff
@@ -281,22 +291,57 @@ export default class Layer {
...
+     */
     hasDataframes () {
         return this._renderLayer.hasDataframes();
     }
 
+    /**
+     *  Get layer id
+     *
+     * @memberof carto.Layer
+     * @instance
```
davidmanzanares: Documenting all of this and exposing it on the API has consequences, some of this methods are probably broken, they certainly are untested, users haven't requested them and some of them can be confusing. For example, `getSource` and `getViz` semantics are not clear, when an `update` is performed the source is not changed immediately. All of this should be documented if we decide to put it on the API. To be honest, I think documenting this is not a good idea at this moment

src/Layer.js
```diff
@@ -305,8 +350,49 @@ export default class Layer {
...
+     * Change layer visibility to visible
+     *
+     * @memberof carto.Layer
+     * @instance
+     * @api
+     *
+     * @fires updated
+     */
+    show () {
+        this.visible = true;
```
davidmanzanares: If `visible` is part of the API, then it should have a setter, if not it should be `_visible`

src/Layer.js
```diff
@@ -305,8 +350,49 @@ export default class Layer {
...
+        this._integrator.changeVisibility(this._id, this.visibility);
+        this._fire('updated', { visibility: this.visibility });
```
davidmanzanares: Do we really need to put the visibility in the event? this change requires to change the documentation and to fire with the same property in all points of the program. I would prefer to keep it simple stupid (KISS)

src/constants/layer.js
```diff
@@ -0,0 +1,4 @@
+export const layerVisibility = Object.freeze({
+    VISIBLE: 'visible',
```
davidmanzanares: In my honest opinion this is less readable than putting `'visible'`

src/integrator/Map.js
```diff
@@ -76,6 +80,37 @@ export default class Map {
...
+                this.show(layerId);
+                break;
+            case layerVisibility.HIDDEN:
+                this.hide(layerId);
+                break;
+        }
+    }
+
+    hide (layerId) {
+        for (let index = 0; index < this._layers.length; index++) {
```
davidmanzanares: This and `show()` can be done with a `Set` with much less complexity

src/constants/layer.js
elenatorro: In this case is not a matter of readability, but of reusability.

elenatorro:
🚀

PR #763, opened 2018-07-24
Modified image examples

makella:
@elenatorro @davidmanzanares 

I modified the Single Image and Multiple Images maps and the wording for the "Raw image" map. 

I changed them so they could be more in line with the [image guide](https://carto.com/developers/carto-vl/guides/introduction-to-image-files/), and also removed some of the other property (color, width) styling complexity. 

For example, the single and multiple image examples were using a URL to access symbols that we have built-in and there was over-complication in the sizing and color. Also, the multiple symbol example was assigning symbols based on a numeric attribute and I think it will be more straightforward to demonstrate on category data.

Please review and let me know if you think these examples are ok.

Thanks!
mamata


examples/styling/raw-image.html
```diff
@@ -2,7 +2,7 @@
 <html>
```
elenatorro: As we renamed other files, I would also change it to `external-image.html` :)

elenatorro: Just realized we already have an `image-external` one, so I would delete this file because it is duplicated, right?

makella: ahhh! sorry! so i named all of the image related ones with the word `image-` first so they were easier to find and organize. 

i don't see `external-image` in my branch locally (weird)

anyway, the three files for these examples are:
screen shot 2018-07-25 at 7 46 53 am

makella:
@elenatorro thank you! how long do these things typically take to make it to the dev center?

elenatorro:
Hi @makella, we have to release a new version before, we will probably release a patch version today :)

Issue #762, opened 2018-07-24
Built-in images plus, square, and squareOutline not working


makella:
Using the map below, you will see that `plus`, `square`, and `squareOutline` are throwing errors.

They are from the list of built-in image files:
https://github.com/CartoDB/carto-vl/issues/535#issuecomment-402489433

https://cartodb.github.io/carto-vl/examples/editor/index.html#eyJhIjoibmVfMTBtX3BvcHVsYXRlZF9wbGFjZXNfc2ltcGxlIiwiYiI6IiIsImMiOiJjYXJ0b3ZsIiwiZCI6Imh0dHBzOi8ve3VzZXJ9LmNhcnRvLmNvbSIsImUiOiJAMTogYmljeWNsZVxuQDI6IGJ1aWxkaW5nXG5AMzogYnVzXG5ANDogY2FyXG5ANTogY2lyY2xlXG5ANjogY2lyY2xlT3V0bGluZVxuQDc6IGNyb3NzXG5AODogZmxhZ1xuQDk6IGhvdXNlXG5AMTA6IG1hcmtlclxuQDExOiBtYXJrZXJPdXRsaW5lXG4vL0AxMjogcGx1c1xuLy9AMTM6IHNxdWFyZVxuLy9AMTQ6IHNxdWFyZU91dGxpbmVcbkAxNTogc3RhclxuQDE2OiBzdGFyT3V0bGluZVxuQDE3OiB0cmlhbmdsZVxuQDE4OiB0cmlhbmdsZU91dGxpbmVcblxuXG53aWR0aDogMTBcbmNvbG9yOiB5ZWxsb3dcbnN5bWJvbDogQDE4IiwiZiI6eyJsbmciOjAsImxhdCI6LTEuMTM2ODY4Mzc3MjE2MTYwM2UtMTN9LCJnIjowLjQ5Nzg1MTcyMzU4MzkzMTg3LCJoIjoiRGFya01hdHRlciIsImkiOiJkYXRhc2V0In0=

elenatorro:
Good catch, @makella!!!! Fixed in https://github.com/CartoDB/carto-vl/pull/764 ✨

makella:
great!! thank you!!

PR #764, opened 2018-07-25
Add missing built in images

elenatorro:
Fixes https://github.com/CartoDB/carto-vl/issues/762


elenatorro:
Good question @Jesus89, I looked to similar lower functions, as `align_center` and it uses underscore... We may allow also this in built-in images. What do you think @makella @jgoizueta @davidmanzanares?

davidmanzanares:
We decided to use camelCase for String API constants, however, since the JS API is in UPPERCASE, I considered that an underscore was needed there.

PR #765, opened 2018-07-25
Another twist on the lines fix

jgoizueta:
This adjusts the condition for using miter (which wasn't used for angles close to 180), and eliminates a couple of triangles in the miter case (as was done before). It also supresses some overlapping in bevel joins, using a larger triangle. This latter triangle can introduce spikes again, and to alleviate that problem the length of the join normals is limited to the length of the connecting segments.

So, this should be almost as efficient as the original code (only an additional triangle for bevels), but it's not perfect and the problems when line width is larger than the lengths are still there.

There's also another problem: I've added a myriad of variables that make the code confusing. (We should give a try at refactoring)

### DIFF
--- a/src/renderer/decoder.js
+++ b/src/renderer/decoder.js
@@ -191,21 +191,7 @@ function addLine (lineString, vertices, normals) {
                     }
                 } else {
                     // Bevel join: adjust vertices and produce bevel triangle
-
-                    // Spike avoidance: joinNormal can produce points farther away than the geometry segments
-                    // This if fundamentally flawed: joinNormal dimensions are based on unit vectors, with no
-                    // relation to the segment geometry.
-                    // This could be performed properly in the vertex shader, if we pass segment information
-                    // (prev and/or next vertices) with each vertex.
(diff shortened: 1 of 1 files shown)


Jesus89:
Hi @jgoizueta. This seems an implementation of the [First approach](https://github.com/CartoDB/carto-vl/pull/219). This approach fixes spiky lines but still have the issue of the lines collapsing when the expanded lines in the shader collide.

Second approach does not have that issue with the expanded lines, but includes more triangles. I'll try to merge both approaches to optimize the number of triangles. We could use this mixed approach if testing for opacity unification per feature works with a good performance (cc @davidmanzanares).

Btw, I have made a lines animation with the lines-fix-branches to collect all the information:

`line-tests`



`flat-corners`



`506-spiky-lines-1`



`506-fix-line-rendering` (test)



`line-solution-shader`

Jesus89:
We can close this in favor of the final implementation in https://github.com/CartoDB/carto-vl/pull/219

PR #766, opened 2018-07-25
Workaround core-js performance bug

davidmanzanares:
Core-js bug https://github.com/zloirock/core-js/issues/377 is making 10x slower CARTO VL on Angular applications.

Since Core-js is so common (https://www.npmjs.com/package/core-js https://github.com/zloirock/core-js/network/dependents) I think it's better to live with this workaround.


src/utils/util.js
```diff
@@ -32,6 +32,15 @@ export function isObject (value) {
     return value !== null && (type === 'object' || type === 'function');
 }
 
+export function getFloat32ArrayFromArray (array) {
+    const length = array.length;
```
IagoLast: Can we just? 

```js
getFloat32ArrayFromArray (array) {
  return new Float32Array(array);
}
```

IagoLast: or in any case use [from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/from) 

```js
getFloat32ArrayFromArray (array) {
  return new Float32Array.from(array)
```

davidmanzanares: The first one, of course not, you are doing the same and you are hitting the same monkey patch, getting the same results.

davidmanzanares: The second one though makes a lot of sense, and I actually tried it before, but no luck, the performance is bad too, core-js is probably monkey patching that too.

IagoLast: 👍 good to know :)

PR #773, opened 2018-07-26
Improve GeoJSON source

Jesus89:
- Fix empty `properties` in GeoJSON features (+tests)
- Allow CW GeoJSON polygons (+tests)


src/sources/GeoJSON.js
```diff
@@ -386,18 +391,22 @@ export default class GeoJSON extends Base {
...
             }
-            if (!this._isClockWise(data[i])) {
-                if (i > 0) {
-                    holeIndex += data[i - 1].length;
-                    polygon.holes.push(holeIndex);
-                } else {
-                    throw new CartoValidationError('source', 'firstPolygonExternal');
-                }
+
+            let l = data[i].length;
```
elenatorro: `l` => `dataLength` and use `const` instead of `let`

src/sources/GeoJSON.js
```diff
@@ -423,7 +432,7 @@ export default class GeoJSON extends Base {
             total += (pt2[1] - pt1[1]) * (pt2[0] + pt1[0]);
             pt1 = pt2;
         }
-        return total >= 0;
+        return total >= 0; // CW
```
elenatorro: Instead of a comment, create a private function:

```js
function _isClockWise (value) {
    return value >= 0;
}
```

src/sources/GeoJSON.js
```diff
@@ -56,6 +56,11 @@ export default class GeoJSON extends Base {
         } else {
             throw new CartoValidationError('source', 'nonValidGeoJSONData');
         }
+
+        // Initialize properties attr
```
elenatorro: Remove the comment and create a private function

```js
this._features = _initializeFeatureProperties(this._features);
```

==> 

```js
function _initializeFeatureProperties (features) {
    ...
  
   return features;
}
```

src/sources/GeoJSON.js
```diff
@@ -386,18 +391,22 @@ export default class GeoJSON extends Base {
...
         };
         let holeIndex = 0;
+        let firstReverse = false;
         for (let i = 0; i < data.length; i++) {
-            for (let j = 0; j < data[i].length; j++) {
-                const point = this._computePointGeometry(data[i][j]);
-                polygon.flat.push(point.x, point.y);
+            let reverse = this._isReversed(data[i]);
+            if (i === 0) {
+                firstReverse = reverse;
```
elenatorro: We can add here a pre condition, to set the first reverse before the loop and start with `i = 1` to avoid this check in every loop.

src/sources/GeoJSON.js
Jesus89: Sure

src/sources/GeoJSON.js
Jesus89: I don't think this is needed, because this function does not check ClockWise, only if the value is positive because there is no context. Also, it adds an extra function call for each point and it could add a bit of extra computing time.

I could improve the comment instead.

Issue #754, opened 2018-07-23
Ramp doesn't show changes in the output values


davidmanzanares:
## Description

When an animated expression is passed to a ramp as the 'palette' parameter, the animation does not work.

**Example:**
```js 
color: ramp(linear(log($numfloors), 2, 4), [red, rgb(0,0,255 * sin(now()))])
```

In the code above, we see we have the `now()` expression, but the visualization color does not change.

**Editor example:**

`examples/editor/index.html#eyJhIjoibW5tYXBwbHV0byIsImIiOiIiLCJjIjoiZG1hbnphbmFyZXMiLCJkIjoiaHR0cHM6Ly97dXNlcn0uY2FydG8uY29tIiwiZSI6ImNvbG9yOiByYW1wKGxpbmVhcihsb2coJG51bWZsb29ycyksIDIsIDQpLCBbcmVkLCByZ2IoMCwwLDI1NSpzaW4obm93KCkpKV0pXG5zdHJva2VDb2xvcjogb3BhY2l0eSh3aGl0ZSwgMC41KVxuZmlsdGVyOiBiZXR3ZWVuKCRudW1mbG9vcnMsIDEwLCAxMjApIiwiZiI6eyJsbmciOi03My45NzI3MTU0Mzg0Mzk2NSwibGF0Ijo0MC43NDk4NjY2NTY3MjAyODZ9LCJnIjoxMi4zMzkzNDE0ODA2MzgyLCJoIjoiRGFya01hdHRlciIsImkiOiJkYXRhc2V0In0=`

PR #777, opened 2018-07-27
Fix ramp animated values

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/754

Fixed:

```js
color: ramp(linear(log($numfloors), 2, 4), [red, rgb(0,0,255 * sin(now()))])
```

ramp_animated

**Progress**

* [x] Fix children chain for ramps
* [x] Fix animation working properly in children chain
* [x] Add specs


src/renderer/viz/expressions/ramp.js
```diff
@@ -287,28 +284,31 @@ export default class Ramp extends BaseExpression {
     _computeGLTextureIfNeeded (gl) {
         const texturePixels = this._computeTextureIfNeeded();
 
-        if (this._GLtexCategories !== this.input.numCategories) {
+        if (this._GLtexCategories !== this.input.numCategories || this.isAnimated()) {
```
davidmanzanares: Uploading the texture is performance intensive and I think this is only needed (regarding animation) when the palette is animated, not when the rest of the ramp parameters are animated.

src/renderer/viz/expressions/ramp.js
```diff
@@ -287,28 +284,31 @@ export default class Ramp extends BaseExpression {
...
+            this._bindGLTexture(gl, texturePixels);
         }
     }
 
+    _bindGLTexture (gl, texturePixels) {
+        gl.bindTexture(gl.TEXTURE_2D, this.texture);
+
+        if (this.type === rampTypes.COLOR) {
+            gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
+            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, COLOR_ARRAY_LENGTH, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, texturePixels);
```
davidmanzanares: This comment may be a premature optimization, but `gl.texImage2D` allocates GPU memory, there is a `gl.texSubImage2D` method that is meant to be used for already allocated texture (see https://www.khronos.org/opengl/wiki/GLAPI/glTexSubImage2D ) The workflow would be to first call `texImage2D` and then, on texture updates, use the `sub` function.

However, if you didn't see a performance problem here, feel free to ignore this.

src/renderer/viz/expressions/now.js
```diff
@@ -28,12 +28,29 @@ export default class Now extends BaseExpression {
...
         super.inlineMaker = inline => inline.now;
     }
-    eval () {
+
+    get value () {
         return this.now.expr;
     }
+
+    eval () {
+        return this.value;
```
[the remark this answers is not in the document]
elenatorro: I really like this idea and I changed it :) However, I think we should also change then these methods in viewport expressions to be consistent, for example:

```js 
export default class ViewportAvg extends ViewportAggregation {
    constructor (property) {
        super({ property });
        this._sum = 0;
        this._count = 0;
    }

    get value () {
        return this._sum / this._count;
    }

    eval () {
        return this.value;
    }
```

But I am not sure if we should do this in this PR, what do you think?

src/renderer/viz/expressions/ramp.js
elenatorro: I do not see a performance problem, but I think this kind of improvements are really great, let me give it a try.

src/renderer/viz/expressions/now.js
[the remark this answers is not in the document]
davidmanzanares: Sure! that could be in another PR, it's out of the scope, and it's not a priority

Issue #780, opened 2018-07-30
top() and viewportHistogram cannot work together


davidmanzanares:
Since `top()` `eval` function returns the bucket number and `viewportHistogram` expects a string when the input is of type `category` the result of viewpotHistogrma is an empty array

PR #782, opened 2018-07-30
Fix top() eval()

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/780

### DIFF
--- a/src/renderer/viz/expressions/top.js
+++ b/src/renderer/viz/expressions/top.js
@@ -35,7 +35,7 @@ export default class Top extends BaseExpression {
         this.type = 'category';
     }
     eval (feature) {
-        const p = this.property.eval(feature);
+        const catID = this._meta.categoryToID.get(this.property.eval(feature));
         const buckets = Math.round(this.buckets.eval());
         const metaColumn = this._meta.properties[this.property.name];
         const orderedCategoryNames = [...metaColumn.categories].sort((a, b) =>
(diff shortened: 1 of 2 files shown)


src/renderer/viz/expressions/top.js
```diff
@@ -44,8 +44,8 @@ export default class Top extends BaseExpression {
 
         let ret;
         orderedCategoryNames.map((name, i) => {
-            if (i === p) {
-                ret = i < buckets ? i + 1 : 0;
+            if (i === catID) {
+                ret = i < buckets ? this._meta.IDToCategory.get(i) : 'CARTOVL_TOP_OTHERS_BUCKET';
```
Jesus89: Where is this used `CARTOVL_TOP_OTHERS_BUCKET`?

davidmanzanares: It's part of the new public API, see https://github.com/CartoDB/carto-vl/pull/782/files#diff-b603c20d9a415dae327ff3eb208e6794R67

Jesus89: OK, so this name is just to avoid collision with the categories

davidmanzanares: Yes, exactly

davidmanzanares: Using a Symbol (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) would be the right thing to do, but the browser support is not super good yet

Issue #775, opened 2018-07-27
Don't use `compile` to pass metadata information


davidmanzanares:
`compile` is only used when an expression is used for styling, if it's not (because it is a variable that is not referenced by a style property expression), then it's never called. This is a potential source of bugs.

PR #784, opened 2018-07-30
Extract `_bindMetadata` functionality from `_compile`

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/775

I fixed a few small things along the way.


src/renderer/viz/expressions/aggregation/global/GlobalAvg.js
```diff
@@ -26,7 +26,7 @@ import GlobalAggregation from './GlobalAggregation';
  * @function
  * @api
  */
-export default class GlobalCount extends GlobalAggregation {
+export default class GlobalAvg extends GlobalAggregation {
```
Jesus89: Good one

PR #785, opened 2018-07-30
Center GeoJSON dataframes at sampled centroid

### LINKED ISSUE #774: Render resolution bug
There is an issue in the render regarding resolution. In the following examples I use [this geojson](https://raw.githubusercontent.com/mapbox/mapbox-gl-js/master/docs/pages/assets/hike.geojson). Blue line is rendered with MGL, red line with VL.

line-resolution issue

I have uploaded the example in `debug/lines-geojson.html` in [this branch](https://github.com/CartoDB/carto-vl/tree/lines-resolution).

jgoizueta:
Fixes #774

A sample of the GeoJSON coordinates is taken and it's centroid computed. This is used as the origin of dataframe coordinates in order to increase available numeric precision.

### DIFF
--- a/src/sources/GeoJSON.js
+++ b/src/sources/GeoJSON.js
@@ -58,6 +58,8 @@ export default class GeoJSON extends Base {
         }
 
         this._features = this._initializeFeatureProperties(this._features);
+
+        this._setCoordinatesCenter();
     }
 
     bindLayer (addDataframe, dataLoadedCallback) {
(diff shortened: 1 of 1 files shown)


Jesus89:
Note: it seems that there is a little little diff, but much better result now.
selection_229

Jesus89:
Also, I'll try to create a test for this one.

src/sources/GeoJSON.js
```diff
@@ -58,6 +58,8 @@ export default class GeoJSON extends Base {
         }
 
         this._features = this._initializeFeatureProperties(this._features);
+
+        this._setCoordinatesCenter();
```
Jesus89: I would try to be more explicit here: 'this._dataframeCenter = this.coordinatesCenter();`

PR #789, opened 2018-07-31
Points with triangles

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/240

Points are rendered now with triangles (one triangle per point, as opposed to the previous implementation based on `GL_POINTS`, which imposed the 128-pixel limit).

For comprehension, you can see this (but of course, the triangle is transparent outside the points, this is just a tweak to show it):
points-tris

Points can be larger now:
bigpoints


jgoizueta:
vlwidth

Jesus89:
Why 1004?

With this approach, the size of the circles has changed. Is this new size the correct size? Compared with MGL or CartoCSS

Jesus89:
I have tested the size and it seems that `width` is 1px bigger than the size in pixels `marker-width`. So the fix is simple. However, `strokeWidth` seems to have the correct size.

Jesus89:
Here there is an animation: orange VL (diameter 90 -> 110), blue MGL (diameter 100). It seems that 98 or 99 do a better job that 100 in this case:



It also seems that `width` 18 fits better than 20. And for example in this case (diameter 600) there is no perfect fit since the jump is too big (due to the exponential decoding I guess)

davidmanzanares:
We are storing the width with 8 bits, it's not possible to make it perfect in every integer from 0 to 1024 with 8 bits. That's why we are using the exponential form, which is our own floating point format.

However, after taking a closer look, it seems that there were 24 more bits reserved for the other color components in `Dataframe`. Right now I think I can remove the exponential trick by using those forgotten bits.

Jesus89:
Nice 16 bit. We have more resolution now, but it seems that fits better with 2px less in our width.

| `enc.x*(255.*4.) + 4.*enc.y` | `enc.x*(255.*4.) + 4.*enc.y - 2.` |
|---|---|
| selection_241 | selection_240 | 
| selection_238 | selection_239 |
| selection_236 | selection_237 |

davidmanzanares:
I see, but MGL may not be right either. Our shader changes the width a little bit later to provide the antialiasing, it shouldn't be triggering this, but, of course, it is a potential source of bugs since the code is quite complex. We could count the pixel width with an image utility program and see if those 2 pixels are actually wrong.

Jesus89:
Yes, maybe the issue is in MGL roof. Btw, I have tested that currend decoding (without -2.) fits perfect with lines.

src/renderer/shaders/symbolizer/symbolizerVertexShader.glsl
```diff
@@ -17,37 +18,42 @@ uniform sampler2D filterTex;
...
-float decodeWidth(float x){
-    x*=255.;
-    if (x < 64.){
-        return x*0.25;
-    }else if (x<128.){
-        return (x-64.)+16.;
-    }else{
-        return (x-127.)*2.+80.;
-    }
+// From [0.,1.] in exponential-like form to pixels in [0., 1024.)
```
jgoizueta: exponential -> linear

src/renderer/shaders/geometry/line/lineVertexShader.glsl
```diff
@@ -14,25 +14,17 @@ uniform sampler2D filterTex;
...
-    float w;
-    if (x < 0.25098039215686274){ // x < 64/255
-        w = 63.75 * x; // 255 * 0.25
-    }else if (x < 0.5019607843137255){ // x < 128/255
-        w = x*255. -48.;
-    }else {
-        w = x*510. -174.;
-    }
-    return w;
+// From [0.,1.] in exponential-like form to pixels in [0., 1024.)
```
jgoizueta: exponential -> linear

src/renderer/shaders/geometry/point/pointVertexShader.glsl
```diff
@@ -19,49 +19,52 @@ uniform sampler2D filterTex;
...
-  float w;
-  if (x < 0.25098039215686274) { // x < 64/255
-    w = 63.75 * x; // 255 * 0.25
-  } else if (x < 0.5019607843137255) { // x < 128/255
-    w = x * 255. - 48.;
-  } else {
-    w = x * 510. - 174.;
-  }
-  return w;
+// From [0.,1.] in exponential-like form to pixels in [0., 1024.)
```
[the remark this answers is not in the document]
Jesus89: :point_up:

davidmanzanares:
Just for the record and since I was curious, I checked some screenshots with GIMP and I think that, at least at 10 pixels, our renderer behaves as expected.

10 pixel width with 1 pixel black stroke:
screenshot from 2018-08-01 17-29-04

Issue #756, opened 2018-07-23
Zoom and Pan not smooth 


rutgerhofste:
I'm using Carto-VL for the first time and looking for ways to improve the performance of my test app.  When I zoom in and out the performance is not smooth. 

My code can be found [here](https://github.com/rutgerhofste/carto_qa/tree/1d7369124092332d15a04c16d2b3eb2fa5aa5a70) in index.html

I'm combining several elements in Carto VL including SQL Join, Style and tooltip and I am wondering what is causing the slow performance. I'm on a windows machine with 16GB RAM and using Chrome. Chrome is using >1 GB RAM.

davidmanzanares:
Hi @rutgerhofste ,

We have found some of the performance problems on your example and at least some of the optimizations will be available on the next CARTO VL release (version `0.6.0`).

Thank you!

rutgerhofste:
Thank you, 

Looking forward to use Carto VL and CartoFrames.contrib.vector in more of our internal apps for quality assurance.

davidmanzanares:
Hi @rutgerhofste !

We just released CARTO VL version `0.6.0`. It comes with multiple optimizations, and some of them have an important impact on your example. We've tested your example again and we saw meaningful improvements. We still think that some optimizations are possible (https://github.com/CartoDB/carto-vl/issues/771) though.

Thank you for your feedback!

rutgerhofste:
Just updated the version to 0.6.0 and unfortunately I get an error:
https://rutgerhofste.github.io/carto_qa/

[rutgerhofste.github.io-1533225201237.log](https://github.com/CartoDB/carto-vl/files/2254182/rutgerhofste.github.io-1533225201237.log)

davidmanzanares:
Agh, sorry, could you try again with version `0.6.2`?

elenatorro:
Hi @rutgerhofste, is it still happening to you? Thanks!

PR #791, opened 2018-08-01
Polygon stroke

Jesus89:
## Related to https://github.com/CartoDB/carto-vl/issues/400

- [x] Refactor polygon decoder to use `addLine` function
- [x] Add closeLoop flag in `addLine` to be used in polygon stroke
- [x] Add skipCallback to `addLine` to skip strokes in connections or clipped
- [x] Fix stroke detection for PolygonShader
- [x] Add polygon-collection render tests
- [x] Reorder geometry render tests
- [x] Rename TriangleShader -> PolygonShader
- [x] Use `carto.on('loaded'` for multilayer tests
- [x] Fix z index between the polygon and its stroke
- [x] Optimize decode time (https://github.com/CartoDB/carto-vl/pull/799)
    - 2.2x faster line decoding
    - 1.7x faster polygon decoding

reference

reference


src/renderer/decoder/lineDecoder.js
```diff
@@ -72,7 +80,7 @@ function addLine (lineString, vertices, normals) {
                 // Third triangle
                 addTriangle(
                     [currentPoint, currentPoint, currentPoint],
-                    [[0, 0],
+                    [[0, 1e-37],
```
jgoizueta: What's that tiny value for?

Jesus89: In the shader is used this expression `normal == vec2(0.)` to check if the point belongs to the fill or stroke. The problem is that we have now a stroke vertex with normal (0,0). So adding this tiny value marks the vertex as "stroke vertex" but does not affect the position. Let me add a comment with this information.

src/renderer/shaders/geometry/polygon/polygonVertexShader.glsl
```diff
@@ -20,17 +20,23 @@ float decodeWidth(vec2 enc) {
...
+    // 64 is computed based on RTT_WIDTH and the depth buffer precision
+    // 64 = 2^(BUFFER_BITS)/RTT_WIDTH = 2^16/1024 = 64
+    float z = mod(featureID.y, 1./64.)*63. + featureID.x / (64.);
+
     vec4 c;
     if (normal == vec2(0.)){
+        z = 2.*z - 1.;
         c = texture2D(colorTex, featureID);
     }else{
+        z = 2.*z - 1. + 1./128.;
```
jgoizueta: So... this uses a separate set of depth values for the stroke than for the filling, right?

Jesus89: Yes, we need different z inside the same feature for fill and stroke. Since we are moving z (0, 1) to (-1, 1) I think we can cover the gaps with the z index for stroke and I think that's what this code does, but please take a closer look to be sure about it :wink:

src/renderer/decoder/polygonDecoder.js
```diff
@@ -1,78 +1,41 @@
...
 export function decodePolygon (geometry) {
-    let vertices = []; // Array of triangle vertices
+    let vertices = [];
     let normals = [];
-    let breakpoints = []; // Array of indices (to vertexArray) that separate each feature
+    let breakpoints = [];
     let featureIDToVertexIndex = new Map();
-    const geometryLength = geometry.length;
 
-    for (let i = 0; i < geometryLength; i++) {
```
davidmanzanares: I'm not the author of this micro-optimization, and, in general, we should avoid optimizations without profiling before and after the change. However, it's true that this function has been found to be a hot spot when profiling real applications. The micro-optimization of the outer loops probably don't have any real impact, but I wouldn't say the same for the inner loops, and clearly, for consistency, there is a reason to apply the same micro-optimization to the outer loop. For all of these, I would leave the code as it was.

src/renderer/shaders/geometry/polygon/polygonVertexShader.glsl
davidmanzanares: I think this line can overflow the valid `z` range of [-1,1]. I think we could leave the `z = 2.*z - 1.;` after both branches, and in the `else` branch have: `z = mod(z-0.5, 1.);`

test/integration/render/scenarios/geometries/line-string/width/scenario.js
```diff
@@ -1,10 +1,14 @@
-const map = new carto.Map({
+const map = new mapboxgl.Map({
```
davidmanzanares: What are the reasons for this change?

davidmanzanares:
I made some visual tests in `pluto` and it looks much better, I'm going to review performance a little bit too

davidmanzanares:
Performance has regressed quite a bit, `master` was on `17` ms for a super small dataset and in `20` ms for `world_borders`, this branch is on `31` and `37` respectively.

src/renderer/decoder/polygonDecoder.js
```diff
@@ -1,78 +1,41 @@
...
-
-                // Second triangle
-                addTriangle(
-                    [a, b, b],
-                    [neg(normal), neg(normal), normal]
-                );
-            }
+            addLine(lineString, vertices, normals, true, (index) => {
+                // Skip adding the line which connects two rings OR is clipped
+                return polygon.holes.includes((index - 2) / 2) || isClipped(polygon, index - 4, index - 2);
```
davidmanzanares: Having a callback may be one of the reasons for the performance regression.

src/renderer/decoder/common.js
```diff
@@ -0,0 +1,92 @@
...
+    let skipLine = false;
+
+    // We need at least two points
+    if (lineString.length >= 4) {
+        // Initialize the first two points
+        prevPoint = [lineString[0], lineString[1]];
+        currentPoint = [lineString[2], lineString[3]];
+        prevNormal = getLineNormal(prevPoint, currentPoint);
+
+        for (let i = 4; i <= lineString.length; i += 2) {
```
davidmanzanares: The `length` micro-optimization has been removed from here too.

src/renderer/decoder/lineDecoder.js
```diff
@@ -1,13 +1,13 @@
...
 import { getFloat32ArrayFromArray } from '../../utils/util';
-import { getJoinNormal, getLineNormal, neg } from '../../utils/geometry';
 
 // If the geometry type is 'line' it will generate the appropriate zero-sized, vertex-shader expanded triangle list with `miter` and `bevel` joins.
 // The geom will be an array of coordinates in this case
 
 export function decodeLine (geometry) {
     let vertices = [];
     let normals = [];
-    let breakpoints = []; // Array of indices (to vertexArray) that separate each feature
```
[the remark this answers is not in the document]
Jesus89: OK, I thought it was not needed

src/renderer/decoder/polygonDecoder.js
Jesus89: Do you mean `geometryLength` by `geometry.length`?

test/integration/render/scenarios/geometries/line-string/width/scenario.js
Jesus89: I just moved a test in mgl-integration here and replace the previous one as part or `geometry tests` reorder. This cover lines with MGL.

src/renderer/decoder/polygonDecoder.js
davidmanzanares: Yes

Jesus89: I have tested the performance, and there is no optimization by doing that:

https://jsperf.com/arraylengthprecaching

Jesus89: Agree. But I see also in your example, in general, that there is no difference between the cached and not cached length (sometimes even `without caching length` is a little bit better).

src/renderer/decoder/common.js
Jesus89: https://github.com/CartoDB/carto-vl/pull/791#discussion_r207173054

Jesus89:
Optimization branch: https://github.com/CartoDB/carto-vl/pull/799

src/renderer/decoder/common.js
```diff
@@ -0,0 +1,123 @@
...
+            drawLine = !(skipCallback && skipCallback(i));
+
+            if (drawLine) {
```
davidmanzanares: It would be nice to document how each triangle is positioned... Could we put a link to an image?

Jesus89: You can find more information here: https://github.com/CartoDB/carto-vl/pull/219. I could add the information in the code.

davidmanzanares:
I benchmarked this PR by using the example on https://github.com/CartoDB/carto-vl/issues/756

The decoding of the tiles and the upload to the GPU process improve the execution time from `6.4` seconds in `master` to `4.7` seconds in this branch. The time spent in decoding was decreased by a lot, but the time to upload the vertices to the GPU increased noticeably since there are more vertices. Anyway, this is an important improvement, particularly since the example main pain point is this time.

However, the rendering times increased quite a bit. I made different tests and I made a patch to fix part of the performance regression:

| Code       | Average Time per Frame           |
| ------------- |:-------------:|
| master      | 33.2 |
| PR      | 52.6 |
| PR + remove extra triangles + Z patch      | 33.2 |
| PR + remove extra triangles    | 38.5 |
| PR + Z patch    | 47.6 |

"remove extra triangles" means removing the additional 2 triangles that are added by replacing `drawLine` in `common.js` line 63 by `false`.

The Z patch is a small change to remove unnecessary WebGL calls from the tileRenderLoop, extracting them before, calling them only once, I'll create a PR soon.

With those times what I see is that we have lost 37% of performance. The lost is divided by the Z changes 27% (the regression dissappears with the patch) and a 73% by the increase in vertex count.

Jesus89:
Thanks @davidmanzanares for the rendering analysis. I'll wait until the Z patch to merge this branch.

Yes, there are more vertices since there are new triangles for the joins in polygon strokes (the same as the lines). In order to improve the number of vertices I see two possible approaches:

- Evaluate a condition to skip those join triangles (when there are a huge number of lines per tile, for example)
- Use different shaders for polygon and polygon stroke, and use LINE_STRIP in polygon stroke and lines. This one requires more research to evaluate the viability.

I agree in merging this one after your path for WebGL calls and then evaluate Rendering optimizations in future work.

davidmanzanares:
:+1:

Your first point would be the first thing I would try. The second point is complicated though, since WebGL have a line width limit of `1` in most implementations, making rendering native lines almost useless, see https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/lineWidth

Jesus89:
Sorry, I wanted to say TRIANGLE_STRIP instead of LINE_STRIP.

Issue #771, opened 2018-07-26
Optimize `decodePolygon`


davidmanzanares:
It's one of the hot spots on many apps, sometimes even the main one.

I think that it could be optimized by removing the usage of the temporal `vertices` and `normals` arrays.

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/799.

davidmanzanares:
Please, close the issue when the fix comes to `master`

Jesus89:
Right! Closing now by https://github.com/CartoDB/carto-vl/pull/791

PR #799, opened 2018-08-02
Decode optimization

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/400, https://github.com/CartoDB/carto-vl/issues/771.

In the `polygon stroke` branch the polygon is properly rendered (bevel and miter joins) but with the current implementation, load time is twice slower compared with master (with the primitive support for strokes).

This PR introduces an optimization for line and polygon stroke decoding to improve load time to values even better than master but with the correct polygon stroke rendering.

### Polygon decoding

Improve decode time **1.7x faster**.

**Decode time**

| `master` | `polygon-stroke` | `polygon-stroke + decode-optimization` |
|:-:|:-:|:-:|
| 1x | 0.5x | 1.7x |

### Line decoding

Improve decode time **2.2x faster**.

**Decode time**

| `master` |`polygon-stroke + decode-optimization` |
|:-:|:-:|
| 1x | 2.2x |

## Optimization approach

The main approach is to do a better management of the memory regarding the geometry buffers. I noticed that `array.push(item)` allocates new memory for the entire buffer + 1 and then sets the value. This becomes slower when the array increases its size. So we need to allocate the memory first and then set the values to improve performance.

The approach of computing the exact number of vertices first, allocate memory and then set the values was discarded because obtaining the number of vertices adds more time than the time saved with only one memory allocation. Even using the maximum number of vertices plus a slice is not a good option too because it implies two memory allocations.

The final approach was to use a **resizable static buffers** of Float32Array instead of Array for both `vertices` and `normals` and then do a slice to get the final array. The buffers are resized to twice its capacity when there is an overflow with the current size. Sharing this memory allows reducing the time with the same results. The default memory for the static buffer is 4MB.

I get rid of `getFloat32ArrayFromArray` function in favor of `vertices.slice(0, index)` to get the final new buffers. Also tested functions like `new Float32Array(geomBuffer.vertices.subarray(0, index))` and `new Float32Array(new Float32Array(geomBuffer.vertices.buffer, 0, index))` but slice is the best option.

There are also other micro-optimizations like improve intermediate functions, optimize draw logic, using a common index, native `set` function instead of `for loop`, etc and adding the `addVertex` function, that improves performance by 10%.

## Performance tests

I have used local performance tests with local GeoJSON data, traces with `performance.now()` in the code to test the tiles and also the Chrome profiling tool to get the decode times and obtain the final results. 

```js
import { decodeGeom } from '../../../../src/renderer/decoder';
import GeoJSON from '../../../../src/sources/GeoJSON';

// let data = require('./sf_stclines.json');
let data = require('./mnmappluto.json');

const geojson = new GeoJSON(data);
const polygonGeometry = geojson._decodeGeometry();

falcon.benchmark('decodePolygon', () => {
    decodeGeom('polygon', polygonGeometry);
}, {runs: 1});
```

### DIFF
--- a/src/renderer/decoder/lineDecoder.js
+++ b/src/renderer/decoder/lineDecoder.js
@@ -1,33 +1,36 @@
-import { addLine } from './common';
+import { addLineString } from './common';
 import { getFloat32ArrayFromArray } from '../../utils/util';
 
 // If the geometry type is 'line' it will generate the appropriate zero-sized, vertex-shader expanded triangle list with `miter` and `bevel` joins.
 // The geom will be an array of coordinates in this case
 
-let vertices = new Array(1000000);
-let normals = new Array(1000000);
+const geomBuffer = {
+    index: 0,
+    vertices: new Array(1000000),
+    normals: new Array(1000000)
+};
 
 export function decodeLine (geometry) {
     let breakpoints = []; // Array of indices (to vertexArray) that separate each feature
     let featureIDToVertexIndex = new Map();
-    let vertexIndex = 0;
 
+    geomBuffer.index = 0;
     for (let i = 0; i < geometry.length; i++) {
         const feature = geometry[i];
         for (let j = 0; j < feature.length; j++) {
-            vertexIndex = addLine(feature[j], vertices, normals, vertexIndex);
+            addLineString(feature[j], geomBuffer);
         }
 
         featureIDToVertexIndex.set(breakpoints.length, breakpoints.length === 0
-            ? { start: 0, end: vertexIndex }
-            : { start: featureIDToVertexIndex.get(breakpoints.length - 1).end, end: vertexIndex });
+            ? { start: 0, end: geomBuffer.index }
+            : { start: featureIDToVertexIndex.get(breakpoints.length - 1).end, end: geomBuffer.index });
 
-        breakpoints.push(vertexIndex);
+        breakpoints.push(geomBuffer.index);
     }
 
     return {
-        vertices: getFloat32ArrayFromArray(vertices, vertexIndex),
-        normals: getFloat32ArrayFromArray(normals, vertexIndex),
+        vertices: getFloat32ArrayFromArray(geomBuffer.vertices, geomBuffer.index),
+        normals: getFloat32ArrayFromArray(geomBuffer.normals, geomBuffer.index),
         featureIDToVertexIndex,
         breakpoints
     };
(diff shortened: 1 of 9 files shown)


src/renderer/decoder/lineDecoder.js
```diff
@@ -4,29 +4,30 @@ import { getFloat32ArrayFromArray } from '../../utils/util';
...
+            : { start: featureIDToVertexIndex.get(breakpoints.length - 1).end, end: vertexIndex });
 
-        breakpoints.push(vertices.length);
+        breakpoints.push(vertexIndex);
     }
 
     return {
-        vertices: getFloat32ArrayFromArray(vertices),
-        normals: getFloat32ArrayFromArray(normals),
+        vertices: getFloat32ArrayFromArray(vertices, vertexIndex),
```
davidmanzanares: Since `vertices` must be copied to a `Float32Array` anyway having the static buffers seems like a really good approach :rocket:

src/renderer/decoder/polygonDecoder.js
```diff
@@ -27,27 +29,30 @@ export function decodePolygon (geometry) {
             const triangles = earcut(polygon.flat, polygon.holes);
             for (let k = 0; k < triangles.length; k++) {
```
davidmanzanares: Could we check for overflow here and in `addLine`, and if necessary, make the buffer 2 times larger?

Jesus89: This is an attempt to address this feature: https://github.com/CartoDB/carto-vl/pull/804

Jesus89:
Resize static buffer PR: https://github.com/CartoDB/carto-vl/pull/804

src/utils/geometry.js
```diff
@@ -67,19 +67,12 @@ export function neg (v) {
...
-    return [p2[0] - p1[0], p2[1] - p1[1]];
-}
-
 /**
  * Return the vector scaled to length 1
  */
-function normalize (v) {
-    const s = Math.hypot(v[0], v[1]);
-    return [v[0] / s, v[1] / s];
+function normalize (v1, v2) {
```
davidmanzanares: If `v1` and `v2` are `x` and `y` coordinates then `v1` and `v2` are bad names

Jesus89: No. v1 and v2 are the start and end of a vector. I borrow the name from the Math.hypo docs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot

davidmanzanares: `v1` and `v2` cannot be vectors because `Math.hypot` receives scalars. Please, review also the usage of this function

davidmanzanares: Yes, they are using those names and `v` comes from `value` if you read their docs, and they have to use numeric values since the function is generic and is not limited to 2D.

In our context, however, we work in 2D and `v1` and `v2` can be confused by `vertex1` and `vertex2`

PR #800, opened 2018-08-02
Enhanced featureID to Z (depth) mapping

jgoizueta:
Our mapping from featureID to Z wasn't optimal to avoid collisions, since the y value of the featureID is already spread as much as possible in the -1, 1 range (so the Ys were colliding a lot).

The effect of this is barely noticeable because collisions were mostly avoided by the X, and collisions would produce visible effects only when colliding features overlap... but well, this is theoretically better.

### DIFF
--- a/src/renderer/shaders/geometry/polygon/polygonVertexShader.glsl
+++ b/src/renderer/shaders/geometry/polygon/polygonVertexShader.glsl
@@ -22,14 +22,16 @@ float decodeWidth(vec2 enc) {
 void main(void) {
     // 64 is computed based on RTT_WIDTH and the depth buffer precision
     // 64 = 2^(BUFFER_BITS)/RTT_WIDTH = 2^16/1024 = 64
-    float z = mod(featureID.y, 1./64.)*63. + featureID.x / (64.);
+    float z = featureID.y * 63. / 64. + featureID.x / (64.);
 
     vec4 c;
     if (normal == vec2(0.)){
         z = 2.*z - 1.;
         c = texture2D(colorTex, featureID);
     }else{
-        z = mod(z - 0.5, 1.);
+        z = mod(z + 1./64., 1.);
+        z = 2.*z - 1.;
+
         c = texture2D(strokeColorTex, featureID);
     }
     float filtering = texture2D(filterTex, featureID).a;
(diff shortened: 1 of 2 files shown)


src/renderer/shaders/geometry/polygon/polygonVertexShader.glsl
```diff
@@ -22,14 +22,16 @@ float decodeWidth(vec2 enc) {
...
+    float z = featureID.y * 63. / 64. + featureID.x / (64.);
 
     vec4 c;
     if (normal == vec2(0.)){
         z = 2.*z - 1.;
         c = texture2D(colorTex, featureID);
     }else{
-        z = mod(z - 0.5, 1.);
+        z = mod(z + (z > 0.5 ? -1./64. : 1./64.), 1.);
+        z = 2.*z - 1.;
```
davidmanzanares: ` z = 2.*z - 1.;` can be extracted after the branches

src/renderer/shaders/geometry/polygon/polygonVertexShader.glsl
```diff
@@ -22,14 +22,16 @@ float decodeWidth(vec2 enc) {
...
-    float z = mod(featureID.y, 1./64.)*63. + featureID.x / (64.);
+    float z = featureID.y * 63. / 64. + featureID.x / (64.);
 
     vec4 c;
     if (normal == vec2(0.)){
         z = 2.*z - 1.;
         c = texture2D(colorTex, featureID);
     }else{
-        z = mod(z - 0.5, 1.);
+        z = mod(z + (z > 0.5 ? -1./64. : 1./64.), 1.);
```
davidmanzanares: I don't understand this change, how does it improve the previous `z = mod(z - 0.5, 1.);`?

Jesus89:
I add a comparison I found with and without these changes. For some zoom levels z index was not properly computed.

| Before | After |
|---|---|
| selection_246 | selection_245 |

src/renderer/shaders/geometry/polygon/polygonVertexShader.glsl
jgoizueta: the feature id texture is filled by rows, and rows spread, then we map each row to a dense block in z space, blocks being spread now in z-space, so space adjacent to the row blocks is the last to get used as all the available rows are filled. The new approach places stroke textures in adjacent dense blocks... 😅 does it make any sense?? I'm afraid I can't explain it without a sheet of paper!

davidmanzanares: > . The new approach places stroke textures in adjacent dense blocks

I think I understand what you mean there, but I don't understand why does that improve the previous code :confused:

PR #802, opened 2018-08-03
PoC of resizing the decoding buffer

davidmanzanares:


Jesus89:
In `decodeLine` and `decodePolygon` check the max buffer size to resize the static buffer if required.

Default values for line buffers are 4Mb and for polygon buffers to 8Mb

PR #803, opened 2018-08-03
Change test image size to be consistent with real DOM size

davidmanzanares:
The map size was set to 500x500, but we were taking image sizes of 400x300. This is a problem for debugging, particularly viewport functions, which were already being tested in render tests.


davidmanzanares:
Actually, this doesn't make any sense. The map size is set to that only on the viewport examples, where the value there, and only there, is not consistent with the test image size. That should be addressed by other PR.

PR #804, opened 2018-08-03
Resize static buffers

Jesus89:
In `decodeLine` and `decodePolygon` check the max buffer size to resize the static buffer if required.

Default values for line buffer are set to 4Mb, for polygon buffer to 8Mb


Jesus89:
More work is done here to improve resize buffer results: https://github.com/CartoDB/carto-vl/pull/802. :rocket:

Closing this.

Issue #805, opened 2018-08-03
Test minified build


davidmanzanares:
After having problems with the minification configuration (the minification produced invalid code), I think we may add some form of testing at some level.

Jesus89:
Now we are using the non-minified bundle to pass the tests. We could use the minified bundle for the tests or even pass the tests with both bundle files.

Jesus89:
Hi. I have created a PR with this approach: https://github.com/CartoDB/carto-vl/pull/812

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/812

PR #812, opened 2018-08-07
Test min bundle

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/805

- Use minified bundle for testing in CI
- Disable bundle warnings for testing
- Clean unneeded packages
- Update bundle size limit to 260KB


davidmanzanares:
But, does it make sense to change all the tests every time to the minified build? I'm worried this could be counterproductive for debugging purposes.

cc: @rochoa

Jesus89:
If you want to cover the minified bundle without adding extra tests this is the only way. However, I could add a flag to use only minified testing in CI, enabling the uncompressed bundle for testing in development.

rochoa:
For local, I would try to avoid minified versions, but I see the value of testing the minified version in CI.

rochoa:
> If you want to cover the minified bundle without adding extra tests this is the only way. However, I could add a flag to use only minified testing in CI, enabling the uncompressed bundle for testing in development.

That would be the best approach, I guess.

Issue #606, opened 2018-06-22
Mention `filter: animation(...)` combination in the docs


davidmanzanares:
Particularly here: https://carto.com/developers/carto-vl/reference/#vizspec

PR #816, opened 2018-08-07
Add animation reference to filter in Viz docs

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/606

### DIFF
--- a/src/Viz.js
+++ b/src/Viz.js
@@ -42,6 +42,24 @@ const SUPPORTED_PROPERTIES = [
     'variables'
 ];
 
+/**
+ * A vizSpec object is used to create a {@link carto.Viz|Viz} and controlling multiple aspects.
+ * For a better understanding we recommend reading the {@link TODO|VIZ guide}
+ * @typedef {object} VizSpec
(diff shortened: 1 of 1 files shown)


src/Viz.js
```diff
@@ -42,6 +42,24 @@ const SUPPORTED_PROPERTIES = [
     'variables'
 ];
 
+/**
+ * A vizSpec object is used to create a {@link carto.Viz|Viz} and controlling multiple aspects.
+ * For a better understanding we recommend reading the {@link TODO|VIZ guide}
```
Jesus89: Could we add now a link for this?

davidmanzanares: We should, although I don't know how should we point to a guide. @csubira do you know how should we do it?

csubira: I'd suggest following the same pattern that here in carto-js https://github.com/CartoDB/carto.js/blob/master/src/api/v4/filter/bounding-box-gmaps.js#L11 with the full link. Would that work for you?

davidmanzanares: :+1:

Issue #529, opened 2018-06-07
Add syntax sugar for imageList expression


elenatorro:
Related with: https://github.com/CartoDB/carto-vl/pull/343

Now, when we define a set of sprites in order to pass them to a ramp, we use the following syntax:

```js
const viz = new carto.Viz(`
    symbol: ramp(buckets($pop_max, [1000000]), sprites([sprite('./marker_1.svg'), sprite('./marker_2.svg')]))
    symbolPlacement: align_bottom
`);
```

It would be nice to be able to avoid the `sprites` "method" and use the expression as follows:

```js
const viz = new carto.Viz(`
    symbol: ramp(buckets($pop_max, [1000000]), [sprite('./marker_1.svg'), sprite('./marker_2.svg')]))
    symbolPlacement: align_bottom
`);
```

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/818

PR #818, opened 2018-08-07
Images array

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/529

This branch adds support for internal `image-array` type to be used inside ramps. As we did for numbers `[1,2,3]`, categories `['a', 'b', 'c']` and colors [red, green, blue], we do the same for images:

 ```js
[car, triangle, marker]
```
instead of

```js
imageList([car, triangle, marker])
```

With these changes I have the feeling that we should deprecate `imageList` because it does not add any value to the API. If we finally decide to do so I will update a bit the code implementation and also the tests.

cc @rochoa @davidmanzanares @makella @elenatorro


rochoa:
What about the JS API?

Jesus89:
For the JS API it is the same: a JS array containing images.

```js
const viz = new carto.Viz(`
  symbol: ramp(
    buckets($featurecla, ["Admin-0 capital","Admin-1 capital","Populated place"]),
    [star,triangle,marker])
`);
```

```js
const s = carto.expressions;
const viz = new carto.Viz({
  symbol: s.ramp(s.buckets(s.prop('featurecla'),
    ["Admin-0 capital","Admin-1 capital","Populated place"]),
    [s.image('star.svg'),s.image('triangle.svg'),s.image('marker.svg')])
});
```

rochoa:
Being `star.svg` a valid URL, right?

Jesus89:
Yes. It should be next to the HTML file in this case.

makella:
Makes sense!

Jesus89:
Yes. I had the same question.

You can access the provided images in the String API but didn't find how to do this in JS API. My two cents for `s.image('car')`. I also vote for using namespaces in both String and JS APIs for our internal constants to avoid future collisions. Something like `[markers.car]`.

Since this proposal is out of the scope of this PR, I can create another issue for that.

davidmanzanares:
It is already available in the JS API with:
```
    const viz = new carto.Viz({
      symbol: carto.expressions.CAR
    });
```

Regarding the namespace, that was discussed with Ariana when we developed the feature and we decided to avoid them to reduce extra characters.

Jesus89:
Oh great. However, in order to be consistent with the palettes (`carto.expressions.palettes.PRISM`) I would add a namespace for images, at least in the JS API, because `carto.expressions.CAR`is very confusing, because CAR is not an expression, so `carto.expressions.images.CAR` would be much better IMHO. Also since we are using `const s = carto.expressions;` this will be `s.images.CAR`, the same as `s.palettes.PRISM`.

There is room for collisions in the String API because there are no namespaces at all so we could think also about that.

davidmanzanares:
` CAR is not an expression` From my point of view, it is an expression, actually, it is `new SVG(svgs.bicycle)` and SVG is an expression class.

Regarding namespaces and collisions, I don't think this is a priority right now since there are more important problems and feature requests.

Jesus89:
We are talking about high-level semantics here. `Palettes` are expressions also (everything inside Viz is an expression), but we are using `s.palettes.PRISM` to tell the user that this, apart from an expression, is a Palette. And we should use the same concept for `s.images.CAR`.

It is not about high priority, it's about an easy change to add consistency to the API.

NOTE: this is off topic about the goal of the issue, so it can be CR and merged apart from this discussion.

src/renderer/viz/expressions/ramp.js
```diff
@@ -87,6 +88,10 @@ export default class Ramp extends BaseExpression {
         input = implicitCast(input);
         palette = implicitCast(palette);
 
+        if (palette && palette.type === paletteTypes.IMAGE_ARRAY) {
+            palette = new ImageList(palette.elems);
```
davidmanzanares: The cast should be done on `implicitCast` not here

examples/styling/triangle.svg
```diff
@@ -0,0 +1 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 15 15"><path d="M7.538 2c-.294 0-.488.177-.615.385l-5.846 9.538C1 12 1 12.153 1 12.308c0 .538.385.692.692.692h11.616c.384 0 .692-.154.692-.692 0-.154 0-.231-.077-.385l-5.77-9.538C8.029 2.177 7.789 2 7.539 2z"/></svg>
```
davidmanzanares: Why is this SVG added here? I think `triangle.svg` is a built-in

Jesus89: Yes, I accidentally added this file. Removing it.

Issue #770, opened 2018-07-26
`in` String API syntax sugar


davidmanzanares:
Allow
`filter: $cat in ['a', 'b', 'c']`

davidmanzanares:
I created this issue since @makella expected similar behavior and I think this is consistent with previous syntax sugar of the String API, but this opens up the issue of syntax sugar for the similar functions: `NIN` and `Between`.

Those 2 are much more problematic:

`filter: $cat nin ['a', 'b', 'c']` Using `nin` seems like a bad name here

`filter: $cat not in ['a', 'b', 'c']` This is difficult to parse with our current parser
`filter: $num between 10 30` This is difficult to parse with our current parser

What do you think?

cc: @rochoa

Jesus89:
I think in this case `filter: $cat nin ['a', 'b', 'c']` has perfect sense, since it has the same sematics than the Function API `filter: nin($cat, ['a', 'b', 'c'])`. So I vote for this option.

As you say, other options are more difficult to parse or not very consistent with our current types. NOTE: I think for `between($num, 10, 30)` the better simple approach is using `10 < $num < 30`.

makella:
Some thoughts:


> Using `nin` seems like a bad name here

Regarding this, are there other options? I realize that we are using logical operators but just throwing out some comparison ones too :)  
    - `notIN`
    - `notin`
    - `<>`
    - `!=`

>`filter: $num between 10 30` This is difficult to parse with our current parser
>`filter: $cat nin ['a', 'b', 'c']`

Is there a reason that numbers and category notation would be different? 

Could we simplify and do something like:

`filter: between($num, [10, 30])` 
  - where we would know in the background that it is values > 10 and < 30?
  - for ranges of numbers, I feel like this is similar to `width: ramp($num,[2,25])`

And for categories:

`filter: notin($cat, ['a','b','c'])`

> the better simple approach is using 10 < $num < 30

Not sure I totally agree that it is a simpler approach!

davidmanzanares:
> Is there a reason that numbers and category notation would be different?

Well, between always takes 2 arguments, therefore, I think it's normal to don't pass a list with "[" and "]". However, `in` takes a list of categories.

rochoa:
Always have in mind that the more different the syntax is between String vs JS APIs, the more difficult is to move between them.

makella:
[...]

PR #826, opened 2018-08-08
Add in/nin syntax sugar

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/770


makella:
>Do you think this should be updated too in the docs to be the preferred option?

definitely!

Issue #795, opened 2018-08-02
Add `strokeCap` property


Jesus89:
Part of https://github.com/CartoDB/carto-vl/issues/738

- [ ] Butt
- [ ] Square
- [ ] Round

PR #828, opened 2018-08-09
Add strokeCap property

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/795

This PR adds the new property `strokeCap`, that defines the shape of the endpoints in lines. Supported types are `BUTT` and `SQUARE`, located in the namespace `carto.expression.caps`.

```js
const s = carto.expressions;
const style = new carto.Viz({
    width: 20,
    strokeCap: s.caps.BUTT
});
```

| BUTT | SQUARE |
|---|---|
| reference | reference |

This property is an Enum (number), the same as resolution. When the property changes, it requires to **re-decode the Dataframe geometry**. To implement this behaviour, the Layer object contains a `geomOptions` object which is passed to `requestData` of the source. The new dataframes are instantiated with this options. The cached dataframes are updated with this options if they have changed.

Unit and Render tests are added to cover the implemented feature.


elenatorro:
I agree that we need more documentation here, but not only with `strokeCap` property. I think we should add it in the `strokeJoin` new property PR (https://github.com/CartoDB/carto-vl/pull/827) so we can document both properties together.

We could add something _similar_ to this [SVG guide](http://apike.ca/prog_svg_line_cap_join.html) (screenshots below) in the devcenter. What do you think? cc @makella @rochoa 

captura de pantalla 2018-08-13 a las 9 32 53
captura de pantalla 2018-08-13 a las 9 33 00

Jesus89:
OK. I'll merge this one into the branch `stroke-join` and then do a full review there.

PR #836, opened 2018-08-10
Properly invoke isAnimated function

### LINKED ISSUE #835: Performance regression
Using the `/examples/basics/add-layer.html` example.

## master

master

## v0.6.2

v0 6 2

You can also see visual differences, references https://github.com/CartoDB/carto-vl/issues/832.

rochoa:
Fixes #835.

### DIFF
--- a/src/Viz.js
+++ b/src/Viz.js
@@ -180,7 +180,7 @@ export default class Viz {
     }
 
     isAnimated () {
-        return this._getRootStyleExpressions().some(expr => expr.isAnimated);
+        return this._getRootStyleExpressions().some(expr => expr.isAnimated());
     }
 
     onChange (callback) {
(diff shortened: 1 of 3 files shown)


rochoa:
The failing test was changed in the previous PR, see https://github.com/CartoDB/carto-vl/pull/797/files?short_path=d4418fd#diff-d4418fd12cc1ea8ec774190a611a76fd.

The test was working because of the visualization being animated all the time. Actually, if you force a repaint, it shows correctly.

top

I'm a bit lost, any help is more than welcome. 🙏

Jesus89:
NOTE: it seems that in our MGL integration `paintCallback` is called twice for non-animated layers, so this bug doesn't arise using MGL maps. There is also a bug regarding the addLayer logic and `_registerMoveObserver` is registered twice for the same layer but with different `uids`. Also I found that `layer.initialize` should be async because of `requestMetadata` so we should use `await layer.initialize()` in the painter callback to fix the inconsistency. I'll create an issue for that.

Issue #838, opened 2018-08-10
Add max resolution to documentation?


ztephm:
### Context

When a resolution greater than 255 is used there is a console error. 

#### Example 1:

![screen shot 2018-08-10 at 5 57 33 pm](https://user-images.githubusercontent.com/1779444/43983217-6834683e-9cc7-11e8-918d-c1bc2a926ed3.png)

The problem is that this error is not easily visible to users - the map just appears blank until they look in the console:

#### Example 2:

![screen shot 2018-08-10 at 6 02 02 pm](https://user-images.githubusercontent.com/1779444/43983237-886ab234-9cc7-11e8-8838-5c9d543cbe33.png)

```
const viz = new carto.Viz(`
     color: namedColor('DeepSkyBlue')
     width: 5
     strokeWidth: 1
     strokeColor: blue
     resolution: 256
`);
```

full code: [index.html.zip](https://github.com/CartoDB/carto-vl/files/2279587/index.html.zip)
dataset: [ne_10m_populated_places_simple.zip](https://github.com/CartoDB/carto-vl/files/2279588/ne_10m_populated_places_simple.zip)

What do you think of adding the max resolution to [our documentation](https://carto.com/developers/carto-vl/reference/#vizspec)?

PR #840, opened 2018-08-13
Add resolution range in documentation

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/838


src/Viz.js
```diff
@@ -55,7 +55,7 @@ const SUPPORTED_PROPERTIES = [
  * @property {Placement} symbolPlacement - when using `symbol`, offset to apply to the image
  * @property {Placement} offset - offset to apply to the features in pixels
  * @IGNOREproperty {Order} order - rendering order of the features, only applicable to points
- * @property {number} resolution - resolution of the property-aggregation functions, a value of 4 means to produce aggregation on grid cells of 4x4 pixels, only applicable to points
+ * @property {number} resolution - resolution, from 1 to 255, of the property-aggregation functions, a value of 4 means to produce aggregation on grid cells of 4x4 pixels, only applicable to points
```
davidmanzanares: @jgoizueta can confirm, but I think that the minimum value is slightly above `0`, like `0.000001`

jgoizueta: Yes the backend supports less than 1 resolutions (https://github.com/CartoDB/Windshaft-cartodb/blob/master/docs/aggregation.md#resolution). (so that 0.5 would be a quadrant of a pixel).

The backend actually only checks that `resolution` is [greater than 0](https://github.com/CartoDB/Windshaft-cartodb/blob/204d246a8cf227bbfde85d7fd662a97bd73fd8c3/lib/cartodb/models/aggregation/aggregation-mapconfig.js#L69), but very small values would cause timeouts, so it would be better to have a explicit minimum.

elenatorro: ok! I am going to change that :) thanks!

elenatorro:
Question 👼: in the [torque documentation](https://carto.com/developers/torque-js/guides/cartocss/), it says:

> -torque-resolution Spatial resolution in pixels. A resolution of 1 means no spatial aggregation of the data. Any other resolution of N results in spatial aggregation into cells of NxN pixels. The value N must be power of 2.

I changed the 4x4 example to the NxN example. Should we add that 'N' does not need to be power of 2?

cc @makella @davidmanzanares

src/Viz.js
```diff
@@ -55,7 +55,7 @@ const SUPPORTED_PROPERTIES = [
  * @property {Placement} symbolPlacement - when using `symbol`, offset to apply to the image
  * @property {Placement} offset - offset to apply to the features in pixels
  * @property {Order} order - rendering order of the features, only applicable to points. See {@link carto.expressions.asc}, {@link carto.expressions.desc} and {@link carto.expressions.noOrder}
- * @property {number} resolution - resolution of the property-aggregation functions, a value of 4 means to produce aggregation on grid cells of 4x4 pixels, only applicable to points
+ * @property {number} resolution - resolution of the property-aggregation functions, only applicable to points. Default resolution is 1. Custom values must be greater than 0 and lower than 256. A resolution of N means points are aggregated to grid cells NxN pixels. Unlinke {@link https://carto.com/developers/torque-js/guides/how-spatial-aggregation-works/|Torque resolution}, the aggregated points are placed in the centroid of the cluster, not in the center of the grid cell.
```
VictorVelarde: For the 3 descriptions:
- I would add "of NxN pixels" as in original one. 
- I also like the idea of adding something regarding to `'N' does not need to be power of 2` (as that is the usual in other libraries).

Issue #522, opened 2018-06-07
Properties with spaces


rochoa:
> With respect to $priceRangeLow variables, it’s possible to define a column name in pgsql with spaces, eg `SELECT 1 AS "foo bar"`, presumably this Won’t Work (tm) (would it be $“foo bar” or “$foo bar” who knows) but does it have any parser/security implications?

cc @pramsey

rochoa:
So the way to solve this situation is to use `width: prop('foo bar')`.

Jesus89:
I have included this information in the guides documentation: https://github.com/CartoDB/carto-vl/pull/815.

I think we can close this after merging the PR.

cc @rochoa

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/815

rochoa:
This is broken for the SQL source.

Example:
```js
const source = new carto.source.SQL('SELECT *, 99 as "best price" FROM populated_places_simple', {
    apiKey: 'default_public',
    user: 'rochoa'
});

const viz = new carto.Viz(`width: prop('best price')`);
const layer = new carto.Layer('layer', source, viz);
```

However, that, for instance, works with the GeoJSON source.

Jesus89:
I suppose that `select.map(c => `"${c}"`)` in the query should be added. We need to care also about `dimensions`. The request should be encoded with LZMA to allow valid characters `"`.

Jesus89:
### Before

This is the current map config:

```js
{
    "type": "mapnik",
    "options": {
        "sql": "SELECT best price,cartodb_id,the_geom,the_geom_webmercator FROM (SELECT *, 99 as \"best price\" FROM populated_places_simple) as _cdb_query_wrapper ",
        "aggregation": {
            "columns": {},
            "dimensions": {
                "best price": "best price"
            },
            "placement": "centroid",
            "resolution": 1,
            "threshold": 1
        },
        "dates_as_numbers": true,
        "metadata": {
            "geometryType": true,
            "columnStats": {
                "topCategories": 32768,
                "includeNulls": true
            },
            "sample": {
                "num_rows": 1000,
                "include_columns": [
                    "best price",
                    "cartodb_id"
                ]
            }
        }
    }
}
```

And the Windshaft response

```
Error: Maps API error: {"errors":["column \"best\" does not exist"],"errors_with_context":[{"type":"unknown","message":"column \"best\" does not exist"}]}
```

### After map(c => `"${c}"`)

```js
[...]

PR #841, opened 2018-08-14
[WIP] Replace spaces with underscore both in properties and SQL query

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/522

## Description

This PR replaces the spaces found both in the SQL query and in every property name between double quotes with underscore (`_`). This should be transparent to the user.

## Example

From https://github.com/CartoDB/carto-vl/issues/522#issuecomment-412085107
```js
const source = new carto.source.SQL('SELECT *, 99 as "best price" FROM populated_places_simple', {
    apiKey: 'default_public',
    user: 'rochoa'
});

const viz = new carto.Viz(`width: prop('best price')`);
const layer = new carto.Layer('layer', source, viz);
```

## Discussion

In Builder, when a user tries to change a column name with spaces, they are also replaced with underscore. 

If we follow this approach, we should also take into account that property names and aliases with spaces can be declared between:
* single quotes
* double quotes
* backticks

As @Jesus89 mentions [here](https://github.com/CartoDB/carto-vl/issues/522#issuecomment-412455327) about *dimensions*, in this case they take the string already parsed.

I think we need to discuss further a solution.
* Instead of parsing the query and the properties, should we throw an error / warn the user if they contain spaces?
* @Jesus89 suggestion: > The request should be encoded with LZMA to allow valid characters `"`

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -4,6 +4,7 @@ import Metadata from '../renderer/Metadata';
 import schema from '../renderer/schema';
 import Time from '../renderer/viz/expressions/time';
 import * as windshaftFiltering from './windshaft-filtering';
+import { parseQuerySpaces } from '../renderer/viz/expressions/utils.js';
 
 const SAMPLE_ROWS = 1000;
 const MIN_FILTERING = 2000000;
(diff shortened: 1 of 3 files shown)


Jesus89:
I don't think this is a good approach. Does Windhaft match `best_price` to `best price`?

Builder allows to create columns and it replaces automatically spaces with underscore characters. But you can access columns with spaces from the Advanced SQL console in Builder.

So I think we should add "" to the columns in our query and compress LZMA to allow ` and " chars.

PR #848, opened 2018-08-16
Avoid linear NaN

elenatorro:
# Description 

When using a linear expression in a `ramp`, if the linear expression returns a NaN input value, the color returned by the ramp `eval` method is:

`{ r: NaN, g: NaN, b: NaN }`

With this approach we avoid sending NaN numbers if `max` === `min`.

## Discussion

As @davidmanzanares mentions here https://github.com/CartoDB/carto-vl/pull/848#issuecomment-413497114, actually the linear expression should not return a value when `max` === `min`. However, I think we should avoid sending `{ r: NaN, g: NaN, b: NaN }`. in ramp eval. I think the best approach is to:

- Send `null` in linear eval if `max` === `min`
- Send `null` in ramp eval if the input is null.


davidmanzanares:
I'm sorry :( , but I still don't see this kind of change is necessary, the `min === max` condition in `linear` seems like an application bug to me.

Regarding `null` vs `NaN` I prefer `NaN` since GLSL doesn't have the concept of null, but it does have the concept of `NaN`. Emulating `null` in GLSL would mean more code, complexity and effort.

elenatorro:
Ok, I agree we can keep the linear expression as it is :) But about returning `{ r: NaN, g: NaN, b: NaN }`, ) I think we do not need to calculate the color if the `input` is `NaN` (or any invalid value).

davidmanzanares:
> But about returning { r: NaN, g: NaN, b: NaN }, ) I think we do not need to calculate the color if the input is NaN (or any invalid value).

I made a quick test and the current GLSL of ramp fetches the first pixel when the input is `NaN`, to be consistent I would do the same in `eval` because, right now, `eval` is returning `{r: NaN, g: NaN, b: NaN, a: NaN}` in these cases.

elenatorro:
> the current GLSL of ramp fetches the first pixel when the input is NaN

Great! Let's do that then :)

src/renderer/viz/expressions/ramp.js
```diff
@@ -143,7 +142,7 @@ export default class Ramp extends BaseExpression {
     }
 
     _getColorValue (texturePixels, m) {
-        const index = Math.round(m * MAX_BYTE_VALUE);
+        const index = !Number.isFinite(m) || Number.isNaN(m) ? 0 : Math.round(m * MAX_BYTE_VALUE);
```
davidmanzanares: > !Number.isFinite(m) 

I guess that GLSL behavior for `Inf` (positive) is to get the last color, not the first one

elenatorro: Does it mean that for +Infinite we must return the last color, but for -Infinite the first one?

davidmanzanares: I think so, yes

src/renderer/viz/expressions/ramp.js
```diff
@@ -455,3 +454,16 @@ function _calcPaletteValues (palette) {
 
     return palette;
 }
+
+function _calcColorValueIndex (texturePixels, m) {
+    switch (m) {
+        case Number.isNaN(m):
+            return 0;
+        case m === Infinity:
+            return texturePixels.length - 4;
```
davidmanzanares: I think this is buggy, see line `r: Math.round(texturePixels[index * 4 + 0]),` the index is going to be multiplied by 4 and I think it will overflow

elenatorro: OPS you are right! let me change it

src/renderer/viz/expressions/ramp.js
```diff
@@ -145,7 +144,7 @@ export default class Ramp extends BaseExpression {
     }
 
     _getColorValue (texturePixels, m) {
-        const index = Math.round(m * MAX_BYTE_VALUE);
+        const index = _calcColorValueIndex(m);
```
Jesus89: Could you add a unit test to cover this one?

elenatorro: sure! ✨

Issue #779, opened 2018-07-30
ramp with direct category property is broken


davidmanzanares:
@elenatorro  and I discovered that expressions like `ramp($category, prism)` are broken since `ramp` is assuming that `$category` internal IDs are between `0` and `numCategories` and that is only true when there is only one category property on the dataset.

Usage of `ramp` with `buckets` or `top` don't show this problem.

davidmanzanares:
Hi @makella , I don't want to work in "workaround guided" development, but this would be easier to solve if we forced a `top` operation in those expressions, which is a previously discussed idea.

To summarize for other people, the idea is to change (the behavior of) expressions like `ramp($category, prism)` into `ramp(top($category, 16), prism)`.

I was reluctant to this change, but, knowing that this has been broken for so long and that nobody actually noticed it (we saw this problem by looking into the code for unrelated reasons) makes me think that we should change it.

What do you think, @makella ?

makella:
@davidmanzanares yes, totally this is something we had discussed in the past and looking back, maybe we never came to a clear resolution: 

in a couple of old issues:
https://github.com/CartoDB/carto-vl/issues/110#issuecomment-369812476
https://github.com/CartoDB/carto-vl/issues/129#issuecomment-379748385

on slack:
https://cartoteam.slack.com/archives/C9391D1J5/p1529513536000682

These were both before we had talked through the buckets logic. Based on the logic we have there, with this case, for `ramp($category,prism)` would mean:

- if there are less categories than colors, use the available colors in `prism`
- if there are equal categories to colors, use the available colors in `prism`
- if there are more categories than colors available in `prism`, symbolize all categories with an interpolated color scheme

For example, if someone did:

`ramp($category,[yellow,magenta])` we would go through all of the steps above to provide them the proper symbolization.

Now, if we think about Builder logic, for the third option above, we would symbolize the top 10 categories and then everything else with the "other" gray color. Since we have `top` I think that by default, we can stick with the logic we already have built-in.

There is definitely a case to not limit the number of categories that we color as many users actually want to "see" all the categories colored in the data. 

If we go this direction, I should definitely start revisiting the work I had started on reordering the category color schemes.

makella:
Ahhh, I just started reading #781 after I had hit comment on this :)

Questions:
- the limit for `top` is 1024 categories? we give an error at 200 because that is when we start seeing performance issues?

> but changing the implementation to hash the input category ID.

what exactly does this mean?

[...]

PR #851, opened 2018-08-16
Fix ramp for categories

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/779

The first idea was to use hashing, but since it caused some issues (similar colors being assigned in some cases, randomly of course) I moved to this approach of "translating".

This approach is based on translating ID of categories (that identify any category of any property) to values in the [0,1] range that is spread to match the categories of the particular property that is being used on the `ramp` expression.

If we have categories "A", "B", "C",  "Madrid", "London" and "NY" in two properties "type" and "city" and we ramp by city like `ramp($city, bold)` the translating approach will translate the global unique identifiers of "Madrid", "London" and "NY" to values `0`, `0.5` and `1`. These values are used to look up in the color texture. 

Previously:

> This is an important breaking change, we should be careful and use a good hashing function before merging since future changes there will imply new breaking changes.
> 
> - Ramp hash initial implementation
> - Review or rewrite hash/PRNG function
> - Fix ramp with classifiers


src/renderer/viz/expressions/basic/property.js
```diff
@@ -63,8 +63,8 @@ export default class Property extends BaseExpression {
         }
         this.type = metaColumn.type;
 
-        if (this.type === 'category') {
-            this.numCategories = metaColumn.categories.length;
+        if (this.type === 'category' && !this.numCategories) {
+            Object.defineProperty(this, 'numCategories', { get: function () { return metaColumn.categories.length; } });
```
elenatorro: I have also added a method named `getCategories` that I needed in the `ramp.getLegend` PR https://github.com/CartoDB/carto-vl/pull/869 to get the categories of a property.

I think it would be better to have a `categories` getter and then get the length when needed, or have both a `categories` and `numCategories` getters. It would return the respective values only if the property type is a category, otherwise it would return undefined. I think it is a better approach because defining a property in the constructor is harder to test and read. What do you think?

src/renderer/viz/expressions/ramp.js
```diff
@@ -184,16 +185,31 @@ export default class Ramp extends BaseExpression {
             };
         }
 
+        let inline = `texture2D(texRamp${this._uid}, vec2((${input.inline}-keyMin${this._uid})/keyWidth${this._uid}, 0.5))`;
+        if (this.input.type === 'category' && this.input.isA(Property)) {
```
elenatorro: `this.input.type === inputTypes.CATEGORY`

src/renderer/viz/expressions/ramp.js
```diff
@@ -325,6 +342,35 @@ export default class Ramp extends BaseExpression {
         gl.uniform1f(this._getBinding(program).keyMinLoc, (this.minKey));
         gl.uniform1f(this._getBinding(program).keyWidthLoc, (this.maxKey) - (this.minKey));
         drawMetadata.freeTexUnit++;
+
+        if (this.input.type === 'category' && this.input.isA(Property)) {
```
elenatorro: `this.input.type === inputTypes.CATEGORY`

src/renderer/viz/expressions/ramp.js
```diff
@@ -325,6 +342,35 @@ export default class Ramp extends BaseExpression {
...
+            gl.activeTexture(gl.TEXTURE0 + drawMetadata.freeTexUnit);
+            if (this._translatedIds !== this._metadata.properties[this.input.name].categories.length) {
+                this._translatedIds = this._metadata.properties[this.input.name].categories.length;
+                this._translateTexture = gl.createTexture();
+                const translatorPixels = new Float32Array(256 * 256);
```
elenatorro: Instead of 256, we have `COLOR_ARRAY_LENGTH`.

src/renderer/viz/expressions/ramp.js
```diff
@@ -325,6 +342,35 @@ export default class Ramp extends BaseExpression {
...
+                for (let i = 0; i < this._metadata.properties[this.input.name].categories.length; i++) {
+                    const id = this._metadata.categoryToID.get(this._metadata.properties[this.input.name].categories[i].name);
+                    const value = i / (this._metadata.properties[this.input.name].categories.length - 1);
+                    const vec2Id = {
+                        x: id % 256,
+                        y: Math.floor(id / 256)
+                    };
+                    translatorPixels[256 * vec2Id.y + vec2Id.x] = value;
+                }
+                gl.bindTexture(gl.TEXTURE_2D, this._translateTexture);
```
elenatorro: I think we can reuse `_bindGLTexture` method, which also checks the ramp type.

We can add a `texture` parameter to the method:

```js
_bindGLTexture (gl, texturePixels, texture) {
        gl.bindTexture(gl.TEXTURE_2D, texture);
        ...
}
```

And call it here as follows:

```js
this._bindGLTexture(gl.TEXTURE_2D, translatorPixels, this._translateTexture);
```

What do you think?

src/renderer/viz/expressions/ramp.js
```diff
@@ -325,6 +342,35 @@ export default class Ramp extends BaseExpression {
         gl.uniform1f(this._getBinding(program).keyMinLoc, (this.minKey));
         gl.uniform1f(this._getBinding(program).keyWidthLoc, (this.maxKey) - (this.minKey));
         drawMetadata.freeTexUnit++;
+
+        if (this.input.type === 'category' && this.input.isA(Property)) {
+            gl.activeTexture(gl.TEXTURE0 + drawMetadata.freeTexUnit);
+            if (this._translatedIds !== this._metadata.properties[this.input.name].categories.length) {
```
elenatorro: Save `this._metadata.properties[this.input.name].categories.length` to a constant (i.e. `PROPERTY_CATEGORIES_LENGTH`)

src/renderer/viz/expressions/ramp.js
```diff
@@ -325,6 +342,35 @@ export default class Ramp extends BaseExpression {
...
+                this._translatedIds = this._metadata.properties[this.input.name].categories.length;
+                this._translateTexture = gl.createTexture();
+                const translatorPixels = new Float32Array(256 * 256);
+                for (let i = 0; i < this._metadata.properties[this.input.name].categories.length; i++) {
+                    const id = this._metadata.categoryToID.get(this._metadata.properties[this.input.name].categories[i].name);
```
elenatorro: Just as an improvement, I would add a method to the `Metadata` class to get the property categories:

```js
    getPropertyCategories (propertyName) {
        return this._metadata.properties[propertyName].categories;
    }
```

So the loop would look like:

```js
const translatorPixels = new Float32Array(COLOR_ARRAY_LENGTH * COLOR_ARRAY_LENGTH);
const propertyCategories = this.metadata.getPropertyCategories(this.input.name);
const PROPERTY_CATEGORY_LENGTH = propertyCategories.length;

for (let i = 0; i < PROPERTY_CATEGORY_LENGTH; i++) {
    const id = propertyCategories[i].id;
    const value = i / (PROPERTY_CATEGORY_LENGTH - 1);
    const vec2Id = {
        x: id % COLOR_ARRAY_LENGTH,
        y: Math.floor(id / COLOR_ARRAY_LENGTH)
    };
    translatorPixels[COLOR_ARRAY_LENGTH * vec2Id.y + vec2Id.x] = value;
}
```

src/renderer/viz/expressions/ramp.js
```diff
@@ -184,16 +185,31 @@ export default class Ramp extends BaseExpression {
...
                 uniform sampler2D texRamp${this._uid};
+                uniform sampler2D texRampTranslate${this._uid};
                 uniform float keyMin${this._uid};
-                uniform float keyWidth${this._uid};`
+                uniform float keyWidth${this._uid};
+
+                float ramp_translate${this._uid}(float s){
+                    vec2 v;
+                    v.y = floor(s/256.);
+                    v.x = s - v.y*256.;
```
elenatorro: `v.x = mod(s, 256)` ?

src/renderer/Metadata.js
```diff
@@ -19,18 +19,24 @@ export default class Metadata {
...
         if (category === undefined) {
             category = null;
         }
         if (this.categoryToID.has(category)) {
             return this.categoryToID.get(category);
         }
+        if (!init) {
+            this.properties[propertyName].categories.push({
+                name: category,
+                frequency: Number.NaN
```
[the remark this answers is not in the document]
davidmanzanares: For correctness, frequency is the total frequency in the entire dataset, this code is executed when a global value is unknown but we just got one of these categories in a tile.

src/renderer/viz/expressions/basic/property.js
davidmanzanares: Seems like a good idea, but I would do that on your PR :slightly_smiling_face:

src/renderer/viz/expressions/ramp.js
davidmanzanares: May I disagree? :joy: I think this is clearer and more concise

src/renderer/viz/expressions/ramp.js
davidmanzanares: Ups, this constant should be extracted and named, although it is not related to that one. It is just the maximum number of categories per property.

src/renderer/viz/expressions/ramp.js
elenatorro: it was a matter to be consistent with the rest of the file, but we can change this in a future refactor, so no problem 👍

src/renderer/viz/expressions/ramp.js
elenatorro: I did not know it was faster, so let's keep it like this :)

src/renderer/viz/expressions/ramp.js
elenatorro: We are currently using `this._bindGLTexture(gl, texturePixels)` in `_computeGLTextureIfNeeded` method. We can call it there like this:

* `this._bindGLTexture(gl, texturePixels, this.texture)`

And here:

* `this._bindGLTexture(gl, translatorPixels, this._translateTexture)`

davidmanzanares: :confused: I don't understand, could you expand the implementation of `_bindGLTexture`?

elenatorro: Yep, sorry!

```js
    _bindGLTexture (gl, texturePixels, texture) {
        gl.bindTexture(gl.TEXTURE_2D, texture);

        if (this.type === rampTypes.COLOR) {
            gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, COLOR_ARRAY_LENGTH, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, texturePixels);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
        } else {
            gl.texImage2D(gl.TEXTURE_2D, 0, gl.ALPHA, COLOR_ARRAY_LENGTH, 1, 0, gl.ALPHA, gl.FLOAT, texturePixels);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
        }

        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
    }
```

So, instead of 

```js
      gl.bindTexture(gl.TEXTURE_2D, this._translateTexture);
      gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
      gl.texImage2D(gl.TEXTURE_2D, 0, gl.ALPHA, 256, 256, 0, gl.ALPHA, gl.FLOAT, translatorPixels);
      gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
      gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
      gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
      gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
``` 
We can call:

```js
this._bindGLTexture(gl, translatorPixels, this._translateTexture)
```

davidmanzanares:
@elenatorro could you re-review this? I made some changes to address some of the points.

elenatorro:
LGTM :)

Issue #820, opened 2018-08-07
Linux and Nvidia drivers display larger points


davidmanzanares:
~High-dpi screens show different result compared to regular screens. With points at least, every geometry should be checked~

The problem can be seen here: https://cartodb.github.io/carto-vl/examples/interactivity/interactive-based-styling.html

makella:
@davidmanzanares even though I couldn't see the large sizes with the maps we were looking at, I can see them on this map:
https://zingbot.github.io/vectoria/locations_explore.html

davidmanzanares:
Fixed by #854

PR #854, opened 2018-08-17
Control NaN widths to avoid "big points" problems

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/820

### DIFF
--- a/src/renderer/shaders/styler/stylerEncodeWidth.glsl
+++ b/src/renderer/shaders/styler/stylerEncodeWidth.glsl
@@ -1,4 +1,10 @@
 vec2 encodeWidth(float x) {
+    if (!(x<=0. || x>=0.) ){
+        // Convert NaN widths to `0`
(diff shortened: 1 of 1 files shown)


src/renderer/shaders/styler/stylerEncodeWidth.glsl
```diff
@@ -1,4 +1,10 @@
 vec2 encodeWidth(float x) {
+    if (!(x<=0. || x>=0.) ){
```
Jesus89: Is `isnan` not always available?

davidmanzanares: In WebGL 1 is not available ever, in modern GLSL it is: https://stackoverflow.com/questions/11810158/how-to-deal-with-nan-or-inf-in-opengl-es-2-0-shaders

Issue #448, opened 2018-05-22
Check that all the reference items follow the same example structure


arianaescobar:
With the code blocks on the right instead of the left. 

For example, [buckets](https://carto.com/developers/carto-vl/reference/#cartoexpressionsbuckets) look like it's not following it properly.

davidmanzanares:
The reason for that inconsistency is that the examples are interleaved with the explanation, and the explanation refers and explain the examples. Maybe it is explained in too much detail, though.

rochoa:
@arianaescobar, can you give us more feedback on this one?

arianaescobar:
My comment was regarding adding code blocks to the left side of the reference. We have an assigned space on the right that it's only for code and we should leverage that more. 

Like here
![screen shot 2018-06-11 at 10 22 27](https://user-images.githubusercontent.com/16247628/41220398-9064e8a0-6d61-11e8-8d7a-c4ea6e3c422c.png)

or here
![screen shot 2018-06-11 at 10 24 34](https://user-images.githubusercontent.com/16247628/41220428-ac065c60-6d61-11e8-9faa-f82a22fc1eb2.png)

I think the level of detail is fine @davidmanzanares, I just think we should use the right side of the layout more. 

Like we did here
![screen shot 2018-06-11 at 10 25 59](https://user-images.githubusercontent.com/16247628/41220537-017ec7f4-6d62-11e8-8f93-bf6537423674.png)

or here
![screen shot 2018-06-11 at 10 26 57](https://user-images.githubusercontent.com/16247628/41220543-04c72dac-6d62-11e8-8d00-b9a101e12e04.png)

PR #855, opened 2018-08-17
Fix reference examples

Jesus89:
Fix for https://github.com/CartoDB/carto-vl/issues/448.

I have noticed that /@ appears in the docs instead of @, but this should be fixed in the Developer Center.

### DIFF
--- a/src/renderer/viz/expressions/basic/variable.js
+++ b/src/renderer/viz/expressions/basic/variable.js
@@ -13,13 +13,13 @@ import { checkString } from '../utils';
  *   variables: {
  *     sum_price: s.clusterSum(s.prop('price'))
  *   }
- *  filter: s.neq(s.var('sum_price'), 'london'),
+ *  filter: s.neq(s.var('sum_price'), 'london')
  * });
  *
  * @example <caption>(String)</caption>
  * const viz = new carto.Viz(`
- *   @sum_price: clusterSum($price)
- *   filter: @sum_price !== 'london'
+ *   \@sum_price: clusterSum($price)
+ *   filter: $price == 30  // Equivalent to eq($price, 30)
  * `);
  *
  * @memberof carto.expressions
--- a/src/renderer/viz/expressions/buckets.js
+++ b/src/renderer/viz/expressions/buckets.js
@@ -4,60 +4,39 @@ import { implicitCast, getOrdinalFromIndex } from './utils';
 /**
  * Given a property create "sub-groups" based on the given breakpoints.
  *
- * Imagine a traffic dataset with a speed property. We want to divide the roads in
- * 3 buckets (slow, medium, fast) based on the speed using a different color each bucket.
- *
- * We´ll need:
- *  - A {@link carto.expressions.ramp|ramp} to add a color for every bucket.
- *  - A {@link carto.expressions.palettes|colorPalette} to define de color scheme.
- *
- * ```javascript
- *  const s = carto.expressions;
- *  const $speed = s.prop('speed');
- *  const viz = new carto.Viz({
- *    color: s.ramp(
- *      s.buckets($speed, [30, 80, 120]),
- *      s.palettes.PRISM
- *    )
- * });
- * ```
- *
- * ```javascript
- *  const viz = new carto.Viz(`
- *    color: ramp(buckets($speed, [30, 80, 120]), PRISM)
- * `);
- * ```
- *
- * Using the buckets `expression` we divide the dataset in 3 buckets according to the speed:
- *  - From 0 to 29
- *  - From 30 to 79
- *  - From 80 to 120
- *
- * Values lower than 0 will be in the first bucket and values higher than 120 will be in the third one.
- *
- * This expression can be used for categorical properties, imagine the previous example with the data already
- * procesed in a new categorical `procesedSpeed` column:
- *
- * ```javascript
- *  const s = carto.expressions;
- *  const $procesedSpeed = s.prop('procesedSpeed');
- *  const viz = new carto.Viz({
- *    color: s.ramp(
- *      s.buckets($procesedSpeed, ['slow', 'medium', 'high']),
- *      s.palettes.PRISM)
- *    )
- * });
- * ```
- *
- * ```javascript
- *  const viz = new carto.Viz(`
- *    color: ramp(buckets($procesedSpeed, ['slow', 'medium', 'high']), PRISM)
- * `);
- * ```
+ * This returns a number or category expression depending on the input values.
  *
  * @param {Number|Category} property - The property to be evaluated and interpolated
  * @param {Number[]|Category[]} breakpoints - Numeric expression containing the different breakpoints.
  * @return {Number|Category}
+ *
+ * @example <caption>Display a traffic dataset is 3 colors depending on the numeric speed.</caption>
+ * const s = carto.expressions;
+ * const viz = new carto.Viz({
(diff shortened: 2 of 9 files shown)


src/renderer/viz/expressions/basic/variable.js
```diff
@@ -13,13 +13,13 @@ import { checkString } from '../utils';
...
  *   }
- *  filter: s.neq(s.var('sum_price'), 'london'),
+ *  filter: s.neq(s.var('sum_price'), 'london')
  * });
  *
  * @example <caption>(String)</caption>
  * const viz = new carto.Viz(`
- *   @sum_price: clusterSum($price)
- *   filter: @sum_price !== 'london'
+ *   \@sum_price: clusterSum($price)
```
davidmanzanares: Why do we need the `\`?

src/renderer/viz/expressions/buckets.js
```diff
@@ -4,60 +4,39 @@ import { implicitCast, getOrdinalFromIndex } from './utils';
...
  *
- * Imagine a traffic dataset with a speed property. We want to divide the roads in
- * 3 buckets (slow, medium, fast) based on the speed using a different color each bucket.
+ * This returns a number or category expression depending on the input values.
  *
- * We´ll need:
- *  - A {@link carto.expressions.ramp|ramp} to add a color for every bucket.
- *  - A {@link carto.expressions.palettes|colorPalette} to define de color scheme.
+ * @param {Number|Category} property - The property to be evaluated and interpolated
+ * @param {Number[]|Category[]} breakpoints - Numeric expression containing the different breakpoints.
```
davidmanzanares: It's not numeric when it is a category

src/renderer/viz/expressions/buckets.js
```diff
@@ -4,60 +4,39 @@ import { implicitCast, getOrdinalFromIndex } from './utils';
...
  *
- * ```javascript
- *  const viz = new carto.Viz(`
+ * @example <caption>Display a traffic dataset is 3 colors depending on the numeric speed. (String)</caption>
+ * const viz = new carto.Viz(`
  *    color: ramp(buckets($speed, [30, 80, 120]), PRISM)
  * `);
- * ```
- *
- * Using the buckets `expression` we divide the dataset in 3 buckets according to the speed:
```
davidmanzanares: I think this clarification is useful, although it contains some errors, could you keep it but fixing it?

src/renderer/viz/expressions/placement.js
```diff
@@ -7,14 +7,6 @@ import { checkLooseType, checkType, implicitCast } from './utils';
  * - `symbolPlacement: placement(0,0)` means to align the center of the image with the point center.
  * - `symbolPlacement: placement(-1,-1)` means to align the top right corner of the image with the point center.
  *
- *           |1
- *           |
- *           |
- * -1 -------+------- 1
```
davidmanzanares: IMO, this is useful

src/renderer/viz/expressions/unary.js
```diff
@@ -176,6 +176,17 @@ export const Abs = genUnaryOp('abs', x => Math.abs(x), x => `abs(${x})`);
  *
  * This returns a numeric expression where 0 means `false` and 1 means `true`.
  *
+ * @example <caption>isNaN.</caption>
```
davidmanzanares: This needs to explain what the filter is doing

src/renderer/viz/expressions/basic/variable.js
Jesus89: Because without \, @ is not parsed by JSDoc
selection_256

src/renderer/viz/expressions/placement.js
```diff
@@ -7,14 +7,6 @@ import { checkLooseType, checkType, implicitCast } from './utils';
...
- *           |
- *           |
- *         -1|
- *
```
elenatorro: Can we keep this? 🙏

src/renderer/viz/expressions/buckets.js
Jesus89: The examples format is one line description and the example. So we can not add a multiline description in the inline example. This extra information could be added in the guides. However, we can extend the description a little bit. What about:

`Display a traffic dataset is 3 colors depending on the numeric speed: (0, 29), (30, 79), (80, 120).`

We could add this in the documentation description:

`Values lower than the minimum will be in the first bucket and values higher than the maximum will be in the last one.`

src/renderer/viz/expressions/placement.js
Jesus89: This ASCII picture generates very bad results:

selection_257

src/renderer/viz/expressions/placement.js
Jesus89: https://github.com/CartoDB/carto-vl/pull/855#discussion_r210871209

src/renderer/viz/expressions/placement.js
davidmanzanares: Can't we fix that instead of removing it?

src/renderer/viz/expressions/unary.js
Jesus89: It seems that the general rule is to use just the name for all the examples in unary expressions. However, I agree with improving the example's description.

src/renderer/viz/expressions/placement.js
elenatorro: +1 to fix it 🙏

Jesus89: :+1: I move it inside a code block

src/renderer/viz/expressions/buckets.js
Jesus89: Finally, I have included the info inside the example with comments.

davidmanzanares: Did you push the commits? I don't see it :confused:

Jesus89: https://github.com/CartoDB/carto-vl/pull/855/commits/1aad7df8f548d310e8748a2ddfe74cb41cd8ba9a

src/renderer/viz/expressions/buckets.js
```diff
@@ -4,60 +4,49 @@ import { implicitCast, getOrdinalFromIndex } from './utils';
...
- * ```
  *
- * ```javascript
- *  const viz = new carto.Viz(`
+ * @example <caption>Display a traffic dataset is 3 colors depending on the numeric speed. (String)</caption>
+ * // Using the buckets `expression` we divide the dataset into 3 buckets according to the speed
+ * // - From 0 to 29
+ * // - From 30 to 79
+ * // - From 80 to 120
+ * // Values lower than 0 will be in the first bucket and values higher than 120 will be in the third one.
```
davidmanzanares: the error I was referring is this `third`, it should be `last` or `fourth`

Jesus89: Sorry but I don't get why `third` wrong. The buckets are (-inf, 29], [30, 79], [80, inf) right?

davidmanzanares: Nop, the buckets are (-inf, 29], [30, 79], [80, 120), [120, inf)

Jesus89: Oh, didn't know that. So the clarification is required.

Jesus89: (Updated)

Issue #675, opened 2018-07-10
Throw error when adding extra arguments


davidmanzanares:
I've seen a few times now the usage of extra parameters in the expressions. The current behavior is to ignore them, but I think that throwing an error would be better.

davidmanzanares:
Fixed by #859

PR #859, opened 2018-08-17
Throw error when an expression has extra arguments

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/675

# Description

This PR adds a check function that throws an error if the number of arguments passed to an expression exceeds the max number of arguments needed.

## TODO

* [x] Add check function for every expression
* [x] Test all the expressions

### DIFF
--- a/src/renderer/viz/expressions/utils.js
+++ b/src/renderer/viz/expressions/utils.js
@@ -3,6 +3,13 @@ import BaseExpression from './base';
 
 export const DEFAULT = undefined;
 
+export function checkMaxArguments (constructorArguments, maxArguments, expressionName) {
+    if (constructorArguments.length > maxArguments) {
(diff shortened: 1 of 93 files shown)


src/renderer/viz/expressions/utils.js
```diff
@@ -3,6 +3,12 @@ import BaseExpression from './base';
 
 export const DEFAULT = undefined;
 
+export function checkMaxArguments (constructorArguments, maxArguments, expressionName) {
+    if (constructorArguments.length > maxArguments) {
+        throw new Error(`Expression ${expressionName} accepts ${maxArguments} arguments, but ${constructorArguments.length} where passed.`);
```
davidmanzanares: typo: `where` => `were`

Issue #733, opened 2018-07-18
Proposal for zoomRange to drive multi-scale property interpolation 


makella:
@davidmanzanares and I were discussing multi-scale symbology and how to make it easier for the user.

After going through a variety of options, we have a proposal for an expression `zoomRange`.

If we can make it a separate expression, we will have more control over how it behaves. Right now, when a user wants to do multi-scale styling, it requires a series of `blends` and `linear` interpolation.

In addition, `linear` only accepts 2 arguments. So if you want to add more than 2 "stops", you have to nest a series of `blend`s

For example if someone wants to have marker width "stops" where a symbol size is defined for particular zoom levels and the non-defined zoom levels interpolated between them, they would have to do the following:

```
width: blend(2,
             blend(8, 12, linear(zoom(), 2^14, 2^16))
             linear(zoom(),2^10,2^14)
)
```
With this proposal, the above would become:

`width: ramp(zoomRange(10,14,16), [2,8,12])`

Where at zooms <= z10 the symbol size would be 2, at z14 8, and at >=z16, 12 and the in-betweens (z11 - z13 and z15 would be interpolated).

We could also apply the same logic to other styling properties like `color`:

`color: ramp(zoomRange(7,10,14),[red,orange,yellow]))`

davidmanzanares:
:heart: 

Another possible feature is the combination of ramps (nesting):
```
ramp(linear($population, 1,10000),
          ramp(zoomRange(10,14,16),
                   [1, 10],
                   [2, 20],
                    [4,40]                    
           )
)  

```

makella:
YES!

davidmanzanares:
This was added by #860

PR #860, opened 2018-08-17
Zoomrange

davidmanzanares:
Adds zoomrange as discussed https://github.com/CartoDB/carto-vl/issues/733

This is part of https://github.com/CartoDB/carto-vl/issues/690

Superseeds #745


src/renderer/viz/expressions/Zoomrange.js
```diff
@@ -0,0 +1,62 @@
...
+    constructor (zoomBreakpointList) {
+        zoomBreakpointList = implicitCast(zoomBreakpointList);
+
+        super({});
+        this.zoomBreakpointList = zoomBreakpointList;
+        this.type = 'number';
+        this.inlineMaker = inline => inline._impostor;
+    }
+    eval () {
+        return this._impostor.eval();
```
elenatorro: Instead of impostor, can name it `_zoomrange`?

src/renderer/viz/expressions/Zoomrange.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+    _bindMetadata (metadata) {
+        checkType('zoomrange', 'zoomBreakpointList', 0, 'number-array', this.zoomBreakpointList);
+        if (this.zoomBreakpointList.elems.length < 2) {
+            setTimeout(() => {
```
elenatorro: I guess we are doing so because we need to launch this error immediately, but I do not understand why.

elenatorro:
Now that https://github.com/CartoDB/carto-vl/pull/859 is merged, we have to check for extra arguments 😅

src/renderer/viz/expressions/Zoomrange.js
davidmanzanares: What are your reasons? I prefer `_impostor`, I think it has better semantics for this case

src/renderer/viz/expressions/Zoomrange.js
davidmanzanares: Actually, I checked it and it's not needed, my bad

src/renderer/viz/expressions/Zoomrange.js
elenatorro: It is true that we use `_impostor` in other expressions such as `viewportPercentile`, but then in `now` expression we use `now`, and in my opinion the second sounds better (but I might be missing something). Anyway, it is just a question.

elenatorro: Since `_impostor.eval()` will call `eval` method in `blend` expression, should not we pass the feature?

davidmanzanares: Yes! I totally forgot!

elenatorro:
I have been thinking about the `scale` nomenclature because, as we have talked it is not _exactly_ the scale ratio. The current implementation of the scale expression is width * pixels per tile / number of tiles (please correct me if i am wrong). So, in the `drawMetadata` object I would name `scale` => `pixelsPerTile`, and instead of `scale` I would say `scalePixelsRatio` or `pixelsRatio`. What do you think? (I am still thinking about it)

elenatorro:
> `width * pixels per tile / number of tiles` 

I meant in the expression: (`div(mul(width, 0), pow(2, zoomlevel))`) 

About the `magnification`, do you mean to use it as the expression name? I think it describes the behaviour better, but still sounds weird to me.

List of proposals:

```
width: magnification(25, 7)
```

```
width: extent(25, 7)
```

```
width: adjustment(25, 7)
```

```
width: proportion(25, 7)
```

davidmanzanares:
Sorry, I thought you were talking about `drawMetadata.scale`.

Regarding the expression name, I'm not sure, I would like to know what @makella and @jgoizueta think about this.

jgoizueta:
Regarding the scale expression name, I'd use a verb since it *scales* its argument. So `scale` is confusing because it can be interpreted as a noun... We could try to find an unambiguous name like `magnify` (but I don't like that one because it seems to mean *enlarge* to me). From @elenatorro list, `adjust` could also be an unambiguous verb.

Any other verb suggestions? `adjustScale`?  `scaleToTerrain`?

makella:
hmmmm reading through, what about `zoomRatio` since we will use `zoomRange` for logarithmic?

just the first thought after reading through everything again.

elenatorro:
I really like `zoomRatio`!

davidmanzanares:
:(

I don't like `zoomRatio` much to be honest. I agree with @jgoizueta in that it would be better if it was a verb and, also, I think that ratio here is quite ambiguous :S

makella:
@davidmanzanares fair enough! i feel like some of the verb descriptions are almost like sentences (haha) but maybe we can see some examples?

elenatorro:
Following @jgoizueta suggestions, maybe `adjustZoom` or even `scaleZoom`? (so we have the zoom word as well which makes the expression clearer)

makella:
@davidmanzanares explained the `scale()` behavior to me in greater detail and now that I have a better understanding, it is very similar to what ArcGIS terms ["reference scale"](http://desktop.arcgis.com/en/arcmap/10.3/map/working-with-arcmap/working-with-data-frame-reference-scales.htm) basically, this is the scale at which I am defining my symbology and as I zoom in and out I want my symbols to stay fixed proportionate to that reference as I zoom in and out of my map.

I wanted to add this here just in case it sparks some more ideas for names.

`setReference`
`scaleReference`

....

davidmanzanares:
I like `scaled`, but in any case we should merge this soon since `zoomrange` can improve things with some of the apps that we are already working/helping on. If there is no consensus about `scale` naming I can make it private so that it doesn't appear on the docs and merge this.

elenatorro:
I agree, we can merge this with `scale` or `scaled` (I would go for the second) and discuss later a better naming.

Issue #853, opened 2018-08-17
Fix interactivity for animated layers


Jesus89:
Events `featureHover`, `featureEnter` and `featureLeave` are only triggered when the mouse moves, but no when the feature changes in an animated layer.

For example in this map: https://cartodb.github.io/CARTOgraphy/vl/active-fires.html popup should change with the animation so `featureHover` should be called automatically when required.

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/865

PR #865, opened 2018-08-17
Fix interactivity when layer is updated

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/853

Events `featureHover`, `featureEnter`, `featureLeave` are fired when the animated feature changes at the current mouse position.

animated-featureevents

### DIFF
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -64,7 +64,6 @@ export default class Layer {
         this._checkViz(viz);
         this._oldDataframes = new Set();
         this._isInitialized = false;
-        this._isAnimated = false;
         this._init(id, source, viz);
     }
 
@@ -173,8 +172,6 @@ export default class Layer {
     async update (source, viz) {
         this._checkSource(source);
         this._checkViz(viz);
-
-        this._isAnimated = viz.isAnimated();
 
         source = source._clone();
         this._atomicChangeUID = this._atomicChangeUID + 1 || 1;
@@ -337,7 +334,7 @@ export default class Layer {
     }
 
     isAnimated () {
-        return this._isAnimated;
+        return this._viz && this._viz.isAnimated();
     }
 
     /**
@@ -472,8 +469,6 @@ export default class Layer {
         if (!this._source) {
             throw new Error('A source is required before changing the viz');
         }
-
-        this._isAnimated = viz.isAnimated();
 
         const source = this._source;
         const loadImagesPromise = viz.loadImages();
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -64,6 +64,7 @@ export default class Layer {
         this._checkViz(viz);
         this._oldDataframes = new Set();
         this._isInitialized = false;
+        this._isAnimated = false;
         this._init(id, source, viz);
     }
 
@@ -172,6 +173,9 @@ export default class Layer {
     async update (source, viz) {
         this._checkSource(source);
         this._checkViz(viz);
+
+        this._isAnimated = viz.isAnimated();
+
         source = source._clone();
         this._atomicChangeUID = this._atomicChangeUID + 1 || 1;
         const uid = this._atomicChangeUID;
@@ -332,6 +336,10 @@ export default class Layer {
             : [];
     }
 
+    isAnimated () {
+        return this._isAnimated;
+    }
+
     /**
      * Change layer visibility to visible
      *
@@ -464,6 +472,9 @@ export default class Layer {
         if (!this._source) {
             throw new Error('A source is required before changing the viz');
         }
+
+        this._isAnimated = viz.isAnimated();
+
         const source = this._source;
         const loadImagesPromise = viz.loadImages();
         const metadata = await source.requestMetadata(viz);
--- a/src/interactivity/Interactivity.js
+++ b/src/interactivity/Interactivity.js
@@ -2,6 +2,7 @@ import mitt from 'mitt';
 import Layer from '../Layer';
 import { WM_R, projectToWebMercator } from '../utils/util';
 import { wToR } from '../client/rsys';
+import * as _ from 'lodash';
 
 /**
  *
@@ -65,8 +66,6 @@ const EVENTS = [
     'featureHover',
     'featureLeave'
 ];
-
-const BATCH_UPDATE_CALLS = 10;
 
 export default class Interactivity {
     /**
@@ -173,9 +172,12 @@ export default class Interactivity {
 
     _subscribeToLayerEvents (layers) {
         layers.forEach(layer => {
-            layer.on('updated', () => {
-                this._onLayerUpdated(layer.isAnimated());
-            });
+            // Debounce `updated` event calls
+            // - Delay: 60 ms
+            // - Maximum delay: 100 ms
+            layer.on('updated', _.debounce(() => {
+                this._onLayerUpdated(layer);
+            }, 60, { maxWait: 100 }));
         });
     }
 
@@ -184,14 +186,9 @@ export default class Interactivity {
         integrator.on('click', this._onClick.bind(this));
     }
 
-    _onLayerUpdated (isAnimated) {
-        if (isAnimated) {
-            this._updateCalls++;
-            // Debounce calls to fire the events
-            if (this._updateCalls > BATCH_UPDATE_CALLS) {
-                this._updateCalls = 0;
-                this._onMouseMove(this._mouseEvent);
-            }
+    _onLayerUpdated (layer) {
+        if (layer.isAnimated()) {
+            this._onMouseMove(this._mouseEvent);
         }
     }
 
--- a/src/interactivity/Interactivity.js
+++ b/src/interactivity/Interactivity.js
@@ -175,9 +175,7 @@ export default class Interactivity {
             // Debounce `updated` event calls
             // - Delay: 60 ms
             // - Maximum delay: 100 ms
-            layer.on('updated', _.debounce(() => {
-                this._onLayerUpdated(layer.isAnimated());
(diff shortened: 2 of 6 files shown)


src/interactivity/Interactivity.js
```diff
@@ -167,13 +171,39 @@ export default class Interactivity {
...
     _subscribeToIntegratorEvents (integrator) {
         integrator.on('mousemove', this._onMouseMove.bind(this));
         integrator.on('click', this._onClick.bind(this));
     }
 
+    _onLayerUpdated (isAnimated) {
+        if (isAnimated) {
+            this._updateCalls++;
+            // Debounce calls to fire the events
+            if (this._updateCalls > BATCH_UPDATE_CALLS) {
```
elenatorro: I am not sure if we should use a constant here, I am thinking of having the count of steps the animation has.

src/Layer.js
```diff
@@ -332,6 +336,10 @@ export default class Layer {
             : [];
     }
 
+    isAnimated () {
+        return this._isAnimated;
```
elenatorro: I am not sure of this, so this is more a question: I think isAnimated method should be a getter:

```js
get isAnimated () {
    return this._viz.isAnimated()
}
```

We will not need then `_isAnimated` property.

src/interactivity/Interactivity.js
Jesus89: Yes, I'm not very happy with the constant too. Maybe I need to implement a time debounce here.

elenatorro: We could something like this:

* If an expression is animated, we should be able to know the animation progress from the expression itself (we will need to check the children recursively). So we can have a `getProgressValue` in the base expression that, if it is animated, calls `getProgressValue` of its children until it gets to the animated value (does it makes sense?)
* Then, in `_onLayerUpdated` method, we could check if the progress has changed in order to trigger the event.

src/Layer.js
Jesus89: Yes. I see that I added the flag because of the first implementation (in order to subscribe only to animated layers) but with the final implementation, the evaluation of the animated layer is in the updated event callback so the flag is not needed anymore.

src/interactivity/Interactivity.js
Jesus89: I think we need to filter first the `updated` event callbacks and then verify if there is a change in the animation.

There are ~60 calls of `updated` per second that we need to debounce, also we can have more than one animated layer and even more than one animated style property.

src/interactivity/Interactivity.js
```diff
@@ -167,13 +171,39 @@ export default class Interactivity {
         });
     }
 
+    _subscribeToLayerEvents (layers) {
+        layers.forEach(layer => {
+            layer.on('updated', () => {
+                this._onLayerUpdated(layer.isAnimated());
```
davidmanzanares: Instead of doing this and the debouncing in one way or another I would create a promise here (checking first that the promise doesn't exist to call `on_mouseMove` just once per frame. Something like:
```
layer.on('updated', ()=>{
    if (!this._layerUpdateTrigger){
        this._layerUpdateTrigger = new Promise(resolve => {
              this._onMouseMove(this._mouseEvent);
              this._layerUpdateTrigger = null;
        });
   }
});

File: src/interactivity/Interactivity.js (hunk 1 above)
Comment (davidmanzanares) [reply src/interactivity/Interactivity.js]: Check my other comment related to this, pls. Keep in mind that `Animation` is not the only source of "animatedness", so you cannot expect `getProgressValue` to exist somewhere. Moreover, you can mix multiple `Animation`s.

File: test/integration/user/test/interactivity.test.js [hunk 4]
```diff
@@ -197,89 +197,114 @@ describe('Interactivity', () => {
...
+        describe('and the mouse enters in a feature', () => {
+            it('should fire a featureHover event with a features list containing the entered feature', done => {
+                interactivity.on('featureHover', event => {
+                    expect(event.features[0].id).toEqual(-0);
+                    expect(event.features[0].layerId).toEqual('layer1');
+                    done();
+                });
+                onLoaded(() => {
+                    // Move mouse inside a feature 1
+                    util.simulateMove({ lng: 5, lat: 5 });
```
davidmanzanares: Maybe I'm not seeing something, but this is not testing the new functionality, right? Did we miss those tests before, right?

test/integration/user/test/interactivity.test.js
```diff
@@ -197,89 +197,114 @@ describe('Interactivity', () => {
...
+                    }, 0);
+                });
+            });
+        });
 
-        describe('when the user move the mouse on the map', () => {
-            describe('and the mouse enters in a feature', () => {
-                it('should fire a featureHover event with a features list containing the entered feature', done => {
+        describe('and the mouse leaves a feature', () => {
+            it('should fire a featureHover event with an empty features list', done => {
```
davidmanzanares: Why are we rewriting these tests?

src/interactivity/Interactivity.js
Jesus89: I have made a test computing an array of the `getProgressValue` from a viz object

```js
return this._getRootExpressions().filter(expr => expr.getProgressValue).map(expr => expr.getProgressValue());
```

The problem is that it is difficult to change if the progress has changed with multiple style properties animated a multiple layers and it adds significant extra computing time and the point is that the progress always changes for each iteration so I think it can not be used to filter the triggers.

I have tested also the approach of debouncing the `updated` function call with a `max wait` of 100ms, so there will be always 10 events per second if we have one or more animated layers and one or more animated properties in each layer.

The second step is to filter `featureEnter`, `featureLeave` and `featureHover` events. Since we are computing the differences  between featuresEntered and featuresLeft to trigger `featureEnter` and `featureLeave` events, we can use these lists to know if there is any change in the features when the function is called from _onLayerUpdated. So when the mouse moves `featureHover` is always triggered because of coordinates and position change, and when the mouse stays and there is an `updated` debounced event from an animated layer, `featureHover` only triggers if features change.

elenatorro: > approach of debouncing the updated function call with a max wait of 100ms, so there will be always 10 events per second

Sorry I think I missed something here and I do not understand this approach very well :( My questions are:
* Why should we debounce the updated function?
* Why 10 events per second?

> we can use these lists to know if there is any change in the features when the function is called from _onLayerUpdated

I like this approach! ✨

test/integration/user/test/interactivity.test.js
Jesus89: I realized that some tests where wrong placed. The `describe('when the user moves the mouse on the map'` was inside `describe('when the user clicks on the map'`. I simply moved the blocks to the correct place and added the `describe('when the layer changes'` section.

test/integration/user/test/interactivity.test.js
Jesus89: https://github.com/CartoDB/carto-vl/pull/865#discussion_r211219967

src/interactivity/Interactivity.js
Jesus89: I think this can be implemented just with a flag instead of a Promise. However, I'm a bit worried about the number of events fired when there are lots of animated layers. With the debounce, we can filter that amount of calls and improve performance.

Jesus89:
Updated code algorithm and tests.

NOTE: I found that when filter < 0.5 the features are not detected. I don't know if this is a feature, but I think we should change this to detect features for filter > 0.

cc @davidmanzanares @elenatorro

davidmanzanares:
> NOTE: I found that when filter < 0.5 the features are not detected. I don't know if this is a feature, but I think we should change this to detect features for filter > 0.

It is expected behavior, changing it to `> 0.0` could be problematic with `animation`.

src/interactivity/Interactivity.js
```diff
@@ -167,13 +169,32 @@ export default class Interactivity {
         });
     }
 
+    _subscribeToLayerEvents (layers) {
+        layers.forEach(layer => {
+            // Debounce `updated` event calls
+            // - Delay: 60 ms
+            // - Maximum delay: 100 ms
+            layer.on('updated', _.debounce(this._onLayerUpdated.bind(this), 60, { maxWait: 100 }));
```
davidmanzanares: As I wrote in the outdated comment a debounce is not necessary, in fact, I think we shouldn't use one. I proposed to throw an event at a maximum rate of one per frame, (coalescing all the `updated` events of the same frame).

I don't think we should use a debounce because it's the application code the one who should be responsible for throttling to avoid performance issues related to intensive processing in response to the event. Some applications will need major throttling but others will need less or none at all. We are also having this behavior on the `updated` event which is part of the public API and is not throttled.

Jesus89: OK, you've convinced me. Let's trigger at the maximum rate and let the users optimize the performance if required.

Jesus89: Done

src/interactivity/Interactivity.js
```diff
@@ -200,14 +217,19 @@ export default class Interactivity {
...
         }
 
-        this._prevHoverFeatures = featureEvent.features;
+        this._prevHoverFeatures = currentFeatures;
 
-        // Launch hover event
-        this._fireEvent('featureHover', featureEvent);
+        // If the event comes from a real mouse move, trigger always (because coordinates and position have changed)
+        // If the event comes from an animated event, trigger only when features have changed (because position is the same)
+        if (!force || (force && (featuresLeft.length || featuresEntered.length))) {
```
davidmanzanares: I would change the name of `force` since it gives me the opposite meaning to me. I think `virtual` or `emulated` is better.

Issue #572, opened 2018-06-18
Add a ramp.getLegend method


davidmanzanares:
By implementing ramp.eval we'll make legend-making much easier. However, I think we make it even easier if we provide some special method for the ramp class that returns, in the case of categories, the list of colors and their respective categories, and in the case of a linear ramp a list with the 1024 interpolated colors.

~Blocked for https://github.com/CartoDB/carto-vl/issues/570~

Captain-Oski:
Following this feature-request tightly.

elenatorro:
Blocked by https://github.com/CartoDB/carto-vl/issues/928

elenatorro:
Closing in favor of #935

PR #869, opened 2018-08-20
Add get legend method to ramp expression

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/572

This PR adds the `getLegend()` method to the ramp expression. There is an example added (http://127.0.0.1:8080/examples/misc/legends.html), but we need a better example (cc @makella ❤️).

* Code
```js
const s = carto.expressions;
const viz = new carto.Viz({
  color: s.ramp(s.prop('vehicles'), s.palettes.PRISM)
});

layer.on('loaded', () => {
  const legend = layer.getViz().color.getLegend();
  // legend = [
  //   { name: 'Bicycle', value: { r: 95, g: 70, b: 144, a: 1 } },
  //   { name: 'Car', value: { r: 29, g: 105, b: 150, a: 1 } }
  //   { name: 'Bus', value: { r: 56, g: 166, b: 165, a: 1 } }
  //   { name: 'Others', value: { r: 15, g: 133, b: 84, a: 1 } }
  // ]
});
```

* Preview
legend
 
* [x] Add and fix specs
* [x] Add documentation
* [x] Add Legends example for colors (to be improved)
* [ ] Add Legends example for imageList

### DIFF
--- a/src/renderer/viz/expressions/basic/property.js
+++ b/src/renderer/viz/expressions/basic/property.js
@@ -63,10 +63,12 @@ export default class Property extends BaseExpression {
         if (!metaColumn) {
             throw new Error(`Property '${this.name}' does not exist`);
         }
+
         this.type = metaColumn.type;
 
         if (this.type === 'category') {
             this.numCategories = metaColumn.categories.length;
+            this.categories = metaColumn.categories;
         }
     }
 
--- a/src/renderer/viz/expressions/basic/property.js
+++ b/src/renderer/viz/expressions/basic/property.js
@@ -58,6 +58,12 @@ export default class Property extends BaseExpression {
         return feature[this.name];
     }
 
+    getCategories () {
+        return this.type === 'category'
+            ? this.categories
+            : [];
+    }
+
     _bindMetadata (meta) {
         const metaColumn = meta.properties[this.name];
         if (!metaColumn) {
--- a/src/renderer/viz/expressions/ramp.js
+++ b/src/renderer/viz/expressions/ramp.js
@@ -121,7 +121,7 @@ export default class Ramp extends BaseExpression {
     eval (feature) {
         this.palette = this._calcPaletteValues(this.palette);
 
-        const texturePixels = this._computeTextureIfNeeded(); 
+        const texturePixels = this._computeTextureIfNeeded();
         const input = this.input.eval(feature);
         const numValues = texturePixels.length - 1;
         const m = (input - this.minKey) / (this.maxKey - this.minKey);
--- a/src/renderer/viz/expressions/ramp.js
+++ b/src/renderer/viz/expressions/ramp.js
@@ -133,7 +133,26 @@ export default class Ramp extends BaseExpression {
         return color;
     }
 
-    getValue (index) {
+    getLegend () {
+        if (this.input.type === inputTypes.CATEGORY) {
+            return this.input.categories.map(this._getLegendCategoryValue);
+        }
+    }
+
+    _getLegendCategoryValue (category, index) {
+        return {
+            name: this._getCategoryName(category),
+            value: this._getRampValueByIndex(index)
+        };
+    }
+
+    _getRampValueByIndex (index) {
+        if (this.palette.type === paletteTypes.IMAGE) {
+            return this.palette[`image${index}`]
+                ? this.palette[`image${index}`].url
+                : null;
+        }
+
         this.palette = this._calcPaletteValues(this.palette);
 
         const texturePixels = this._computeTextureIfNeeded();
@@ -148,14 +167,9 @@ export default class Ramp extends BaseExpression {
         return color;
     }
 
-    getLegend () {
-        if (this.input.type === inputTypes.CATEGORY) {
-            const legend = this.input.categories.map((category, index) => {
-                return { name: category.name, color: this.getValue(index) };
-            });
-
-            return legend;
-        }
+    _getCategoryName (category) {
+        const DEFAULT_OTHERS_NAME = 'Others';
+        return category.name ? category.name : DEFAULT_OTHERS_NAME;
     }
 
     _getPaletteValues () {
--- a/src/renderer/viz/expressions/ramp.js
+++ b/src/renderer/viz/expressions/ramp.js
@@ -133,6 +133,78 @@ export default class Ramp extends BaseExpression {
         return color;
     }
 
+    /**
+     * Get the value associated with each category
+     *
+     * @returns {Array} Array of { name, value } objects
+     *
+     * @example <caption>Get the color associated with each category</caption>
+     * const s = carto.expressions;
+     * const viz = new carto.Viz({
+     *   color: s.ramp(s.prop('vehicles'), s.palettes.PRISM)
+     * });
+     *
+     * layer.on('loaded', () => {
+     *   const legend = layer.getViz().color.getLegend();
+     *   // legend = [
+     *   //   { name: 'Bicycle', value: { r: 95, g: 70, b: 144, a: 1 } },
+     *   //   { name: 'Car', value: { r: 29, g: 105, b: 150, a: 1 } }
+     *   //   { name: 'Bus', value: { r: 56, g: 166, b: 165, a: 1 } }
+     *   //   { name: 'Others', value: { r: 15, g: 133, b: 84, a: 1 } }
+     *   // ]
+     * });
+     *
+     * @example <caption>Get the color associated with each category (String)</caption>
+     * const s = carto.expressions;
+     * const viz = new carto.Viz(`
+     *   color: ramp($vehicles, PRISM)
+     * ´);
+     *
+     * layer.on('loaded', () => {
+     *   const legend = layer.getViz().color.getLegend();
+     *   // legend = [
+     *   //   { name: 'Bicycle', value: { r: 95, g: 70, b: 144, a: 1 } },
+     *   //   { name: 'Car', value: { r: 29, g: 105, b: 150, a: 1 } }
+     *   //   { name: 'Bus', value: { r: 56, g: 166, b: 165, a: 1 } }
+     *   //   { name: 'Others', value: { r: 15, g: 133, b: 84, a: 1 } }
+     *   // ]
+     * });
+     *
+     * @example <caption>Get the image url associated with each category</caption>
+     * const s = carto.expressions;
+     * const viz = new carto.Viz({
+     *   symbol: s.ramp(s.prop('vehicles'), s.imageList([s.BICYCLE, s.CAR, s.BUS]))
+     * });
+     *
+     * layer.on('loaded', () => {
+     *   const legend = layer.getViz().symbol.getLegend();
+     *   // legend = [
+     *   //   { name: 'Bicycle', value: bicycleImageUrl },
+     *   //   { name: 'Car', value: carImageUrl }
+     *   //   { name: 'Bus', value: busImageUrl }
+     *   // ]
+     * });
+     *
+     * @example <caption>Get the image url associated with each category (String)</caption>
+     * const s = carto.expressions;
+     * const viz = new carto.Viz(`
+     *   symbol: ramp('$vehicles'), imageList([BICYCLE, CAR, BUS]))
+     * `);
+     *
+     * layer.on('loaded', () => {
+     *   const legend = layer.getViz().symbol.getLegend();
+     *   // legend = [
+     *   //   { name: 'Bicycle', value: bicycleImageUrl },
+     *   //   { name: 'Car', value: carImageUrl }
+     *   //   { name: 'Bus', value: busImageUrl }
+     *   // ]
+     * });
+     *
+     * @memberof carto.expressions.Ramp
+     * @name getLegend
+     * @instance
+     * @api
+     */
     getLegend () {
         if (this.input.isA(Linear)) {
             return this._computeTextureIfNeeded();
--- a/src/renderer/viz/expressions/ramp.js
+++ b/src/renderer/viz/expressions/ramp.js
@@ -134,8 +134,12 @@ export default class Ramp extends BaseExpression {
     }
 
     getLegend () {
+        if (this.input.isA(Linear)) {
+            return this._computeTextureIfNeeded();
+        }
+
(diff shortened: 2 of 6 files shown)


makella:
hey @elenatorro!

This is going to make things so much easier!! Awesome!

### Updated example
I made you a new example that shows tree species in San Francisco (fewer categories, fewer colors, shorter legend!)
 -  I went ahead and replaced your `legends.html` file with this new one

### Issues/Questions:

#### Other style properties
- I noticed that we don't bring the outline color over to the legend. Is there a reason for that? For that reason, I removed `stroke` from the points.
- Another example here would be opacity so the colors in the legend match the features on the map

#### Top and categories breaks legend
- I tested using `top` with categories, and the legend still lists all the categories in the data. But only colors the `top` categories symbolized on the map
  - I wasn't sure how else to show you, so in the same folder there is a file **`legend-top-5.html`** that shows this

#### Numeric attributes
- I also tested using a numeric attribute to see how the legend would populate and it doesn't work. Is that because we are only retrieving string information for the legend?
  - I wasn't sure how to show you this either, so in the same folder, the file **`legend-number.html`** tests this

#### Polygons and lines
- What is the plan for other geometry types and legend styles?
  - I tested polygons and lines and realized that we still show a point for polygons and lines in the legends. Are there plans to update the legend styles to account for multiple geometry types?
  - I know there is work planned for legends and Airship... is that what we will be using?

elenatorro:
Hi! Lots of good points here! 🎉 

**Other style properties**

You can also use the stroke color for that, I have updated the example 'legend-stroke'. We can also get the opacity and apply it to the color, but the user is responsible to build the legend and apply it. It can be done by getting the opacity value and apply it to the color in the loop.

**Top and categories breaks legend**

I had a personal example using top but listing only the number of features I passed to the top expression. I think the user should not be responsive of this, so I am going to fix this use case to send only the top N values :) 

**Numeric attributes**

I think I missunderstood the issue and though it was only going to be available for category values, I will take a look.

**Polygons and lines**

As I mention in 'Other style properties', these points I added are just HTML & CSS elements to display the legend, but this is just an example and it is not related with CartoVL library itself.  As you said, in the case of Airship we will use this method only to get the values, then Airship is responsible to build the legends beautifully :)

**Extra: images!**

I fixed an issue with the legends for images (symbols) and added an example that looks like this, I think it is also useful:

captura de pantalla 2018-08-21 a las 11 46 51

makella:
@elenatorro NICE!! Great examples!! Thank you for the stroke color example too!! I've always wished we could have that in Builder :)

And awesome w/ the `top`! Regarding that, I think I see another issue 
`color: ramp(top($common_species,5), bold)` should color the top 5 values and then everything else to other, right?

Regarding polygons and lines...

Can we display a square (instead of point) for polygons and a line for lines? I understand it will come from airship but in the meantime, is there a way we can access the dataset geometry for the legend?

makella:
> we are going to add an option to name the 'others' name value to 'Others' by default, unless the user passes configuration options (this must be added clearly in the documentation)

If you think it would be helpful, in our example, we can set all others to `Other Species` or something so it is clear where that is done.

your call!

elenatorro:
I added this example http://127.0.0.1:8080/examples/misc/legends/legend-lines.html to test:
* Line geometry in the legend
* Custom options: by doing `getLegends({ defaultOthers: 'Default' })` we set the 'others' category name to 'Default'. I have not added this example to the documentation *yet* because I would like to discuss the naming.

captura de pantalla 2018-08-21 a las 17 44 38

makella:
this is great @elenatorro !!! love, love, love

src/renderer/viz/expressions/basic/property.js
```diff
@@ -58,15 +58,29 @@ export default class Property extends BaseExpression {
...
+        const metaColumn = metadata.properties[this.name];
         if (!metaColumn) {
             throw new Error(`Property '${this.name}' does not exist`);
         }
+
+        this._metadata = metadata;
+        this._categories = metaColumn.categories;
         this.type = metaColumn.type;
 
+        // FIXME
```
Jesus89: Why FIXME?

elenatorro:
Thanks @Jesus89 ! There are still things to discuss (that is why I added the DNM label, but wanted to open the PR to keep track of the issue). I am going to update the description with the current status of the task.

src/renderer/viz/expressions/basic/property.js
elenatorro: Because I am not very happy generating a getter in the constructor 😅but this can be addressed in other task, this one is pretty big already.

davidmanzanares:
Good work!

Regarding the API, I think there is an extra array that is not needed, see the `"values": [` in:
```
[
    {
        "name": " Brisbane Box",
        "values": [
            {
                "r": 127,
                "g": 60,
                "b": 141,
                "a": 1
            }
        ]
    },
```

Also I think that your util function `rgbToHex` should probably be a method in the returned color object, but I also think this is outside the scope of this PR, do you agree on creating an issue with that?

Regarding `top` and `others` I don't agree on making it configurable versus using a long cartovl-prefixed . I think users tend to not read all the docs, and therefore name clashing will be common if the default is a simple "Others". Also, if we make this configurable for `getLegend` we should do the same for `eval` for consistency reasons, and therefore we should do the same for the `.value` getter, but you cannot pass an argument there because it is a getter.

Regarding the API for numeric properties with a `linear` I think we should avoid returning ranges (from-to) because that gives the impression of classification, but we are not really classifying the input (not for linear). The usage of ranges sounds great with classifiers though (quantiles, ...).

This `http://127.0.0.1:8080/examples/misc/legends/legend-polygon.html` feels broken to me because the maximum number of floors is not 47

elenatorro:
> Also I think that your util function rgbToHex should probably be a method in the returned color object, but I also think this is outside the scope of this PR, do you agree on creating an issue with that?

Yes, I thought the same, but as you suggest I prefer to do it in a different issue.

> Regarding top and others ... 

What do you mean by 'using a long cartovl-prefixed'? I am also concerned about the fact of the 'Others' collision, but in my opinion this is about labeling, and we do not return labels in `eval()` method.

> from, to 

You totally are right, I will return an array as I did before.

> legend-polygon example

Checking it!

davidmanzanares:
:+1:

I meant to use the same as the current `top.eval`: `'CARTOVL_TOP_OTHERS_BUCKET'`

elenatorro:
About what @davidmanzanares mentions here: 

>  I think there is an extra array that is not needed

In my opinion each legend should return a 'values' array, that may vary depending on the data type (numeric, categorical) or even the expression. That is way it is an array, even if it contains only one value. The first value is **always** the value provided by `eval`. 

At the beginning, I though the values could be something like:

```json
[
    {
        "name": " Brisbane Box",
        "values": {
            "color": {
                "r": 127,
                "g": 60,
                "b": 141,
                "a": 1
            }
    }
```
I think this would be hard to maintain, but maybe someone has a better solution for this :) Now, this is an example of how it works by using numeric values (from the documentation):

```js
const s = carto.expressions;
const viz = new carto.Viz({
  color: s.ramp(s.linear(s.prop('numeric'), 1, 100), s.palettes.PRISM)
});
layer.on('loaded', () => {
  const legend = layer.getViz().color.getLegend({
      samples: 4
  });

  // legend = [
  //   { name: 'numeric', values: [// rgba color, [0, 25]] },
  //   { name: 'numeric', values: [// rgba color, [25, 50]] }
  //   { name: 'numeric', values: [// rgba color, [50, 75]] }
  //   { name: 'numeric', values: [// rgba color, [75, 100]] }
  // ]
});
```

davidmanzanares:
I'm sorry, but I'm still not convinced about the API.

When dealing with categories I still see an extra level of indirection (the array) and I think this is, in the end, to try to be consistent with the numeric case, but I also think that the numeric case is not handled well.

When dealing with numbers the part `{name: 'numeric'` is repeated for every color, which for me doesn't make sense because logically, the type of the legend is above each particular entry in that array (i.e. you cannot mix 'numeric' with category names, but the API gives the impression that could happen).

Also regarding numbers, I see you removed the `from` - `to` keywords, but the original API is basically the same, a range is still outputted. As I said earlier, I don't agree with showing ranges because that gives the impression that every value in that range will be colored the same, and that's not the case. My proposal for this is to show the evaluated colors at some points, not at some ranges (since ranges are not being used for rendering).

Some ideas for the API:

```
// For categories
{
  type: 'category',
  mapping: [
   {key: "TypeA", value: {r:255,g:0,b:0, a:1}},
   {key: "TypeB", value: {r:1,g:22,b:1, a:1}},
   {key: "CARTOVL_TOP_OTHERS_BUCKET", value: {r:33,g:33,b:33,a:1}}
  ]
} 

// For numbers
{
  type: 'number',
  mapping: [
    {key: 0, value: {r:255,g:0,b:0, a:1}},
    {key: 20, value: {r:255,g:0,b:0, a:1}},
    {key: 40, value: {r:255,g:0,b:0, a:1}},
    {key: 60, value: {r:255,g:0,b:0, a:1}},
    {key: 80, value: {r:255,g:0,b:0, a:1}},
    {key: 100, value: {r:255,g:0,b:0, a:1}},
  ]
}
```

elenatorro:
Good examples! The API is totally open for discussion.

I love the idea of using 'name' instead of key, and also what you say about the ranges. I used this approach because it was _easier_ to build the legend, but I think your suggestion is more accurate.

> `name: 'numeric'`

My mistake here, I used a bad example. "numeric" is not the type, but the property name. It is true that it is repeated in each legend, though, so I like the idea of including the type in the API. What about including the property name in the API as well?

```js
{
  type: 'number',
  name: 'whatever',
  mapping: [
    {key: 0, value: {r:255,g:0,b:0, a:1}},
    {key: 20, value: {r:255,g:0,b:0, a:1}},
    {key: 40, value: {r:255,g:0,b:0, a:1}},
    {key: 60, value: {r:255,g:0,b:0, a:1}},
    {key: 80, value: {r:255,g:0,b:0, a:1}},
    {key: 100, value: {r:255,g:0,b:0, a:1}},
  ]
}
```

I am not sure about 'mapping' for legend values, I am looking for alternatives.

davidmanzanares:
Regarding `name` the problem is that I wouldn't expect to see the same name for `top($cat, 5)`, `$cat` and `buckets($cat, 'A', 'B')`. At the end, the input for `ramp` can be much more than a property.

src/renderer/viz/expressions/ramp.js
```diff
@@ -135,6 +143,262 @@ export default class Ramp extends BaseExpression {
...
+        const firstChild = input[input.childrenNames[0]];
+        return firstChild.isA(Property)
+            ? this._getLegendLinearProperty(firstChild, config)
+            : this._getLegendLinear(firstChild, config);
+    }
+
+    _getLegendLinearProperty (property, config) {
+        const name = property.name;
+        const feature = {};
+        const breakpoints = _getBreakpoints(this._metadata.properties[name], config.samples);
```
davidmanzanares: Don't you think it would be better (and easier) to get the min/max values from `linear`. Of course, that won't output always the same as this, but if the user is using `linear` with custom `min`/`max` values is probably for a good reason (like having extreme outliers). In that case, using regular intervals will produce a bad result since the outliers will cause to sample much on the clamped region. Also, I think it's more expected to see values ranges from the linear's min to linear's max. On top of that, I think we would be more generic and would reduce the coupling with linear's internals and property's internals.

src/renderer/viz/expressions/ramp.js
```diff
@@ -135,6 +143,262 @@ export default class Ramp extends BaseExpression {
...
+        return {
+            name: this._getCategoryName(category.expr),
+            values: value ? [ value ] : null
+        };
+    }
+
+    _getLegendCategoryValue (category, index, config) {
+        const value = this._getRampValueByIndex(index);
+
+        if (this.input.isA(Top) && this.input.numBuckets === index) {
```
davidmanzanares: I think the problem here is more in general, not just with this line, but the line is a good example.

The problem I see is that instead of using the public API with `eval()` which I think it's powerful enough for most of the implementation here, we are using introspection and private APIs to go case by case. 

I know that `eval` may not be enough or may need to be improved in some cases, but I think that's the way to go since I would expect less complexity (less case by case managament) and less code.

src/renderer/viz/expressions/ramp.js
```diff
@@ -135,6 +143,262 @@ export default class Ramp extends BaseExpression {
...
+        }
+
+        return {
+            name: this._getCategoryName(category.name),
+            values: value ? [ value ] : null
+        };
+    }
+
+    _getRampValueByIndex (index, config) {
+        if (this.palette.type === paletteTypes.IMAGE) {
```
davidmanzanares: I see a similar problem here too, since `ramp` already has an `eval` method, we could use that

src/renderer/viz/expressions/ramp.js
```diff
@@ -10,6 +10,15 @@ import ImageList from './ImageList';
 import Linear from './linear';
 import Top from './top';
 
+const DEFAULT_OTHERS_NAME = 'CARTOVL_TOP_OTHERS_BUCKET';
```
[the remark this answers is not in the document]
elenatorro: But there are other expressions that can have 'others', right? Not only top.

src/renderer/viz/expressions/ramp.js
```diff
@@ -135,6 +143,262 @@ export default class Ramp extends BaseExpression {
...
+                ? this._getLegendList(config)
+                : this._getLeyendCategories(config);
+        }
+    }
+
+    _getLegendLinear (input, config) {
+        const firstChild = input[input.childrenNames[0]];
+        return firstChild.isA(Property)
+            ? this._getLegendLinearProperty(firstChild, config)
+            : this._getLegendLinear(firstChild, config);
```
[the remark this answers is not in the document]
elenatorro: Why? It is looking for the first property, let me think of a better way to do this 👍

elenatorro:
Question: what about 'data' instead of 'mapping'? Is it too generic? cc @davidmanzanares

davidmanzanares:
I'm not a fan of either actually :joy: although `data` seems better. But if there are more ideas...

ramiroaznar:
Love it! :heart: @makella @elenatorro

Issue #858, opened 2018-08-17
Using viewportFeatures with MVT source does not work as expected with different zoom levels


elenatorro:
# Description

Having a MVT source, `viewportFeatures` expression does not count properly the features in the bbox.

## Code example

```js
    const viz = new carto.Viz(`
      @attr: viewportFeatures($attr_0)
     ...
    `);

    const layer = new carto.Layer('layer', sourceMVT, viz);
    
    layer.addTo(map);

    layer.on('updated', () => {
      const features = layer.getViz().variables.attr.eval();
      if (features && features.length > 0) {
        console.log(features.length);
      }
    });
```

davidmanzanares:
Fixed by #870

PR #870, opened 2018-08-20
Fix viewport counting

davidmanzanares:
Two corner cases have been fixed for issue https://github.com/CartoDB/carto-vl/issues/858

- Fix viewport counting with polygons with zero vertices. (Yes, we were receiving zero-vertex polygons from Maps API :joy:, cc: @jgoizueta @rochoa @Algunenano ). In fact, this was the most visible issue.
- Fix viewport counting when there were no tiles

### DIFF
--- a/src/renderer/Dataframe.js
+++ b/src/renderer/Dataframe.js
@@ -60,6 +60,10 @@ export default class Dataframe {
                         aabb = _updateAABBForGeometry(feature[j], aabb, type);
                     }
 
+                    if (aabb.minx === Number.POSITIVE_INFINITY) {
+                        aabb = null;
+                    }
+
                     aabbList.push(aabb);
                 }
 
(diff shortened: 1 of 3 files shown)


Algunenano:
> Fix viewport counting with polygons with zero vertices. (Yes, we were receiving zero-vertex polygons from Maps API joy, cc: @jgoizueta @rochoa @Algunenano ). In fact, this was the most visible issue.

If you have an example (dataset + tile [z/x/y]), can you open a issue in Windshaft so we can have a look please?

davidmanzanares:
Sure! I'll contact you off-channel due to privacy issues

davidmanzanares:
After debugging it in more detail with @jgoizueta too, we saw that the problem was that the source used a high buffersize, which made complete features disappear due to clipping. Since we already manage the 0-vertex after clipping case with this PR everything seems ok, both the frontend (now) and the backend (always).

:rocket:

PR #876, opened 2018-08-22
Check second param for buckets() is an array

VictorVelarde:
### Context
Current `buckets()` constructor breaks if the second parameter is not an array, throwing: `Cannot read property 'map' of undefined`message.

This PR adds the type check, which throws a more specific error. For example, ` s.buckets($myProp, 10)` produces:

```
'buckets(): invalid second parameter 'list' 
    expected type was one of number-array,category-array, actual type was 'number''
```


src/renderer/viz/expressions/buckets.js
```diff
@@ -71,6 +71,8 @@ export default class Buckets extends BaseExpression {
             looseType = input.type;
         }
 
+        checkType('buckets', 'list', 1, ['number-array', 'category-array'], list);
```
davidmanzanares: The check should be on `_bindMetadata` because in the constructor you don't know the type of references to variables

test/unit/renderer/viz/expressions/buckets.test.js
```diff
@@ -65,6 +65,11 @@ describe('src/renderer/viz/expressions/buckets', () => {
 
                     expect(response).toEqual(expected);
                 });
+
+                it('should throw an error when not using an array for the breakpoint', () => {
```
davidmanzanares: I think the util function `validateCompileTypeError` would be better, at least that will take care of calling `bindMetadata` for you

src/renderer/viz/expressions/buckets.js
VictorVelarde: Thx David!, I saw a similar one on that method, but I wasn't sure

test/unit/renderer/viz/expressions/buckets.test.js
VictorVelarde: 👍 I'll have a look at it

src/renderer/viz/expressions/buckets.js
VictorVelarde: Done! I have also fixed the constructor accordingly

test/unit/renderer/viz/expressions/buckets.test.js
VictorVelarde: I applied your suggestions. Same results with less code, thx!

Issue #833, opened 2018-08-10
source.MVT fails for missing layer in multiple layers tile


rochoa:
While working with the MVT source and single layer tiles, we early-return an empty Dataframe as soon as we see the buffer is empty. See https://github.com/CartoDB/carto-vl/blob/b41cd8c43a388fbfcfec36f1fbe72aff2f476d39/src/sources/MVT.js#L148-L150

However, that's when you have a multiple layers tile, the buffer might not be zero but your layer might not be present because the tile contains only data for other layers. In that case, we throw an exception as if we were handling a single layer tile where we cannot find the specific layerId https://github.com/CartoDB/carto-vl/blob/b41cd8c43a388fbfcfec36f1fbe72aff2f476d39/src/sources/MVT.js#L153-L155

Example:

```js
const tiles = [URL1, URL2, URL3, URL4];
const metadata = {'...': '...'};
const source0 = new carto.source.MVT(tiles, metadata, { layerID: 'layer0' });
const source1 = new carto.source.MVT(tiles, metadata, { layerID: 'layer1' });
```

This will throw an exception if one of the tiles don't have data for one of the layers.

davidmanzanares:
By reading the code I would expect that
```
        if (!mvtLayer) {
            throw new Error(`LayerID '${this._options.layerID}' doesn't exist on the MVT tile`);
        }
```
would be the throwing code instead of
```
        if (Object.keys(tile.layers).length > 1 && !this._options.layerID) {
            throw new Error(`LayerID parameter wasn't specified and the MVT tile contains multiple layers: ${JSON.stringify(Object.keys(tile.layers))}`);
        }
```

The code I mention could be removed.

davidmanzanares:
Fixed by #880

PR #880, opened 2018-08-22
Fix multilayer MVT

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/833

### DIFF
--- a/src/sources/MVT.js
+++ b/src/sources/MVT.js
@@ -110,7 +110,7 @@ export default class MVT extends Base {
      * @memberof carto.source
      * @api
      */
-    constructor (templateURL, metadata = new Metadata(), options = { layerId: undefined, viewportZoomToSourceZoom: Math.ceil, maxZoom: undefined }) {
+    constructor (templateURL, metadata = new Metadata(), options = { layerID: undefined, viewportZoomToSourceZoom: Math.ceil, maxZoom: undefined }) {
         super();
         this._templateURL = templateURL;
         if (!(metadata instanceof Metadata)) {
@@ -157,7 +157,7 @@ export default class MVT extends Base {
         const mvtLayer = tile.layers[this._options.layerID || Object.keys(tile.layers)[0]];
 
         if (!mvtLayer) {
-            throw new Error(`LayerID '${this._options.layerID}' doesn't exist on the MVT tile`);
+            return { empty: true };
         }
 
         const { geometries, properties, numFeatures } = this._decodeMVTLayer(mvtLayer, this._metadata, MVT_EXTENT);
--- a/debug/maps-api-MVT.html
+++ b/debug/maps-api-MVT.html
@@ -1,0 +1,92 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>Add layer | CARTO</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta charset="UTF-8">
+    <!-- Include CARTO VL JS -->
+    <script src="../dist/carto-vl.js"></script>
+    <!-- Include Mapbox GL JS -->
+    <script src="https://libs.cartocdn.com/mapbox-gl/v0.45.0-carto1/mapbox-gl.js"></script>
+    <!-- Include Mapbox GL CSS -->
+    <link href="https://libs.cartocdn.com/mapbox-gl/v0.45.0-carto1/mapbox-gl.css" rel="stylesheet" />
+    <link rel="stylesheet" type="text/css" href="../examples/style.css">
+</head>
+
+<body>
+    <div id="map"></div>
+    <aside class="toolbox">
+        <div class="box">
+            <header>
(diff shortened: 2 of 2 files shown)


src/sources/MVT.js
```diff
@@ -157,7 +157,7 @@ export default class MVT extends Base {
         const mvtLayer = tile.layers[this._options.layerID || Object.keys(tile.layers)[0]];
 
         if (!mvtLayer) {
-            throw new Error(`LayerID '${this._options.layerID}' doesn't exist on the MVT tile`);
+            return { empty: true };
```
rochoa: I don't know if this is enough.

If you have `layer0` and `layer1` as layer IDs, the tile has data for `layer0` but not for `layer1`, you ask for data for `layer1`, then `tile.layers[this._options.layerID || Object.keys(tile.layers)[0]]` will put the data from `layer0` in `mvtLayer`, it won't be a false-ish value but it will contain data for the incorrect layer.

davidmanzanares: Yes but in that case, the user is doing something that is explicitly documented as forbidden. The first tile with both layers will throw in that example.

Also, is there a way to control that without removing the functionality of allowing users to not define a layerID when there is only one layer? The way I see it, this is a limitation of the MVT format which doesn't include the layer name of layers that have no features. Am I missing somthing?

rochoa: The user doesn't know in advance if different tiles will or will not have data for one of the layers.

As an example, you could have a two layers tileset, accessible through `example.com/{z}/{x}/{y}.mvt`, where different tiles are like:

```
tile z=0,x=0,y=0:
  -layer0: data-data-data
  -layer1: data-data

tile z=1,x=0,y=0:
  -layer1: data-data-data-data

tile z=1,x=1,y=1:
  -layer0: data
```

My point is:
 - If the user specified a `layerID`, we cannot assume there is no layer like that, in that case, we need to return `{ empty: true }` as the only alternative because we might be in one of the previous scenarios.
 - If no `layerID` is provided, we will attempt to access the first layer, if there is no data: first, we should not reach this part of the code anyway; in the case we do, we will again return an `{ empty: true }`. 🤔
 - But with the current code we could be getting data from another layer (the first layer found in the tile) if there is no data for the requested `layerID`.

Does that make sense?

davidmanzanares: > But with the current code we could be getting data from another layer (the first layer found in the tile) if there is no data for the requested layerID.

Yes, with the current code that could happen, but I think it is inevitable unless we remove the ability to use one single layer without specifying its ID, no?

rochoa: My proposal is just a bit different:

```diff
diff --git a/src/sources/MVT.js b/src/sources/MVT.js
index 5657fb6e..b8abdbc4 100644
--- a/src/sources/MVT.js
+++ b/src/sources/MVT.js
@@ -154,10 +154,10 @@ export default class MVT extends Base {
             throw new Error(`LayerID parameter wasn't specified and the MVT tile contains multiple layers: ${JSON.stringify(Object.keys(tile.layers))}`);
         }

-        const mvtLayer = tile.layers[this._options.layerID || Object.keys(tile.layers)[0]];
+        const mvtLayer = this._options.layerID ? tile.layers[this._options.layerID] : Object.keys(tile.layers)[0];

         if (!mvtLayer) {
-            throw new Error(`LayerID '${this._options.layerID}' doesn't exist on the MVT tile`);
+            return { empty: true };
         }

         const { geometries, properties, numFeatures } = this._decodeMVTLayer(mvtLayer, this._metadata, MVT_EXTENT);
```

davidmanzanares: I still don't understand it. The second change is the same as the one I did, the first one though... I don't understand its intent, and I think it's broken too: `Object.keys(tile.layers)[0]` will return the layer name I think, but you need the layer object. Changing `keys` by `values` will solve that, but then, I think the code behaves the same as the original one.

rochoa: Sorry, I obviously wrote a bug, I meant:

```js
const mvtLayer = this._options.layerID ? tile.layers[this._options.layerID] : tile.layers[Object.keys(tile.layers)[0]];
```

rochoa: So, if there is a layerID, you retrieve by IayerID, if there is no layerID you use the first layer. Previously there is a check for multiple layers. That way we avoid getting data from another layer if there is a layerID specified.

davidmanzanares: But previously you have `const mvtLayer = tile.layers[this._options.layerID || Object.keys(tile.layers)[0]];` which I think it's the same right?

rochoa: I was thinking all the time about `layerID` being undefined.  🤦‍♂️ 

You are right.

Now I get your point in:
> Yes, with the current code that could happen, but I think it is inevitable unless we remove the ability to use one single layer without specifying its ID, no?

Sorry for the noise.

Issue #704, opened 2018-07-13
Guides – How to work with labels


rochoa:

rochoa:
Past week I worked in an alternative for approaching labels with Mapbox GL. The idea was to use the `carto.source.*` as a Mapbox GL source to have a Mapbox GL symbols layer. It was not straightforward and required some hacks around synchronizing events. Furthermore, the source references are cloned for each viz object, making really difficult to work with your original CARTO VL source definition. It was not the best of the ideas because the labels are associated with the visualization and not with the source.

Apart from that, we still have the limitation of working with changes in the visualization, which will make the labels unsynced with any change in the visualization.

So, with all that in mind, I reached the same conclusion/realization that @VictorVelarde and @davidmanzanares mentioned in previous instances of this discussion. Instead of using a workaround for the labels—something that could regret in the future—we propose using interactivity and on-hover pop-ups as an alternative for labels until we have the right approach implemented within CARTO VL.

PR #883, opened 2018-08-22
Label guide

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/704

Incomplete!

I made a PoC that works by instantiating the map and using a custom MVT source in CARTO VL.

I think that we could add a `getTilesURL` function to Maps API source to avoid the need to instantiate the map directly and allow the usage of the easier Maps API sources. cc: @rochoa 

Regarding cartography, the example is just a PoC to see that the approach works, so it would be better to have a good map that really benefits from the labels. cc: @makella 

Also, after doing those two points we will need to add the real guide with all the documentation.

### DIFF
--- a/debug/label-guide.html
+++ b/debug/label-guide.html
@@ -1,0 +1,112 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>Add layer | CARTO</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta charset="UTF-8">
+    <!-- Include CARTO VL JS -->
+    <script src="../dist/carto-vl.js"></script>
+    <!-- Include Mapbox GL JS -->
+    <script src="https://libs.cartocdn.com/mapbox-gl/v0.45.0-carto1/mapbox-gl.js"></script>
+    <!-- Include Mapbox GL CSS -->
+    <link href="https://libs.cartocdn.com/mapbox-gl/v0.45.0-carto1/mapbox-gl.css" rel="stylesheet" />
+    <link rel="stylesheet" type="text/css" href="../examples/style.css">
+</head>
+
+<body>
+    <div id="map"></div>
+    <aside class="toolbox">
+        <div class="box">
+            <header>
+                <h1>Add layer</h1>
+            </header>
+            <section>
+                <p class="description open-sans">Add one CARTO layer to your map</p>
+            </section>
+            <footer class="js-footer"></footer>
+        </div>
+    </aside>
+    <script>
+        const map = new mapboxgl.Map({
+            container: 'map',
+            style: 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json',
+            center: [-74, 40.7],
+            zoom: 0
+        });
+
+        async function loadMVT() {
+            const mapConfig = {
+                buffersize: { mvt: 0 },
+                layers: [
+                    {
+                        id: 'myCartoLayer2',
+                        type: 'mapnik',
+                        options: {
+                            sql: 'SELECT * FROM pop_density_points'
+                        }
+                    }
+                ]
+            };
+            const response = await fetch('https://cartovl.carto.com/api/v1/map', {
+                method: 'POST',
+                headers: {
+                    'Accept': 'application/json',
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify(mapConfig)
+            });
+            const layergroup = await response.json();
+            const tilejson = layergroup.metadata.tilejson.vector;
+
+            map.on('load', function () {
+                map.addLayer({
+                    "id": "myCartoLayerMGLID",
+                    "type": "symbol",
+                    "source": {
+                        "type": "vector",
+                        "tiles": tilejson.tiles,
+                    },
+                    "source-layer": "myCartoLayer2",
+                    "layout": {
+                        "text-field": "{dn}",
+                        "text-size": 20,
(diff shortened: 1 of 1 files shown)


rochoa:
What if—considering the future Mapbox GL integration—it worked as another layer? That way you could define the available properties as variables.

```js
const source = new carto.source.Dataset('mytable');

const viz = new carto.Viz(`
  @name: $name
  @description: $description
`);

const layer = new carto.contrib.mapbox.SymbolLayer('labels', source, viz, {
    "layout": {
        "text-field": "{name}: {description}",
        "text-size": 20,
        "text-offset": [0, 0.6],
    },
    "paint": {
        "text-color": "#FFF"
    }
});

map.addLayer(layer);

layer.addTo(map, 'watername_ocean', {
    "layout": {
        "text-field": "{name}: {description}",
        "text-size": 20,
        "text-offset": [0, 0.6],
    },
    "paint": {
        "text-color": "#FFF"
    }
});
```

davidmanzanares:
I don't think I'm following you completely, the text properties are repeated for example.

In any case, I think that including MGL API within our API is a slippery slope.

I think we have 4 options:

1. Do nothing, leaving the example as is now
2. Adding `getTiles` to avoid the need for manual Maps API instantiation
3. Adding MGL text API to our API (your approach)
4. Adding our own text API, which could be implemented with MGL at first, but without coupling it to it

From this list, I think 2 and 4 are the best ones.

rochoa:
Hahaha, you got me. I agree with the slippery slope situation, that's why I used the trick of using the namespace `carto.contrib.mapbox`. 😀

Going with 2 is the most approachable option right now, it doesn't leak anything critical, neither couples us with Mapbox GL. I would use `getTilejson()`, though.

We could go with 4 when we have more time to think thoroughly.

makella:
soooo great!! i will take a closer look and propose some ideas for a stellar example! nice work!!

davidmanzanares:
Great!

100% agree about `getTilejson()`

davidmanzanares:
I was writing the `getTileJSON` method when I realized one problem with that API, it could change whenever the viz changes. Of course, we can add more API to let users overcome this, but I don't see a easy way out while sticking to solution `2`.

With this in mind, maybe the slippery slope is not that bad :joy: , what do you think?

davidmanzanares:
Another way would be to freeze the source disabling changes that require a reinstantiation, that would be the simplest and easiest solution, but it would be a limitation, of course.

rochoa:
The freezing will introduce a lot of confusion and it's hard to document. So that leads either to 3 or to 4 options. Initially, you preferred option 4 rather than 3, what do you think now?

davidmanzanares:
I think number 4 is much better, but the required effort is bigger too. But that also depends on what kind of things do we want to support.

If we do number 4 right, we would need to add to each text-available expression a MGLJSONgeneration method to generate MGL compatible expressions from ours. This shouldn't be difficult but it's clearly an appreciable amount of new code.

@makella What do you think that we should support for texts? (constant size, size by property, font, color, color by property...)

makella:
@davidmanzanares let me spend some time playing around with labeling today and looking at the Mapbox GL properties, and then I will respond to your question more thoughtfully :)

rochoa:
I see option 4 as a leaky abstraction. We will implement our API based on MGL capabilities. If so, how different is that from option 3?

davidmanzanares:
It will be different in that users won't need to understand MGL styles and in that, we could implement text ourselves without breaking the API.

With that said, I don't know which is better.

makella:
**WARNING: Incomplete comment...**: no need to respond until I investigate further and put in examples.

- the symbol sizes look a lot different in this set-up vs. the regular editor
- it doesn't look like aggregations work. what i'm seeing are labels but no features
for example:
`width: sqrt(clusterAVG($amount))`

screen shot 2018-08-23 at 3 39 28 pm

- concern with the automatic collision detection for labels. Is there a way to define collision hierarchy? For example, if you have a large point on your map that should get labeled before other smaller, values, is there a way to make collision data-driven vs. location-driven? 
  - reading a blog, found two ways

  > Layers higher in the layer listing are considered more important. They will be rendered first.

  > Within a layer, labels are rendered in the order they are defined in the data source.
  
want to test if this is influenced by `order:` since we can't control the draw order of features with SQL.

davidmanzanares:
> the symbol sizes look a lot different in this set-up vs. the regular editor

What do you mean exactly?

> it doesn't look like aggregations work

Ops, sorry, I should have commented that, no, they don't work because the example is just a PoC that is using the MVT source of CARTO VL

> Layers higher in the layer listing are considered more important. They will be rendered first.

> want to test if this is influenced by order: since we can't control the draw order of features with SQL.

No, this won't affect the tabel ordering

makella:
An update here. 

I showed @davidmanzanares in more detail the symbol size differences I was noticing with the BCN spend data. 

He is going to look into what is happening and then we will make a more solid plan for a guide, etc.

davidmanzanares:
After seeing it with @rochoa 's help we saw that the HTML PoC was using sampling. That's why there was that difference. It's not important since the idea of creating something like `getTileJSON()` won't have that issue.

After talking with you (@makella and @rochoa ) we decided to create a PoC branch with the `getTileJSON` method without taking into account the difficultness to make it work in dynamic scenarios.

makella:
@davidmanzanares ok, so should I wait for the other PoC branch? Let me know how to proceed

davidmanzanares:
Yes, I think it's better to wait for the branch

Issue #850, opened 2018-08-16
Inconsistent constant expressions


Jesus89:
There are some issues regarding constant expressions in Viz lang. Constant expressions can be constant numbers, palettes, images and placements. They are uppercase in the JS API and not sensitive in the String API. But there are some inconsistencies regarding names and namespaces.

### Constant Numbers

```js
carto.expressions.TRUE
carto.expressions.FALSE
carto.expressions.PI
...
```

```
true
False
PI
```

### Constant Palettes

There is no namespace in the String API.

```js
carto.expression.palettes.PRISM
carto.expression.palettes.EARTH
...
```

```
prism
earth
...
```

### Constant Images

There is no namespace in JS API or String API. Also, String API removes the underscore.

```js
carto.expressions.CAR
carto.expressions.TRIANGLE
carto.expressions.CIRCLE_OUTLINE
...
```

```
car
triangle
circleoutline
```

### Constant Placements

```js
carto.expressions.ALIGN_CENTER
carto.expressions.ALIGN_BOTTOM
...
```

```
align_center
align_bottom
```

My proposal is to review this from the point of view of the user and consider adding namespaces and more consistency to the API.

Jesus89:
I agree with adding those namespaces. However, I feel that using `aligncenter` for the String API and `ALIGN_CENTER` could be confusing for the user because IMO you expect to have the underscore in both constants.

elenatorro:
Yes, you are right. Although they are case agnostic, I would always show in the String API examples uppercase also to make it more intuitive.

PR #885, opened 2018-08-22
Fix inconsistent constant expressions

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/850


Jesus89:
Hi. Could you add in the description the final API in detail? To verify easily the final implementation :)

elenatorro:
Done @Jesus89 ! ✨

Jesus89:
Regarding singular and plural: we are always using the plural (`carto.expressions`, `carto.expressions.palettes`) with the exception of the carto.source (`carto.source.GeoJSON`).

Just a minor issue but, are all you OK with that? Do you think we should change it? How?

src/renderer/viz/parser.js
```diff
@@ -1,6 +1,6 @@
 
 import jsep from 'jsep';
-import * as functions from './expressions';
+import * as expressions from './expressions';
```
Jesus89: :heart:

test/unit/renderer/viz/expressions/binary.test.js
```diff
@@ -2,8 +2,8 @@ import * as s from '../../../../../src/renderer/viz/expressions';
 import { validateStaticType, validateDynamicType, validateMaxArgumentsError, validateCompileTypeError } from './utils';
 
 // Add custom toString function to improve test output.
-s.TRUE.toString = () => 's.TRUE';
-s.FALSE.toString = () => 's.FALSE';
+s.constants.TRUE.toString = () => 's.TRUE';
```
Jesus89: 's.constants.TRUE';
's.constants.FALSE';

Issue #847, opened 2018-08-16
Fix MGL integration inconsistencies


Jesus89:
- Register `moveObservers` twice with different `uids` for the same layer.
- Not synchronized layer.initialize inside paintCallback.
- Call `paintCallback` twice for a non-animated layer.

elenatorro:
About the `layer.initialize` should be async issue, I think we should relate it with https://github.com/CartoDB/carto-vl/issues/342 instead, what do you think?

Jesus89:
I think it is not related. The point with `layer.initialize` is to wait inside callback to requestData (inside initialize) to be completed.

Issue #700, opened 2018-07-13
Mapbox GL Custom WebGL Layers API integration


rochoa:
Depending on the status of https://github.com/mapbox/mapbox-gl-js/issues/6456, we should evaluate if we migrate from our current approach.

rochoa:
Yes, let's use 0.48 or master (if needed).

davidmanzanares:
The performance regression has been mostly fixed by the last merge, see https://github.com/mapbox/mapbox-gl-js/pull/7132

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/886

Issue #428, opened 2018-05-17
Add layer.removeFromMap() method


davidmanzanares:

davidmanzanares:
Integrating CARTO VL with the upstream custom layers proposal would fix this

Jesus89:
This is implemented as `layer.removeFrom(map)` in the branch https://github.com/CartoDB/carto-vl/pull/886

Jesus89:
Closed by https://github.com/CartoDB/carto-vl/pull/886

PR #886, opened 2018-08-22
Native MGL integration

Jesus89:
Related to https://github.com/CartoDB/carto-vl/issues/700, https://github.com/CartoDB/carto-vl/issues/428, https://github.com/CartoDB/carto-vl/issues/847.

This branch adds a re-implementation with some refactors for the MGL integration interface using the new Custom Layers API. Since this feature is not already in a stable version I have released from our fork the version `v0.48.0-carto1` ([branch](https://github.com/cartodb/mapbox-gl-js/tree/carto-custom-layers)), from the stable `v0.48.0` tag and the changes from `custom-layers`.

This is a pretty big PR. However, the main activity is in the files Layer, Renderer and Map. Here is a list of changes in this branch:

- Use `onAdd`, `onRemove`, `prerender` and `render` functions of Custom Layers
- Implement `Layer.removeFrom`
- Refactor `maps/Map` to match the new interface
- Allow `layer.update(source)` using current viz by default
- Get zoom and center from the matrix
- Request data when matrix changes
- Remove integrators. Instead of a set of integrators, keep a set of renderers associated to each map.
- Clean unused functions and attributes in class Layer
- Minor fixes in the test framework
- Use local basemaps for testing
- Update examples to the new MGL version
- Fix animation example
- Fix interactivity examples

If we are interested in testing MGL upstream I can release another version (v0.48.0-carto2) pulling the changes from master.


src/Layer.js
```diff
@@ -62,35 +48,26 @@ export default class Layer {
...
         this._context = new Promise((resolve) => {
             this._contextInitialize = resolve;
         });
 
-        this._integratorPromise = new Promise((resolve) => {
-            this._integratorCallback = resolve;
-        });
-
+        this.id = id;
+        this.type = 'custom';
```
davidmanzanares: Why is this added?

src/Layer.js
```diff
@@ -138,7 +115,7 @@ export default class Layer {
     }
 
     /**
-     * Add this layer to a map.
+     * Add this layer to a map. It waits for the map to be loaded.
```
davidmanzanares: I would change the wording

src/Layer.js
```diff
@@ -147,29 +124,51 @@ export default class Layer {
...
+    /**
+     * Remove this layer from a map.
+     *
+     * @param {mapboxgl.Map} map - The map on which to remove the layer
+     * @memberof carto.Layer
+     * @instance
+     * @api
+     */
+    removeFrom (map) {
+        map.removeLayer(this);
```
davidmanzanares: This will throw an error if the map hasn't been loaded yet and the layer hasn't been added yet

src/Layer.js
```diff
@@ -147,29 +124,51 @@ export default class Layer {
...
      * @memberof carto.Layer
-     * @async
      * @instance
+     * @async
      * @api
      */
     async update (source, viz) {
+        if (viz === undefined) {
+            // Use current viz
+            viz = this._viz;
```
davidmanzanares: I don't remember well but I think this wasn't done before for a reason, this should be tested well

src/Layer.js
```diff
@@ -303,26 +322,10 @@ export default class Layer {
...
-    getViz () {
-        return this._viz;
-    }
-
     getNumFeatures () {
         return this._renderLayer.getNumFeatures();
     }
 
-    getIntegrator () {
-        return this._integrator;
```
davidmanzanares: You removed this, but it is being used in interactivity.test `            layer.getIntegrator = () => mockIntegrator;`

src/Layer.js
```diff
@@ -334,46 +337,77 @@ export default class Layer {
...
-     *
-     * @fires updated
+     * Custom Layer API: `onAdd` function
      */
-    show () {
-        this._visible = true;
-        this._integrator.changeVisibility(this);
-        this.requestData();
-        this._fire('updated');
+    onAdd (map, gl) {
```
davidmanzanares: Can we change the name of this? Users may look at it and think that this is public and it is used to get an event when the map is loaded

src/Layer.js
```diff
@@ -334,46 +337,77 @@ export default class Layer {
...
+        this.renderer = getRenderer(map, gl);
+
+        // Register map event to request data tiles
+        // map.on('move', this.requestData.bind(this));
+
+        // Initialize render layer
+        this._renderLayer.renderer = this.renderer;
+        this._renderLayer.dataframes.forEach(d => d.bind(this.renderer));
+        this._contextInitialize();
+        this.requestMetadata();
```
davidmanzanares: and no requestData too?

src/Layer.js
```diff
@@ -334,46 +337,77 @@ export default class Layer {
...
-     *
-     * @fires updated
+     * Custom Layer API: `onRemove` function
      */
-    hide () {
-        this._visible = false;
-        this._integrator.changeVisibility(this);
-        this._fire('updated');
+    onRemove (map, gl) {
+        // Unregister map event
```
davidmanzanares: Why is this commented?

src/renderer/Renderer.js
```diff
@@ -51,6 +52,47 @@ export default class Renderer {
...
     }
 
+    /**
+     * Initialize renderer:
+     * - WebGL context
+     * - Map zoom and center
+     * @param {Map} map
+     * @param {WebGLRenderingContext} gl
+     */
+    initialize (map, gl) {
```
davidmanzanares: Passing the map here seems like a bad idea to me since this is coupling the renderer with the map

src/renderer/Renderer.js
```diff
@@ -95,52 +137,20 @@ export default class Renderer {
...
-     * Get Renderer visualization zoom
-     * @return {number}
-     */
-    getZoom () {
-        return this._zoom;
+    _initZoom (map) {
+        const b = map.getBounds();
+        const nw = b.getNorthWest();
+        const sw = b.getSouthWest();
+        const z = (util.projectToWebMercator(nw).y - util.projectToWebMercator(sw).y) / util.WM_2R;
```
davidmanzanares: I don't think this is a good idea either, the Renderer didn't know about webmercator or maps at all before

src/renderer/Renderer.js
```diff
@@ -95,52 +137,20 @@ export default class Renderer {
...
 
-    /**
-     * Set Renderer visualization zoom
-     * @param {number} zoom
-     */
-    setZoom (zoom) {
-        this._zoom = zoom;
+    _initCenter (map) {
+        const c = map.getCenter();
+        this.setCenter(c.lng / 180.0, util.projectToWebMercator(c).y / util.WM_R);
```
davidmanzanares: same

src/utils/util.js
```diff
@@ -54,6 +54,16 @@ export function isSetsEqual (a, b) {
     return a.size === b.size && [...a].every(value => b.has(value));
 }
 
+export function equalArrays (m1, m2) {
+    if (m1 && m2 && m1.length === m2.length) {
+        let ret = true;
+        for (let i = 0; i < m1.length; i++) {
+            ret &= m1[i] === m2[i];
```
davidmanzanares: this would be simpler and faster without the ret variable, early returning a false when the values mismatch

test/acceptance/e2e/cluster-aggr/scenario.js
```diff
@@ -1,8 +1,6 @@
 const map = new carto.Map({
     container: 'map',
-    background: 'black',
-    center: [-75, 40],
```
davidmanzanares: this was set to avoid draw ordering issues

test/integration/util.js
```diff
@@ -57,6 +57,6 @@ function project (coordinates) {
     const wm = projectToWebMercator(coordinates);
     return {
         x: mapSize * (0.5 + wm.x / WM_2R),
-        y: mapSize * (0.5 - wm.y / WM_2R + 0.03 /* offset */)
+        y: mapSize * (0.5 - wm.y / WM_2R)
```
davidmanzanares: Could you explain this since you introduced the magic `0.03`? Why was it added? The addition was done on: https://github.com/CartoDB/carto-vl/commit/7b46dee5273f30cc48216b9f1c04f180f2f7e834

examples/basics/add-layer.html
```diff
@@ -37,10 +37,7 @@ <h1>Add layer</h1>
       container: 'map',
       style: 'https://basemaps.cartocdn.com/gl/voyager-gl-style/style.json',
       center: [0, 30],
-      zoom: 2,
-      scrollZoom: false,
```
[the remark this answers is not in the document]
Jesus89: Just for testing. I'll restore it

src/Layer.js
Jesus89: The idea is to convert `carto.Layer` in a native MGL CustomLayer. I'll add a comment there

src/Layer.js
Jesus89: `Wait for the map to load`?

davidmanzanares: I would say something like: If the map's style is not loaded yet it will wait for it to add the layer as soon as possible

src/Layer.js
Jesus89: Yes and MGL this is the expected behaviour IMO.

src/Layer.js
Jesus89: I found an example using `layer.getViz()` as a second argument, so this could be the default behaviour. I'll add tests for this.

src/Layer.js
Jesus89: The interactivity object uses directly the map now, so I will update the tests with that. Thanks!

src/renderer/Renderer.js
Jesus89: Yes, you are right

src/utils/util.js
Jesus89: Yes, I think I changed it already :thinking:

test/acceptance/e2e/cluster-aggr/scenario.js
Jesus89: But this test uses `carto.Map` that has no `center`, `zoom` attributes...

test/integration/util.js
Jesus89: This changed was required to match the bounding box in the simulated mouse events. I have updated the bounding box in the `carto.Map` so this is not required anymore.

src/Layer.js
Jesus89: This was just a test. Removed now.

src/renderer/Renderer.js
Jesus89: Yes, I have decoupled the map so this functions are not in the Renderer anymore

src/renderer/Renderer.js
Jesus89: https://github.com/CartoDB/carto-vl/pull/886#discussion_r212274806

src/Layer.js
Jesus89: The first requestData is done in the `update` function called by the constructor.

src/Layer.js
davidmanzanares: ah, true... yes, I think we should ask them

src/Layer.js

test/acceptance/e2e/cluster-aggr/scenario.js
davidmanzanares: ah, true, yes, I probably copy pasted them from the example I was thinking about

test/integration/util.js
davidmanzanares: I don't really understand it, but if you are sure... :+1:

Jesus89:
Suggestions and more improvements added.

src/Layer.js
davidmanzanares: Yes, but since we are making things easier in addTo for a similar problem, we could improve this, but I realize that this feature request is not simple, I think we could create an issue with it, are you ok?

src/Layer.js
```diff
@@ -1,39 +1,25 @@
...
 
-/**
- * Updated events are fired every time that viz variables could have changed, like: map panning, map zooming, source data loading or viz changes.
- * This is useful to create external widgets that are refreshed reactively to changes in the CARTO VL map.
- *
- * @event updated
- * @type {LayerEvent}
- * @api
-*/
+function getRenderer (map, gl) {
```
elenatorro: This is a private function that should be placed at the bottom (after the class) and should be `_getRenderer` instead of `getRenderer`.

src/Layer.js
```diff
@@ -513,15 +499,45 @@ export default class Layer {
...
-            return this._integrator.renderer.getBounds();
+    _setZoomCenter (matrix) {
+        if (matrix) {
+            // Compute the zoom and center from the matrix.
+            // This is a solution to avoid subscribing to map events and
+            // make a better and efficient use of the Custom Layers interface.
+            // TODO: the best solution is to use the matrix at the shader
+            // level and remove the aspect and scale logic from the renderer
+            this.renderer.setCenter(
+                -(1 + 2 * matrix[12] / matrix[0]),
```
elenatorro: Could we move this `center` logic to an external utils file or similar?

src/Layer.js
```diff
@@ -513,15 +499,45 @@ export default class Layer {
...
+        const b = this.map.getBounds();
+        const nw = b.getNorthWest();
+        const sw = b.getSouthWest();
+        const z = (util.projectToWebMercator(nw).y - util.projectToWebMercator(sw).y) / util.WM_2R;
+        this.renderer.setZoom(z);
+    }
+
+    _setRendererCenter () {
+        const c = this.map.getCenter();
+        this.renderer.setCenter(c.lng / 180.0, util.projectToWebMercator(c).y / util.WM_R);
```
elenatorro: Use constant instead of a number (`180.0`)

Jesus89: This is legacy code. It can be done with projectToWebMercator(c).x, etc but this is shorter. However, I'll move this to an utils file

Issue #778, opened 2018-07-27
Add ViewportHistogram example


elenatorro:
In `v0.6.0` we have implemented the `ViewportHistogram` expression. There is an example that is not published in the examples page  [here](https://github.com/CartoDB/carto-vl/blob/master/examples/interactivity/histogram.html). 

It would be great to have an example with a legend showing the histogram.

PR #889, opened 2018-08-23
Add viewport histogram example

elenatorro:
Related with  https://github.com/CartoDB/carto-vl/issues/778

New method introduced: `viewportHistogram.getJoinedValues`. It is useful to be used with `ramp.getLegendValues` in order to draw charts.

* Using viewportHistogram expression to draw a bar chart from a categorical property

histogram-categorical

* Using viewportHistogram expression to draw a histogram from a numeric property

histogram-numeric


Jesus89:
Could you change the PR base to `572-add-get-legend-method` in order to check better the changes?

examples/misc/legends/legend-base.html
```diff
@@ -81,12 +81,15 @@ <h1>Tree Species in San Francisco</h1>
...
     });
+
     function hideLoader() {
       document.getElementById('loader').style.opacity = '0';
     }
+
     function rgbToHex(color) {
       return "#" + ((1 << 24) + (color.r << 16) + (color.g << 8) + color.b).toString(16).slice(1);
     }
+
```
VictorVelarde: +1 cause space matters!

src/renderer/viz/expressions/basic/property.js
```diff
@@ -71,20 +71,23 @@ export default class Property extends BaseExpression {
     }
 
     _bindMetadata (metadata) {
-        const metaColumn = metadata.properties[this.name];
-        if (!metaColumn) {
-            throw new Error(`Property '${this.name}' does not exist`);
-        }
+        if (metadata.properties) {
```
VictorVelarde: Maybe we could just end early, to avoid a big if
```js
if (!metadata.properties){
  return;
}
```

examples/misc/histogram.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
```
VictorVelarde: Regarding to the **misc** category / path ... Should we add this category and example to `examples.json` so it appears in the `http://127.0.0.1:8080/examples/` root?

examples/misc/histogram.html
```diff
@@ -0,0 +1,117 @@
...
+  <link rel="stylesheet" type="text/css" href="../style.css">
+</head>
+
+<body>
+  <!-- Map goes here -->
+  <div id="map"></div>
+  <aside class="chart-toolbox">
+    <div class="box">
+      <header>
+        <h1>Histogram</h1>
```
VictorVelarde: This looks more like a **bar-chart**. I think we should add a different title (even rename the file): https://www.forbes.com/sites/naomirobbins/2012/01/04/a-histogram-is-not-a-bar-chart/

Another graphical detail is that histograms usually are presented in horizontal and without a space between the bars

examples/misc/histogram.html
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    function computeColor(color) {
+      return `rgba(${color.r},${color.g},${color.b},${color.a})`;
+    }
+
+  </script>
+</body>
+
+</html>
```
VictorVelarde: ✨ The example looks pretty nice!

Bar animations are smooth when changing zoom, though sometimes they present a little "trembling" when using drag. Maybe we should add a **debounce**

examples/misc/histogram.html
elenatorro: Yes! But not yet. These examples are now for internal use, we have opened a new task to review legend examples and publish them here https://github.com/CartoDB/carto-vl/issues/948

examples/misc/histogram.html
```diff
@@ -0,0 +1,117 @@
+<!DOCTYPE html>
+<html>
```
davidmanzanares: I've seen a lot of flickering in the histogram when moving the map

src/renderer/viz/expressions/basic/property.js
davidmanzanares: Actually, I don't see why is this `if` in the first place. If there are no properties defined in the metadata this should fail hard, not silently ignore the problem (I think)

examples/misc/histogram.html
elenatorro: Yes, because it is updated on every layer update. Victor already mentioned that, so I am going to se how to debounce it.

davidmanzanares: I don't think we should debounce, that would only mitigate the problem and it would look like CVL was slow and debouncing its data. I think that the animation should be removed or be set in a way that was done from the previous state to the new one, not from the zero-state to the new one.

examples/misc/legends/legend-categorical-chart.html
```diff
@@ -124,15 +124,20 @@ <h1>Daytime sales</h1>
     function drawHistogramLegend() {
       const histogram = layer.viz.variables.histogram;
       const colorRamp = layer.viz.color;
-      const labels = histogram.value.map(value => value.x);
-      const data = histogram.value.map(value => value.y);
+
       const legendData = colorRamp.getLegendData().data;
       const histogramData = histogram.getAssingedData(legendData);
```
VictorVelarde: 👁 a little typo here  `getAssingedData`

src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
```diff
@@ -75,10 +74,102 @@ export default class ViewportHistogram extends BaseExpression {
...
+     * const data = viz.variables.histogram.getJoinedValues(legendData);
+     * // returns the following array
+     * // [
+     * //   { frequency: 10, key: 'truck', value: { r: 95, g: 70, b: 144, a: 1 } }
+     * //   { frequency: 20, key: 'bike', value: { r: 29, g: 105, b: 150, a: 1 } }
+     * //   { frequency: 30, key: 'car', value: { r: 56, g: 166, b: 165, a: 1 } }
+     * // ]
+     *
+    */
+    getJoinedValues (values) {
```
VictorVelarde: ✨ Nice useful method!

VictorVelarde: Just a question, doesn't values come from histogram itself? In that case we could remove the parameter. Regarding to its name, maybe it could be clearer without the "joined" word... how about `getFrequencyValues`?

examples/misc/legends/legend-categorical-chart.html
VictorVelarde: Forget it! I saw a commit and not the final result 😄

examples/misc/legends/legend-categorical-chart.html
```diff
@@ -0,0 +1,162 @@
...
+
+    const ctx = document.getElementById('histogram').getContext('2d');
+    // Define the source
+    const source = new carto.source.Dataset('ne_10m_populated_places_simple');
+    // Define the visualization
+    const viz = new carto.Viz(
+    ` @color: ramp($featurecla, bold) 
+      @histogram: viewportHistogram($featurecla)
+      color: @color
+      width: 35
```
VictorVelarde: With that big width, symbols overlap a lot, maybe we could reduce it a little

src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
davidmanzanares: But this is actually joining the results of the histogram with the ones of the legend

examples/misc/legends/legend-categorical-chart.html
VictorVelarde: image

examples/misc/legends/legend-categorical-chart.html
```diff
@@ -0,0 +1,162 @@
...
+            display: false
+          },
+          barPercentage: 1.0,
+          categoryPercentage: 1.0
+        }]
+      }
+    };
+
+    const chart = new Chart(ctx, {
+      type: 'bar',
```
VictorVelarde: I can see some 'trembling' on each update. I'll try to solve it somehow. A debounce like this might work:

```js
    const debounce = function (func, wait = 100) {
      let timeout;
      return function (...args) {
        clearTimeout(timeout);
        timeout = setTimeout(() => {
          func.apply(this, args);
        }, wait);
      };
    }

    const drawHistogramLegend = debounce(_drawHistogramLegend, 100);

    layer.on('loaded', () => {
      hideLoader();
      drawHistogramLegend();
    });

    layer.on('updated', drawHistogramLegend);

    function _drawHistogramLegend() {

      // WHATEVER...
```

src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
VictorVelarde: I see... It's just that the name didn't sound very clear to me

elenatorro: After asking people, we're going to keep using `getJoinedValues`, but in the documentation it is explained that it is sorted by frequency.

PR #891, opened 2018-08-24
Fix debug examples

Jesus89:


debug/interactive-based-styling.html
```diff
@@ -57,10 +57,9 @@ <h1>Interactive based styling</h1>
     // Define layer
     const source = new carto.source.Dataset('world_borders');
     const viz = new carto.Viz(`
-      // width: sqrt($pop_2015)/200
       color: opacity(BlueViolet, 0.5)
       strokeColor: opacity(BlueViolet, 0.8)
-      offset: placement(0, $pop2005/1000000)
+      // offset: placement(0, $pop2005/1000000)
```
davidmanzanares: I think that offset was there to test that offset worked fine with interactivity

Jesus89: :+1:

Issue #407, opened 2018-05-14
Allow global*($prop) usage with agg references (clusterAvg($prop))


davidmanzanares:
We should enable the usage of `global*($property)` expressions even when the same property is referenced inside a cluster aggregation function.

davidmanzanares:
Well... It works, but... I think it's 99% luck. Moreover, there are some possible optimizations around this problem.

Right now we are not controlling in CARTO VL this case at all. In fact, this is what we send to Maps API:
```
{
    "buffersize": {
        "mvt": 1
    },
    "layers": [
        {
            "type": "mapnik",
            "options": {
                "sql": "SELECT count,cartodb_id,the_geom,the_geom_webmercator FROM traffic_accidents_copy ",
                "aggregation": {
                    "columns": {
                        "_cdb_agg_avg_count": {
                            "aggregate_function": "avg",
                            "aggregated_column": "count"
                        }
                    },
                    "dimensions": {
                        "count": "count"
                    },
                    "placement": "centroid",
                    "resolution": 1,
                    "threshold": 1
                }
            }
        }
    ]
}
```

Maps API instantiation doesn't fail, although it should (right?) fail.

It would be easy to change CARTO VL to stop asking for globally aggregated data, but, in that case, we would need to get the metadata for all the columns in all cases (this may or may not hurt performance significantly in GeoJSON).

Another option would be to have two lists of columns: data required ones and metadata required ones. However, this would require more work.

cc: @jgoizueta

davidmanzanares:
Regarding the new point of the lack of control of the situation with un-aggregated and aggregated properties, this was already fixed (I tested it but I haven't search for a PR id).

Regarding the fix for this, we could abuse the MNS internal API more to hack more with strings. However, since @jgoizueta is going to work in temporal aggs., it doesn't make sense to leave this internal API.

My plan is to propose a new internal API for the MNS, get @jgoizueta review on it, apply it on existing expressions fixing this along the way and paving the way for the temporal agg expressions.

davidmanzanares:
Proposal for the MNS type:

```
const mns = new Set([
  {
  name: 'price',
  aggregations: new Set(['avg', 'max])
  usage: 'all'
  },
  {
  name: 'amount',
  usage: 'all'
  },
  {
  name: 'other',
  usage: 'metadata'
  },
  {
  name: 'date',
  granurality: ??? // to be defined later by @jgoizueta
  usage: 'all'
  }
]);
```

[...]

PR #893, opened 2018-08-24
Refactor MNS private API

davidmanzanares:
Regarding https://github.com/CartoDB/carto-vl/issues/407


src/client/windshaft.js
```diff
@@ -322,9 +329,8 @@ export default class Windshaft {
         if (!response.ok) {
             throw new Error(`Maps API error: ${JSON.stringify(layergroup)}`);
         }
-        this._URLTemplates = layergroup.metadata.tilejson.vector.tiles;
         return {
-            url: getLayerUrl(layergroup, LAYER_INDEX, conf),
+            urlTemplates: layergroup.metadata.tilejson.vector.tiles,
```
jgoizueta: This seems more sensible now

src/renderer/schema.js
```diff
@@ -1,40 +1,66 @@
...
-};
+
+export const IDENTITY = {};
 
 /*
-const schema = {
-    columns: ['temp', 'cat']
-}; */
+const mns = {
+  price:  new Set(['avg', 'max]),
```
jgoizueta: They're not Sets anymore, right?

src/client/windshaft.js
davidmanzanares: yep, updated

src/renderer/schema.js
davidmanzanares: yep, updated

davidmanzanares:
Tests finally pass, but performance may have regressed a lot when dealing with many properties, I'll review that case

davidmanzanares:
I was worried about the performance impact of line https://github.com/CartoDB/carto-vl/pull/893/files#diff-70dfd5511c05de98bb8c59e31215cf55R89 but after profiling and taking a closer look, it seems that shouldn't cause trouble since it will only be called once.

davidmanzanares:
Btw, I thought that this wouldn't solve the corner case of using a property only with global aggregations. I thought that it would still download the per-feature/tile data instead of just the metadata.

However, unless I'm missing something it works:

Instantiation request:
```
{
    "buffersize": {
        "mvt": 1
    },
    "layers": [
        {
            "type": "mapnik",
            "options": {
                "sql": "SELECT \"category\",\"amount\",\"cartodb_id\",\"the_geom_webmercator\" FROM spend_data ",
                "aggregation": {
                    "columns": {
                        "_cdb_agg_mode_category": {
                            "aggregate_function": "mode",
                            "aggregated_column": "category"
                        }
                    },
                    "dimensions": {},
                    "placement": "centroid",
                    "resolution": 1,
                    "threshold": 1,
                    "filters": {}
                },
                "dates_as_numbers": true,
                "metadata": {
                    "geometryType": true,
                    "columnStats": {
                        "topCategories": 32768,
                        "includeNulls": true
                    },
                    "sample": {
                        "num_rows": 1000,
                        "include_columns": [
                            "category",
                            "amount",
                            "cartodb_id"
                        ]
                    }
                }
            }
        }
    ]
}
```

Instantiation response:
```
...
},
                            "amount": {
                                "type": "number",
                                "min": 0.01,
                                "max": 15000,
                                "avg": 57.2069958615219,
                                "sum": 7188059.03000023
                            },
                            "cartodb_id": {
                                "type": "number",
                                "min": 22,
                                "max": 1178596,
                                "avg": 591151.2784241942,
                                "sum": 74278158134
                            },
                            "the_geom_webmercator": {
                                "type": "geometry"
                            }
...
```

MVT tile (converted to GeoJSON):
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          2.197265625,
          41.37680856570236
        ]
      },
      "properties": {
        "_cdb_agg_mode_category": "Moda y calzado",
        "_cdb_feature_count": 125650,
        "cartodb_id": 22
      },
      "id": 1
    }
  ]
}
```

Am I missing something? I thought that Maps API was unprepared for that.

cc: @jgoizueta

src/client/windshaft.js
```diff
@@ -244,26 +244,28 @@ export default class Windshaft {
...
                 }
             });
 
         return aggregation;
     }
 
     _buildSelectClause (MNS) {
-        const columns = MNS.columns.map(name => schema.column.getBase(name))
-            .concat(['the_geom_webmercator', 'cartodb_id']);
+        const columns = Object.keys(MNS).concat(['the_geom_webmercator', 'cartodb_id']);
```
jgoizueta: Is the following uniqueness filter needed now?

davidmanzanares: I think so, the `cartodb_id` could be referenced within the viz and it would be appended now too

PR #896, opened 2018-08-27
Add clusterCount expression

VictorVelarde:
### Context
There is an unused column in aggregations with the number of features in the cluster aggregation (`_cdb_feature_count`). This creates a new expression `clusterCount` to expose it when available, falling back to default value of 1 when it's not.

### Status

- [x] clusterCount() expression
- [x] basic tests
- [x] tests with aggregations
- [x] tests without aggr.
- [x] doc
- [x] CR 

### Example 1: with aggregation
image

### Example 2: without aggregation (default: 1)
image

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -9,6 +9,8 @@ const SAMPLE_ROWS = 1000;
 const MIN_FILTERING = 2000000;
 const REQUEST_GET_MAX_URL_LENGTH = 2048;
 
+export const CLUSTER_FEATURE_COUNT = '_cdb_feature_count';
+
 export default class Windshaft {
     constructor (source) {
         this._source = source;
@@ -45,11 +47,7 @@ export default class Windshaft {
         this._checkAcceptableMNS(MNS);
         const resolution = viz.resolution;
         const filtering = windshaftFiltering.getFiltering(viz, { exclusive: this._exclusive });
-        // Force to include `cartodb_id` in the MNS columns.
-        // TODO: revisit this request to Maps API
-        if (!MNS['cartodb_id']) {
-            MNS['cartodb_id'] = [{type: 'id'}];
-        }
+        this._forceIncludeCartodbId(MNS);
         if (this._needToInstantiate(MNS, resolution, filtering)) {
             const instantiationData = await this._repeatableInstantiate(MNS, resolution, filtering);
             this._updateStateAfterInstantiating(instantiationData);
@@ -57,14 +55,20 @@ export default class Windshaft {
         return this.metadata;
     }
 
+    _forceIncludeCartodbId (MNS) {
+        // Force to include `cartodb_id` in the MNS columns.
+        // TODO: revisit this request to Maps API
+        if (!MNS['cartodb_id']) {
+            MNS['cartodb_id'] = [{ type: 'id' }];
+        }
+    }
+
     requiresNewMetadata (viz) {
         const MNS = viz.getMinimumNeededSchema();
         this._checkAcceptableMNS(MNS);
         const resolution = viz.resolution;
         const filtering = windshaftFiltering.getFiltering(viz, { exclusive: this._exclusive });
-        if (!MNS['cartodb_id']) {
-            MNS['cartodb_id'] = [{type: 'id'}];
-        }
+        this._forceIncludeCartodbId(MNS);
         return this._needToInstantiate(MNS, resolution, filtering);
     }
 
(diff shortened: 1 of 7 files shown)


VictorVelarde:
This is ready for a CR. The new expression has been implemented for **windshaft**, with a fallback to 1 as @jgoizueta suggested. Does it make sense to implement this also in GeoJSON class?

src/client/windshaft.js
```diff
@@ -9,6 +9,8 @@ const SAMPLE_ROWS = 1000;
 const MIN_FILTERING = 2000000;
 const REQUEST_GET_MAX_URL_LENGTH = 2048;
 
+export const CLUSTER_FEATURE_COUNT = '_cdb_feature_count';
```
davidmanzanares: For consistency and to keep track of every ID and pattern like this, this should be on `schema.js`

src/renderer/viz/expressions/aggregation/cluster/ClusterCount.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @return {Number} Cluster feature count
+ *
+ * @example <caption>Use cluster count as width.</caption>
+ * const s = carto.expressions;
+ * const viz = new carto.Viz({
+ *   width: s.clusterCount() / 50
+ * });
+ *
+ * @example <caption>Use cluster cluster count as width. (String)</caption>
```
davidmanzanares: `cluster` is repeated twice. String examples should always match JS examples with the addition of ` (String)` since the dev-center uses that to put the String examples and the JS ones in two columns

src/renderer/viz/expressions/aggregation/cluster/ClusterCount.js
```diff
@@ -0,0 +1,52 @@
...
+ * @name clusterCount
+ * @function
+ * @api
+ */
+export default class ClusterCount extends BaseExpression {
+    constructor () {
+        checkMaxArguments(arguments, 0, 'clusterCount');
+        super({});
+        this.type = 'number';
+        this._expressionName = 'clusterCount';
```
davidmanzanares: Since this doesn't inherit from ClusterAggregation I think `_expressionName` is not needed

src/renderer/viz/expressions/aggregation/cluster/ClusterCount.js
```diff
@@ -0,0 +1,52 @@
...
+
+    _bindMetadata (metadata) {
+        super._bindMetadata(metadata);
+        this._hasClusterFeatureCount = metadata.properties[CLUSTER_FEATURE_COUNT] !== undefined;
+    }
+
+    _applyToShaderSource (getGLSLforProperty) {
+        return {
+            preface: '',
+            inline: `${this._hasClusterFeatureCount ? getGLSLforProperty(CLUSTER_FEATURE_COUNT) : '1.'}`
```
davidmanzanares: `inline: this._hasClusterFeatureCount ? getGLSLforProperty(CLUSTER_FEATURE_COUNT) : '1.'` is simpler

test/unit/renderer/viz/expressions/aggregation/clusterAggregation.test.js
```diff
@@ -60,5 +63,18 @@ describe('src/renderer/viz/expressions/clusterAggregation', () => {
...
+        it('clusterCount() should return fakeFeature._cdb_feature_count if defined', () => {
+            const actual = s.clusterCount().eval(fakeFeature);
+            expect(actual).toEqual(fakeFeature._cdb_feature_count);
+        });
+
+        it('clusterCount() should return 1 if _cdb_feature_count is undefined in the feature', () => {
+            const aFeature = Object.assign({}, fakeFeature);
+            delete aFeature._cdb_feature_count;
+
+            const actual = s.clusterCount().eval(aFeature);
```
davidmanzanares: Just passing `{}` should be ok and then you could remove the construction of `aFeature`

src/renderer/viz/expressions/aggregation/cluster/ClusterCount.js
VictorVelarde: Thx!, that's true, it is not used

src/renderer/viz/expressions/aggregation/cluster/ClusterCount.js
VictorVelarde: 👍 Good catch

test/unit/renderer/viz/expressions/aggregation/clusterAggregation.test.js
VictorVelarde: Nice simplification!

VictorVelarde:
We could merge this in, if @jgoizueta doesn't see the need to deal specifically with GeoJSON class

davidmanzanares:
Yes, I think this is ready to merge, GeoJSON doesn't work with cluster aggregation anyway so I think it's safe to merge.

:rocket:

PR #899, opened 2018-08-28
Allow ramp with numeric arrays with feature properties

davidmanzanares:
https://github.com/CartoDB/carto-vl/issues/690#issuecomment-416219549

See `examples/guides/interpolation/step-2.html` for understanding the new feature


src/renderer/viz/expressions/basic/array.js
```diff
@@ -51,12 +51,6 @@ export default class BaseArray extends BaseExpression {
 
         this.type = `${type}-array`;
         this.elems = elems;
-
```
VictorVelarde: This looks interesting. I think that will be useful, for example when features have several properties for registering the same phenomenon in different dates

src/renderer/viz/expressions/ramp.js
```diff
@@ -120,10 +122,24 @@ export default class Ramp extends BaseExpression {
     }
 
     eval (feature) {
-        this.palette = this._calcPaletteValues(this.palette);
+        const input = this.input.eval(feature);
+        if (this.palette.type === 'number-array') {
+            const m = (input - this.minKey) / (this.maxKey - this.minKey);
```
VictorVelarde: IMO it would be great to extract this if-case to its own function, to improve legibility, maybe something like `_evalNumberArrayWithMix` or a better name

src/renderer/viz/expressions/ramp.js
```diff
@@ -180,39 +196,76 @@ export default class Ramp extends BaseExpression {
...
 
-        let inline = `texture2D(texRamp${this._uid}, vec2((${input.inline}-keyMin${this._uid})/keyWidth${this._uid}, 0.5))`;
+        let inline = '';
+        let preface = `
+        uniform float keyMin${this._uid};
+        uniform float keyWidth${this._uid};
+        `;
+
+        let inputGLSL = `((${input.inline}-keyMin${this._uid})/keyWidth${this._uid})`;
         if (this.input.type === 'category' && this.input.isA(Property)) {
```
VictorVelarde: Same here, I'll go for some extract methods, such as `_applyToShaderSourceImage`, `_applyToShaderSourceNumberArray`... and alike, to improve understanding for people without knowledge on shaders like me

src/renderer/viz/expressions/basic/array.js
davidmanzanares: :confused: what do you mean?

src/renderer/viz/expressions/ramp.js
davidmanzanares: :+1:

src/renderer/viz/expressions/ramp.js
davidmanzanares: :+1:

src/renderer/viz/expressions/basic/array.js
VictorVelarde: Just that there could be cases in which a feature, for example a county, has three properties with population for 3 census dates (2000, 2005, 2010), and that now it could be used in an array, for some kind of visualizations. I think that non-normalized schema is quite common

davidmanzanares:
I've updated the PR. I made the proposed refactor on `eval` and some refactoring in `applyToShaderSource`, but in that one, I haven't extracted much of the functionality because it was accumulating things on `preface` and, even though that could be extracted I think it adds unnecessary boilerplate that complicates things even more. However, I made some changes and some clarifications to try to make it easier to understand (although it clearly isn't easy).

VictorVelarde:
@davidmanzanares thx! I really appreciate it

PR #901, opened 2018-08-28
Add Layer.viz getter

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/897

### DIFF
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -269,6 +269,20 @@ export default class Layer {
     }
 
     /**
+     * Viz attached to this layer.
+     *
+     * Calls to `blendToViz` and `update` wont' update the viz until those calls "commit",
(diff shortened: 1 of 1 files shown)


src/Layer.js
```diff
@@ -268,6 +268,20 @@ export default class Layer {
...
+     * Viz attached to this layer.
+     *
+     * Calls to `blendToViz` and `update` wont' update the viz until those calls "commit",
+     * having performed and completed all asynchronous necessary sanity checks.
+     *
+     * @returns {carto.Viz} - Viz object currently bound to the layer
+     * @memberof carto.Layer
+     * @api
+     */
+    get viz () {
```
VictorVelarde: LGTM. How do those calls "commit"? should we show that behaviour in a test?

PR #902, opened 2018-08-28
Add default export to index.js

VictorVelarde:
### Context
There is currently no default export in our library (*index.js*). This means that:
* `const carto = require('@carto/carto-vl');` works
* but `import carto from '@carto/carto-vl';` **fails**.

You can workaround it with `import * as carto from '@carto/carto-vl';` but it feels less natural, so we are adding the main object as default export.


VictorVelarde:
Keeping previous `export` and adding also `export default`, allows us to keep importing selectively (eg. `import { Viz, expressions as s } from '../../../src/index';`)

VictorVelarde:
This should be ready for a CR

PR #903, opened 2018-08-28
Improve Maps API error control and report

davidmanzanares:
Fix for the last 3 situations described on https://github.com/CartoDB/carto-vl/issues/210#issuecomment-416566274

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -321,7 +321,12 @@ export default class Windshaft {
         const response = await fetch(getMapRequest(conf, mapConfigAgg));
         const layergroup = await response.json();
         if (!response.ok) {
-            throw new Error(`Maps API error: ${JSON.stringify(layergroup)}`);
+            if (response.status === 401) {
+                throw new Error(`Unauthorized access to Maps API: invalid combination of user('${this._source._username}') and apiKey('${this._source._apiKey}')`);
+            } else if (response.status === 403) {
+                throw new Error(`Unauthorized access to dataset: the provided apiKey('${this._source._apiKey}') doesn't provide access to the requested datasets and tables`);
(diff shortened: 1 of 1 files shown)


src/client/windshaft.js
```diff
@@ -321,7 +321,12 @@ export default class Windshaft {
         const response = await fetch(getMapRequest(conf, mapConfigAgg));
         const layergroup = await response.json();
         if (!response.ok) {
-            throw new Error(`Maps API error: ${JSON.stringify(layergroup)}`);
+            if (response.status === 401) {
```
VictorVelarde: 👍 Maybe we could use a switch, if that feels a little bit clearer for you:
```
let message;
switch (response.status) {
    case 401:
        message = `Unauthorized access to Maps API: invalid combination of user('${this._source._username}') and apiKey('${this._source._apiKey}')`
        break;
    case 403:
        message = `Unauthorized access to dataset: the provided apiKey('${this._source._apiKey}') doesn't provide access to the requested datasets and tables`
        break;
    default:
        message = `SQL errors: ${JSON.stringify(layergroup.errors)}`;
} 
throw new Error(message);
```

davidmanzanares: To be honest, do you think it's necessary? In this case, I think both options are equally good, but changing the code means to manually test the three cases again :(

VictorVelarde: Nop, it was just a style comment 😄

VictorVelarde: Out of curiosity I looked if this could impact performance. Short answer, with a few options there are no much differences. Interesting 2 min reading here: https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04s02.html

PR #908, opened 2018-08-29
Change zoom semantics

davidmanzanares:
To finally complete https://github.com/CartoDB/carto-vl/issues/690


examples/examples.json
```diff
@@ -63,10 +63,6 @@
                     "title": "Viewport based styling",
                     "file": "styling/viewport.html"
                 },
-                {
```
VictorVelarde: I see that we remove zoom example, why?. Should we have another one with `zoomRange`?

src/renderer/viz/expressions/blend.js
```diff
@@ -15,15 +15,15 @@ import BaseExpression from './base';
  * const viz = new carto.Viz({
  *   width: s.blend(3,
  *                  s.prop('dn'),
- *                  s.linear(s.zoom(), s.pow(2, 10), s.pow(2, 14))
+ *                  s.linear(s.zoom(), 10, 14))
```
VictorVelarde: Now it looks nicer, without the s.pow...

src/renderer/viz/expressions/zoom.js
```diff
@@ -3,19 +3,19 @@ import { number } from '../expressions';
 import { checkMaxArguments } from './utils';
 
 /**
- * Get the current zoom level. Multiplying by zoom() makes features constant in real-world space respect their size at zoom level 0.
+ * Get the current zoom level.
  *
  * @return {Number}
  *
- * @example <caption>Show constant width in zoom.</caption>
+ * @example <caption>Only show feature at zoom levels les than 7.</caption>
```
VictorVelarde: typo in less

test/unit/renderer/viz/expressions/ramp.test.js
```diff
@@ -975,7 +975,7 @@ describe('src/renderer/viz/expressions/ramp', () => {
                 const r = ramp(zoomrange([3, 9]), [100, 200]);
                 r._bindMetadata();
                 const fakeGL = {uniform1f: () => {}, uniform1i: () => {}};
-                const fakeDrawMetadata = {zoom: 1000.0};
+                const fakeDrawMetadata = {zoomLevel: 100};
```
VictorVelarde: Why this change in magnitude?

examples/examples.json
davidmanzanares: The zoom example was actually being similar to `scaled` not to `zoomRange` actually. We could add an example, though

src/Layer.js
```diff
@@ -399,7 +399,9 @@ export default class Layer {
     _paintLayer () {
         if (this._viz && this._viz.colorShader) {
             this._renderLayer.setViz(this._viz);
-            this.renderer.renderLayer(this._renderLayer);
+            this.renderer.renderLayer(this._renderLayer, {
+                zoomLevel: this.map.getZoom()
+            });
             if (this.isAnimated() || this._fireUpdateOnNextRender || !util.isSetsEqual(this._oldDataframes, new Set(this._renderLayer.getActiveDataframes()))) {
```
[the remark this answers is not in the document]
davidmanzanares: I agree with the second part `const framesHaveChanged = !util.isSetsEqual(this._oldDataframes, this._getActiveDataFrames()))`. But not with the first one (`_getActiveDataFrames() {...`), I don't agree with that because the only real thing that the method is doing (creating a set) is being hidden from the caller and it is not an implementation detail, I also don't think that extracting that behavior (of 1 line) is worth of 2 extra lines

src/renderer/viz/expressions/zoom.js
davidmanzanares: :+1:

test/unit/renderer/viz/expressions/ramp.test.js
davidmanzanares: With `1000.0` the test pass too. I just changed it for aesthetics and because since now the zoomLevel is in other range (logarithmic) using such a big value seemed to me like a sign of not using the logarithmic level.

src/Layer.js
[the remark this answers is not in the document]
VictorVelarde: 👍 I just saw it was being used twice, in the if and to get the old frames, but fair enough!

test/unit/renderer/viz/expressions/ramp.test.js
VictorVelarde: Oh I see, thx for the explanation 👍

davidmanzanares:
Updated with the suggestions

examples/examples.json
elenatorro: It would be nice to add this example ☝️ ✨

davidmanzanares:
Actually, I already added the example, see the commit https://github.com/CartoDB/carto-vl/pull/908/commits/c2006312aa08dc9f5948e5d8ca8e4d4169ee12c9

Thank you! :slightly_smiling_face:

PR #909, opened 2018-08-29
Add support for Mean Standard Deviation classification method

VictorVelarde:
### Context
This PR includes Mean-Standard Deviation classification method, in two modes:
- [x] **globalStandardDev**
- [x] **viewportStandardDev**

### Tasks
- [x] global - implementation
- [x] global - doc & tests
- [x] viewport - implementation
- [x] viewport - doc & tests
- [x] render tests (to be locally prepared in linux)
- [x] performance optimizations (map, reduce, math.pow...)
- [x] CR

### Status
After reading the cartography reference that @makella sent us (thx a lot!), I have a working implementation for the global & viewport options.

Considering the parameters usually required in GIS software, I've prepared this interface, with an optional parameter:
* `globalStandardDev($field, 5)` --> being 5 the number of buckets
* `globalStandardDev($field, 5, 0.5)` --> being 5 the number of buckets and using 1/2 standard deviation instead of 1 (the default value)
* ... and the same options are available for viewport with `viewportStandardDev`.

This is the result for a global example:
image

### Questions

Before we can move forward with the CR, I would like to confirm some aspects with you:
- [x] are you ok with the expression **name & params**?
- [x] we use standard deviation for *population formula*. Should we also consider **sample** option? (https://www.thoughtco.com/population-vs-sample-standard-deviations-3126372)
- [x] **buckets logic**.  
   I have checked the breakpoints in QGIS3, and the calculations are pretty similar. Notably in QGIS3 the number of classes is not honored when it is odd!!. See for example this case, where 3 classes are required but 4 are returned.
image
We return 3 if 3 are required, and following the authors in your reference (could you please Mamata send me the full reference?), we also double-sized the central class when odd. So our breakpoints for 3 classes, using the very same sample, are these —> [0.792174872340067, 4.207825127659933], that correspond to:
    * Class 1: < avg - 1std
    * Class 2: [avg - 1std, avg + 1std] << this is the key difference
    * Class 3: > avg + 1std


VictorVelarde:
@davidmanzanares @rochoa what are your thoughts on this? 
My 2 cts. go to using the size of the buckets in STD units, but I think that @makella should enlighten us

davidmanzanares:
Yes, as you said, I think we could wait for @makella here

VictorVelarde:
hi @makella and @davidmanzanares I have updated the status 🔝 and added 3 questions, to move forward.

src/renderer/viz/expressions/classification/GlobalMeanStandardDev.js
```diff
@@ -0,0 +1,103 @@
...
+    let breaks;
+    let over = [];
+    let under = [];
+    const isEven = buckets % 2 === 0;
+    let factor = isEven ? 0.0 : 1.0; // if odd, central class is double sized
+    do {
+        const step = factor * (stDev * classSize);
+        over.push(avg + step);
+        under.push(avg - step);
+        breaks = [...new Set(over.concat(under))];
```
davidmanzanares: I don't understand this line. I guess that using a set is to remove duplicates, but I don't see why elements of `over` and elements of `under` may be equal

davidmanzanares: Ok, now I see that `step` will be `0` in the first iteration when `isEven`

src/renderer/viz/expressions/stats.js
```diff
@@ -0,0 +1,29 @@
...
+export const variance = (values, average) => values.map(value => Math.pow(value - average, 2));
+
+/**
+ * Calculate Standard Deviation (STD), using population deviation formula
+ *
+ * @param {Number[]} values
+ * @returns {Number} - standard deviation
+ */
+export const standardDeviation = (values) => {
+    const avg = average(values);
```
davidmanzanares: The viewport version will need to call this function per frame, and `values` can be a large array. This should probably be optimized, but first, it should be profiled.

In particular, I'm suspicious of the performance of `map`, `reduce` and `Math.pow`. Using an old-fashioned `for (let i=0 ...` and `x*x` instead of `Math.pow` will probably improve things.

test/unit/renderer/viz/expressions/classification/classifier.test.js
```diff
@@ -0,0 +1,265 @@
...
+    viewportEqIntervals,
+    viewportMeanStandardDev
+} from '../../../../../../src/renderer/viz/expressions';
+
+import Metadata from '../../../../../../src/renderer/Metadata';
+
+describe('src/renderer/viz/expressions/classifier', () => {
+    describe('error control', () => {
+        describe('global', () => {
+            validateStaticTypeErrors('globalQuantiles', []);
```
davidmanzanares: With one my last PRs this is not going to work anymore, but it's easy to fix, just use the function `validateTypeErrors` after merging with `master`

src/renderer/viz/expressions/classification/GlobalMeanStandardDev.js
VictorVelarde: That's right, that is the only corner case

src/renderer/viz/expressions/stats.js
VictorVelarde: In the viewport version, I changed it to use its own private methods, to calculate average and std from histogram frequencies. But I agree that further optimizations could be applied. I'm gonna add another task in this PR. Thx!

test/unit/renderer/viz/expressions/classification/classifier.test.js
VictorVelarde: Ok @davidmanzanares  so all `validateStaticTypeErrors` should be replaced with `validateTypeErrors`. 
And what about:
* `validateDynamicTypeErrors`?
* `validateCompileTypeError`?

davidmanzanares: they should too, you can look at https://github.com/CartoDB/carto-vl/pull/917/files

makella:
a quick update here... @VictorVelarde and I spoke about his three questions and for #'s 2 and 3, we will consult a few other resources before making a final decision.

makella:
@VictorVelarde so I don't have a strong opinion about the Population vs. Sample methodology... the data that people will visualize will most likely be ok calculated using the Population method since most will be things that are based on a population like crime, votes, etc. and other data like that of the Census, even though it is from a sample, is already taken into account from the organization itself when they calculate the attributes in their data. I haven't been able to find any specific links to which method other tools use. 

With regards to question 3, I'm still doing some searching... 

Was there a mention of a full reference in the chapter? If so, let me know and I can see if I can dig that up.

VictorVelarde:
Ok @makella then I think we're going to move on using **Population formula**. 

Regarding to last question, buckets logic, there is no extra-reference in the chapter. They just say (page 63) *"For the five-class map shown in Figure 4.2C. the two middle classes were combined."*
and you can see that in the Table 4.3 the class nº 3 they use has a double size...

All in all, if there are no more references, maybe we should reconsider and just always keep the same class size, for clarity

VictorVelarde:
These are the results of some performance tests, regarding to stats calculations:

| Function  | Initial version   | Revised | Result
|---	      |---        |---	    |---
| average 	| (reduce)  | (for)   | (for) is 76% faster [jsperf](https://jsperf.com/1-average/1)
| variance 	|       	  |       	|

makella:
> that's why it is better if the number of classes is odd, so I think we could keep our current logic, if you are ok with it.

All good for me! Thanks for such a clear and thoughtful investigation :)

VictorVelarde:
Then, after preparing the image references, this should be ready to be merged in

VictorVelarde:
@davidmanzanares This is ready to be merged, after you create the new image references

src/renderer/viz/expressions/classification/GlobalMeanStandardDev.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+export default class GlobalMeanStandardDev extends Classifier {
+    constructor (input, buckets, classSize = 1.0) {
+        checkMaxArguments(arguments, 3, 'globalMeanStandardDev');
+        checkInstance('globalMeanStandardDev', 'input', 0, Property, input && (input.property || input));
+        checkNumber('globalMeanStandardDev', 'buckets', 1, buckets);
+        checkNumber('globalMeanStandardDev', 'classSize', 2, classSize);
+
+        super({ input }, buckets);
+        this._classSize = classSize;
```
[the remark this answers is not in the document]
davidmanzanares: I guess that `buckets` has the same problem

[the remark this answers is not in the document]
VictorVelarde: Good catch @davidmanzanares ! I'll add buckets validation >=2 to *Classifier.js*, its parent.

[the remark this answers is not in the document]
VictorVelarde: Regarding to the name, `MeanStandardDev`, we could change it to just `StandardDev` again, as it was in the beginning. What do you think?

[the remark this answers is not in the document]
davidmanzanares: Yes, I think short is better

PR #910, opened 2018-08-29
Add support for Jenks Natural Breaks classification method

VictorVelarde:
### Context
This PR will include Jenks classification method, in two flavours:
- [ ] **viewportJenks**
- [ ] **globalJenks**

### Tasks
- [ ] global - implementation
- [ ] global - doc & example
- [ ] viewport - implementation
- [ ] viewport - doc & example
- [ ] CR

### DIFF
--- a/src/renderer/viz/expressions/classification/GlobalJenks.js
+++ b/src/renderer/viz/expressions/classification/GlobalJenks.js
@@ -1,0 +1,46 @@
+import Classifier from './Classifier';
+import Property from '../basic/property';
+import { checkNumber, checkInstance, checkType, checkExpression, checkMaxArguments } from '../utils';
+
+/**
+ * Classify `input` by using the Jenks method with `n` buckets.
+ *
+ * It will classify the input based on the entire dataset without filtering by viewport or by `filter`.
+ *
+ * @param {Number} input - The input expression to classify
+ * @param {number} n - Number of buckets
+ * @return {Category}
+ *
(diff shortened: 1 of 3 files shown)


makella:
I will write more in another comment, but Jenks (while tough to implement) is very very standard and when there is a clustered distribution, this is what we want to recommend ppl to use. I'm concerned if we fully remove since it is so common not only in ArcGIS but across every cartography/GIS textbook out there that people use in University and in their professional lives.

There are different flavors of Jenks so I'm not sure if the ckmeans is in line with one of those, but that could be the next step in the process. 

If you guys want, we can read this chapter and then discuss in more detail... (I'm inclined to go w/ cartography literature vs. Tom MacWright)

https://drive.google.com/drive/u/0/folders/0B9oDwtYnQbuPQnhTbjV4NjVQRzA

makella:
For classification, overall, @andy-esch and I put together this document a while back which was to kick-off a project to fix existing classification methods in Builder. As you will see, we noted that we need to find an alternative and/or fix our implementation of Jenks. 

Also, you will note our comments for standard deviation (#909), etc. in the document.

https://docs.google.com/document/d/1CB_MQasY-4YeMWcNW4VlvFkXbXA-CktcsBLW0g8HwPI/edit#heading=h.eilx4ykmr119

makella:
sorry david and victor about the link to the chapter... here it is: https://drive.google.com/file/d/0B9oDwtYnQbuPTVBkc2FPclY2bVV5OV9sTFpfNmdlbWlJZ1FJ/view

Issue #772, opened 2018-07-26
Optimize `decodeProperty`


davidmanzanares:
It's one of the main hot spots in many apps. I think that the performance could be improved if we re-used the feature object from the MVT decoder for `Dataframe.getFeature` and if didn't put every property on `Float32Array`s unless they are needed (lazy).

davidmanzanares:
The laziness idea was tested and was found pretty bad in some common cases.

However, the Web Worker PR has addressed this issue.

PR #912, opened 2018-08-30
Web Worker for MVT decoding

davidmanzanares:
Addresses https://github.com/CartoDB/carto-vl/issues/756 and https://github.com/CartoDB/carto-vl/issues/772 by using a Web Worker for the decoding of the MVT tiles

- [x] Working PoC
- [x] Fix for all geometry types
- [x] Improve hacky PoC code
  - [x] Remove tileclient code from worker
  - [x] Create a WorkerDataframe class to avoid code repetition and to avoid bloating the original Dataframe
  - [x] Geom decoders should use arraybuffers
- [x] Fix metadata modified by worker issue
- [x] Fix aggs
- [x] Fix Windshaft Date issue
- [x] Fix tests
- [x] Check and minify if possible shared code between Worker and Main (because the code is copied and the bundle size increase)
- [x] Check bundle size increase (right now it went from 71KB to 76KB, minified and gzipped)
- [x] Check everything is ok and faster in a real application
  - [x] Fix metadata race condition

## Bundleing

I've used webpack for geneting the workers bundle and I used the `inline` option to include the worker code as a `Blob` in the bundle, to only need one bundle.

Users of the library should use the generated bundle (which could be used to generate another bundle with user code). Users shouldn't try to create a bundle from the sources (I think the current package.json configuration already aims to the bundle) because all the WebWorker configuration (and the GLSL configuration too) should be repeated in that case in the users webpack configuration file.

#### Don't forget to update the dependencies with `yarn`


src/sources/MVTWorker.js
```diff
@@ -0,0 +1,237 @@
...
+    2: geometryTypes.LINE,
+    3: geometryTypes.POLYGON
+};
+
+export class MVTWorker {
+    // Worker API
+    onmessage (event) {
+        this.processEvent(event).then(message => {
+            message.dataframe.geom = null;
+            const transferables = [];
```
davidmanzanares: TODO: extract

davidmanzanares: and move

src/sources/MVTWorkers.worker.js
```diff
@@ -0,0 +1,11 @@
+import { MVTWorker } from './MVTWorker';
+import { WindshaftWorker } from '../client/WindshaftWorker';
+
+const workers = {
```
davidmanzanares: TODO explain

src/client/WindshaftWorker.js
```diff
@@ -0,0 +1,29 @@
...
+    decodeProperty (metadata, propertyName, propertyValue) {
+        const basename = schema.column.getBase(propertyName);
+        const column = metadata.properties[basename];
+        if (!column) {
+            return;
+        }
+        switch (column.type) {
+            case 'date':
+            {
+                const d = new Date();
```
VictorVelarde: Maybe these six lines could be packed in a little self-explaining function... and could also use destructuring with `const {min, max} = column`;

src/renderer/decoder/polygonDecoder.js
```diff
@@ -57,9 +57,19 @@ export function decodePolygon (geometry) {
         breakpoints.push(index);
     }
 
+    const verticesArrayBuffer = new ArrayBuffer(4 * index);
```
VictorVelarde: This seems to be exactly the same code as in `lineDecoder.js`, so we have an opportunity to eliminate duplicity, importing from there.

src/sources/MVT.js
```diff
@@ -135,184 +141,36 @@ export default class MVT extends Base {
...
-                if (x < -1 || x >= 1 || y < -1 || y >= 1) {
-                    continue;
+        return this._tileClient.requestData(zoom, viewport, (x, y, z, url) => {
+            return new Promise(resolve => {
+                // Relative URLs don't work inside the Web Worker
+                if (url[0] === '.') {
+                    let parts = window.location.pathname.split('/');
+                    parts.pop();
+                    const path = parts.join('/');
+                    url = window.location.protocol + '//' + window.location.host + '/' + path + '/' + url;
```
VictorVelarde: Having ES6 we can use template literals, like:
``` 
url = `${window.location.protocol}//${window.location.host}/${path}/${url}`
```

src/client/WindshaftWorker.js
davidmanzanares: :+1:

src/sources/MVT.js
davidmanzanares: :+1:

src/renderer/decoder/polygonDecoder.js
davidmanzanares: Yes, it's the same now... But I'm not sure that extracting this would generate a good abstraction: an easy to understand function.

src/renderer/DummyDataframe.js
```diff
@@ -0,0 +1,21 @@
...
+        this.type = type;
+        this.decodedGeom = decodeGeom(type, geom);
+        this.numVertex = type === 'point'
+            ? size * 3
+            : this.decodedGeom.vertices.length / 2;
+        this.numFeatures = type === 'point' ? size : this.decodedGeom.breakpoints.length || this.numVertex;
+        this.propertyTex = [];
+        this.metadata = metadata;
+        this.propertyCount = 0;
+        this._aabb = computeAABB(geom, type);
```
elenatorro: Awesome refactor!

src/sources/MVT.js
```diff
@@ -81,6 +61,7 @@ const MVT_TO_CARTO_TYPES = {
  *
  * @api
  */
+import Worker from './MVTWorkers.worker';
```
elenatorro: I would put the import below the rest of import declarations.

davidmanzanares: :+1:

src/renderer/DummyDataframe.js
davidmanzanares: thx!

src/renderer/decoder/polygonDecoder.js
VictorVelarde: I aggree, that might be coincidental. I was just thinking on exposing the linear one and consuming it from the polygon one, but it may be "killing flies with cannons" 😄

src/client/WindshaftWorker.js
```diff
@@ -0,0 +1,30 @@
...
+                return metadata.categorizeString(basename, propertyValue);
+            case 'number':
+                return propertyValue;
+            default:
+                throw new Error(`Windshaft MVT decoding error. Feature property value of type '${typeof propertyValue}' cannot be decoded.`);
+        }
+    }
+}
+
+function decodeDate (column, propertyValue) {
```
jgoizueta: nice refactor. all this is actually better now, being encapsulated in a subclass

PR #913, opened 2018-08-30
Add CORS documentation

davidmanzanares:
See https://github.com/CartoDB/support/issues/1729#issuecomment-417270418


Algunenano:
Could you please also add the recommendation to test with the built-in images if the images are not working to discard issues with the code?

davidmanzanares:
:+1:

Issue #904, opened 2018-08-28
Enable String API in `blendTo`


rochoa:
Once you have defined your Viz object with the String API, there is no good alternative to keep working with the String API apart from using the `layer.blendToViz` mechanism. That leads to some confusion, as seen in https://github.com/CartoDB/carto-vl/issues/892.

Users of the String API should be able to continue using the String API after defining their initial Viz object. Ideally, they should use the same `blendTo` API but this should accept String API input.

With:

```js
const viz = new carto.Viz(`
    color: yellow
`);
```

Later, instead of doing:

```js
viz.color.blendTo(carto.expressions.namedColor('red'))
```

You should be able to do:

```js
viz.color.blendTo('red');
```

PR #915, opened 2018-08-31
Blendto strings

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/904

And fix an unreported bug in which ViewportQuantiles throw an error while loading (afterwards things were ok).

Also, remove some unneeded code from Layer.


src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
```diff
@@ -84,8 +84,9 @@ export default class ViewportHistogram extends BaseExpression {
     }
 
     _resetViewportAgg (metadata) {
-        metadata = metadata || [];
-        this._bindMetadata(metadata);
+        if (metadata) {
```
VictorVelarde: Good catch!

src/Layer.js
```diff
@@ -295,7 +295,6 @@ export default class Layer {
         this.map.setLayoutProperty(this.id, 'visibility', 'visible');
         this._visible = true;
         this.requestData();
-        this._fire('updated');
```
VictorVelarde: Why don't we need to raise the event anymore?

davidmanzanares: I think we never needed it, I removed it because, while looking for something else I realized that. The reason is not needed is because the change of visibility is going to trigger a repaint and the repaint is going to trigger the `updated` event

PR #916, opened 2018-09-03
Fix animation integration with MGL

davidmanzanares:
As reported by @skgsergio animation is broken in some cases. He has an actual reproducible example (contact him or me for the code).

After debugging we discovered that the cause was that the last layer was not animated and it was overriding the "animatedness" of the other layers. This seems to have been introduced in the new MGL integration.

Fortunately, the MGL team already thought about this and we can remove the old hacky `repaint` with the new method `triggerRepaint` (see the upstream PR for details https://github.com/mapbox/mapbox-gl-js/pull/7039).

We've discussed the need for regression tests, but since this happened because of a very unusual change (the integration one), we decided to avoid them. Anyway, we agreed that animation, in general, is lacking tests.

### DIFF
--- a/src/maps/Map.js
+++ b/src/maps/Map.js
@@ -195,4 +195,7 @@ export default class Map {
         m[15] = 1;
         return m;
     }
+
+    triggerRepaint () {
+    }
 }
--- a/src/maps/Map.js
+++ b/src/maps/Map.js
@@ -29,7 +29,6 @@ export default class Map {
 
         this._layers = new Set();
         this._hiddenLayers = new Set();
-        this._repaint = true;
         this._canvas = this._createCanvas();
         this._container.appendChild(this._canvas);
         this._gl = this._canvas.getContext('webgl');
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -207,7 +207,7 @@ export default class Layer {
         this._viz = viz;
         viz.onChange(this._vizChanged.bind(this));
         this._compileShaders(viz, metadata);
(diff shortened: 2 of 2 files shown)


src/maps/Map.js
```diff
@@ -196,4 +195,7 @@ export default class Map {
         m[15] = 1;
         return m;
     }
+
+    triggerRepaint () {
```
elenatorro: Should we add here `window.requestAnimationFrame(this._update.bind(this));` ? Or add a comment explaining we need it to avoid problems when calling `this.map.triggerRepaint();` in the Layer is not a mgl map? (not a big fan of comments, but leaving it empty looks weird to me). Just a minor comment.

davidmanzanares: Well, as you said, this needs to do something (the code you posted seems good actually) to make animations work. The reason I left it unimplemented is that I think that `Map` already didn't work with animation (so it wasn't needed to implement it).

I agree with you in that it is weird. But I think the real problem is that we are not clarifying the whole purpose of the stand-alone `Map`. We are only using it for some tests (and I think it wouldn't be very problematic to replace it by MGL in every test, we would need to check times though), we are exposing it on the `carto.` namespace without an underscore (although it is not exposed in the public documentation). On the other hand, it is good to have something that forces us to don't couple CVL to MGL.

elenatorro: Yes, I agree that is good to have it, and it is really useful for testing. I think it would be great to explain what `Map` is, but I am not sure if this is necessary at this point.

Issue #819, opened 2018-08-07
Offset with `symbol:` doesn't work


davidmanzanares:

davidmanzanares:
Fixed by #919

Issue #822, opened 2018-08-07
Create example for offset properties


makella:
Create an example showing how to use `offset: placement()`. Dependant on a couple of outstanding issues:

https://github.com/CartoDB/carto-vl/issues/819
https://github.com/CartoDB/carto-vl/issues/361

**Update**

`offset` property is going to be renamed to `transform`, more info in this issue: https://github.com/CartoDB/carto-vl/issues/934

elenatorro:
Closing in favor of https://github.com/CartoDB/carto-vl/issues/969

Issue #361, opened 2018-04-30
Enable location to be modified by attributes 


stuartlynn:
For point datasets, it would be fantastic to be able to be able to modify where a point is displayed via other attributes. So something like 

color: rgb(255,0255)
xPixOffset: linear($value, 2,5)*10
yPixOffset: linear($value, 2,5) +2

Or 

color: rgb(255,0255)
LatOffset: linear($value, 2,5)*10
LngOffset: linear($value, 2,5) +2

Or 

color: rgb(255,0255)
x: linear($value, 2,5)*10
y: linear($value, 2,5) +2

what would this enable:

1) For datasets with many coincident points, the availability to stack those points like stacked chips 
2) For datasets with many coincident points, the availability to gitter points apart 
3) Torque for points where we can calculate the path of a point from an attribute but don't want to generate a table of all the intermediate points. Just specify a start and end point on the row and interpolate between them.
4) Other weird effects.

davidmanzanares:
Hi,

On our weekly meeting of the last Friday, we talked about this issue. We would like to know your point of view @makella regarding the priority of this.

zingbot:
I think this is great and will definitely give us an advantage in the market. The ability of CartoCSS to precisely place points around a marker was really valuable. Here's an example of a design for a parking garage parking space dashboard, centered around a point.

https://team.carto.com/u/jferzoco/builder/724abc51-4104-490b-a608-5ce6cb19c953/embed

This kind of work is in demand from lots of users.

makella:
@davidmanzanares and I were discussing the offset properties available and the syntax that we could use to be more specific. 

Here is an example:
`transform: translate(x,y),rotate(degrees)`

PR #919, opened 2018-09-03
Offset symbol

davidmanzanares:
Fix for https://github.com/CartoDB/carto-vl/issues/819


elenatorro:
It works great ✨ I know we are adding offset examples here https://github.com/CartoDB/carto-vl/issues/822, but I think it would be great to add an example in `placement.js` showing the difference between `symbolPlacement` and `offset`, or at least that you can combine both. What do you think?

davidmanzanares:
Ok, but I wouldn't publish it since I think that the example should be with an explanation in one guide (and the guide is out of the scope for this).

elenatorro:
I agree it is out of the scope, but I mean to add an example using placement with offset, so the user can see it does not need to be used only with symbolPlacement.

elenatorro:
I did not mean an html example, but a code example in `placement.js` with JS API / String API.

davidmanzanares:
To be honest, I wouldn't add an example with `offset` in the placement reference. I think they don't have special meaning when combined, they just (now) play well with each other, or better said, they don't interfere with each other.

elenatorro:
Ok, not combined but at least an example using only placement with offset, because we do not have any code example showing this, and it seems that placement is only available for `symbolPlacement` while it can be used with `offset`. I think it is a useful and simple example, but this is just a personal opinion, so if you think it is not necessary, then it is ok. The rest LGTM.

davidmanzanares:
Oh, ok, now I understand you. I was confusing "placement" with "symbolPlacement". Yes, we totally miss an example with `offset` (which needs a "placement").

In fact, there exist an issue that deprecates placement for offset (https://github.com/CartoDB/carto-vl/issues/361) and introduces "translate" and "rotate".

I think we should tackle that one.

elenatorro:
LGTM since we do not need to add documentation to offset because we are going to add an equivalent `transform` property and `translate` expression.

Issue #743, opened 2018-07-19
Improve error list and refactor


elenatorro:
## Description

We currently have an [error list](https://github.com/CartoDB/carto-vl/blob/master/src/errors/error-list.js) file.

It would be great to update this list with the errors we have added lately, in order to have an structured place with all the errors we have. We were asked once for a list of errors in Builder, so I think it is a good practice to maintain and use that list. It is useful also for testing, cause if we change an error message we do not need to change in every place it is called.

PR #924, opened 2018-09-04
Use common error list

VictorVelarde:
### Related to https://github.com/CartoDB/carto-vl/issues/743

[UPDATED 2018-09-14]

### Context
We want to improve how we manage the errors. But we have also changed our approach here.

### a) 'error-list'
Having a common list with all the errors seemed pretty reasonable. There was already one, but multiple errors had been added without including them there. Tasks were:
- [x] carto-error update + specs
- [x] Include errors in common list for: windshaft, Interactivity, Layer... and many more classes

But this has been reviewed (first part of this PR), an commits were reverted, as there are several issues with this approach.

### b) New simpler approach
- No error-list
- No regExp to manage messages
- More custom error types: CartoValidationError, CartoParsingError... 

See PR from this [point 👇 ](https://github.com/CartoDB/carto-vl/pull/924#issuecomment-420243528)


elenatorro:
I failed to think through the cost implications of this issue, and it is true that following this approach is error-prone :(

I still think we need to keep track of the errors we have somehow, but for sure it needs to be easy to maintain and have a minimum cost. I like the idea of having specific error classes.

For me, the main idea of this issue is to have a simple manual list of error constants, independent of the built-in error message, to identify where the error takes place, for instance:

* We currently have:
`src/client/windshaft.js`

```js
throw new Error(`Incompatible combination of cluster aggregation usages (${
    JSON.stringify(usages.filter(x => x.type === 'aggregated'))
}) with unaggregated usage for property '${propertyName}'`);
```

* We could have:

**Option 1**
File (example): `src/errors/list.js`

```js
export default const Error = {
  Windshaft: {
    AGGREGATION_USSAGES: 'Incompatible combination of cluster aggregation usages'
  }
}
```

`src/client/windshaft.js`
```js
throw new Error(`${Error.Windshaft.AGGREGATION_USSAGES}: (${
    JSON.stringify(usages.filter(x => x.type === 'aggregated'))
}) with unaggregated usage for property '${propertyName}'`);
```

**Option 2**
File (example): `src/errors/list.js`

```js
export default const Error = {
  Windshaft: {
    AGGREGATION_USSAGES: 'Aggregation Ussages'
  }
}
```

`src/client/windshaft.js`
```js
throw new Error(`${Error.Windshaft.AGGREGATION_USSAGES}: Incompatible combination of cluster aggregation usages (${
    JSON.stringify(usages.filter(x => x.type === 'aggregated'))
}) with unaggregated usage for property '${propertyName}'`);
```

However, I think that if this needs further discussion, we could address this after the public release.

rochoa:
✨ I prefer that approach than the current one. 

Now, about my proposal of extracting the errors from the code itself. I just did a quick test with [`espree`](https://github.com/eslint/espree), and it, obviously, requires much more work than the 20 minutes I spent looking for the right library and writing a crappy script, but I think is feasible.

@elenatorro, from your example with `Windshaft.js`, I did a small change for demo purposes:

```diff
diff --git a/src/client/windshaft.js b/src/client/windshaft.js
index 5e4d666e..f141e666 100644
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -77,9 +77,9 @@ export default class Windshaft {
             const aggregatedUsage = usages.some(x => x.type === 'aggregated');
             const unAggregatedUsage = usages.some(x => x.type === 'unaggregated');
             if (aggregatedUsage && unAggregatedUsage) {
-                throw new Error(`Incompatible combination of cluster aggregation usages (${
-                    JSON.stringify(usages.filter(x => x.type === 'aggregated'))
-                }) with unaggregated usage for property '${propertyName}'`);
+                const aggregatedUsages = JSON.stringify(usages.filter(x => x.type === 'aggregated'));
+                // We could use annotations via commented code to explain, for instance, the variables in the template.
+                throw new MapsAPIError(`Incompatible combination of cluster aggregation usages (${aggregatedUsages}) with unaggregated usage for property '${propertyName}'`);
             }
         });
     }
```

Using `espree` with the following config:

```js
var ast = espree.parse(code, {
    attachComment: true,
    ecmaVersion: 8,
    sourceType: 'module',
});
```

Within the AST, I get:

```json
{
            "type": "ThrowStatement",
            "start": 3244,
            "end": 3402,
            "range": [
                3244,
                3402
            ],
            "argument": {
                "type": "NewExpression",
                "start": 3250,
                "end": 3401,
                "range": [
                    3250,
                    3401
                ],
                "callee": {
                    "type": "Identifier",
                    "start": 3254,
                    "end": 3266,
                    "range": [
                        3254,
                        3266
                    ],
                    "name": "MapsAPIError"
                },
                "arguments": [
                    {
                        "type": "TemplateLiteral",
                        "start": 3267,
                        "end": 3400,
                        "range": [
                            3267,
                            3400
                        ],
                        "expressions": [
                            {
                                "type": "Identifier",
                                "start": 3326,
                                "end": 3342,
                                "range": [
                                    3326,
                                    3342
                                ],
                                "name": "aggregatedUsages"
                            },
                            {
                                "type": "Identifier",
                                "start": 3385,
                                "end": 3397,
                                "range": [
                                    3385,
                                    3397
                                ],
                                "name": "propertyName"
                            }
                        ],
                        "quasis": [
                            {
                                "type": "TemplateElement",
                                "start": 3268,
                                "end": 3324,
                                "range": [
                                    3267,
                                    3326
                                ],
                                "value": {
                                    "raw": "Incompatible combination of cluster aggregation usages (",
                                    "cooked": "Incompatible combination of cluster aggregation usages ("
                                },
                                "tail": false
                            },
                            {
                                "type": "TemplateElement",
                                "start": 3343,
                                "end": 3383,
                                "range": [
                                    3342,
                                    3385
                                ],
                                "value": {
                                    "raw": ") with unaggregated usage for property '",
                                    "cooked": ") with unaggregated usage for property '"
                                },
                                "tail": false
                            },
                            {
                                "type": "TemplateElement",
                                "start": 3398,
                                "end": 3399,
                                "range": [
                                    3397,
                                    3400
                                ],
                                "value": {
                                    "raw": "'",
                                    "cooked": "'"
                                },
                                "tail": true
                            }
                        ]
                    }
                ]
            },
            "leadingComments": [
                {
                    "type": "Line",
                    "value": " We could use annotations via commented code to explain, for instance, the variables in the template.",
                    "start": 3124,
                    "end": 3227,
                    "range": [
                        3124,
                        3227
                    ]
                }
            ]
        }
    ]
}
```

I can identify `ThrowStatement` for `MapsAPIError`, what the arguments are, and I can even access the comment for annotation purposes.

As I said, that requires more work than what I did, but we can create a script that walks the AST and creates the list of errors.

VictorVelarde:
Thx for your review @rochoa and @elenatorro. Regarding to your comments:

1. Although this might look to address the original issue [...] The error-list.js file is only a list of error message patterns.
> My fault, I thought that had already been discussed, so I just let myself go with that approach.

2. Maintaining that list up-to-date is hard. [...] This doesn't improve the situation over having error messages with string interpolation.
> It is true, building and keeping the list up-to-date feels tough. We could improve the current system, displaying also the type > category > params, but I guess that won't solve some other issues.

3. I would expect to have a higher error reusability [...]
> We are mostly using generic `Error` and `CartoValidationError` for everything. The last one is the piece we have reused, not the messages themselves. Introducing more specific errors, like `WebGLError` or  `CartoSecurityError` (403, 404...) sounds good.

4. Regular expressions are hard to maintain and harder to reason [...]
> +1

I like the use of consts as enums that Elena propose but I also agree with her that we should discuss a little bit more on the approach before tackling it.

VictorVelarde:
After our meeting, these are the main points I have taken notes:

1. Problems with current approach:
- complex / fragile regex
- separation of message management from original error context
- current value of the error-list is perceived as low, comparing to simpler inline throw +

2. Proposal:
- Simpler approach
- Create a set of meaningful typed CARTO-errors
- Document errors in a useful way (list, table with categories...?)
- Use throws at jsdoc when needed
- Create an error list from the code itself, if needed

davidmanzanares:
Regarding the list of error types, I propose:
- SourceError
   - MapsAPI error
      - AuthError
- SyntaxError (parser errors)
- ParameterError (invalid parameter)
  - TypeError (viz typechecks)
  - RangeError (parameter is of a valid type, but outside the valid range)
  - MaxArgumentsError (too many args)
- WebGLError
   - WebGLSupportError
   - WebGLUnexpectedError

VictorVelarde:
I made some research and I want to share some related notes with you:

## 1. User point of view
- An external dev will probably use errors like this:
  - a) for debugging
  ```js
    console.log(e); // being e the error thrown by carto-vl
    console.log(e.name);
    console.log(e.message);
    console.log(e.stack); // (not standard)
  ```

  - b) Once he's got an error he could:
    ...google the error message --> eg. `Percentile must be a fixed literal number`
    ...look for more info at carto-vl's docs --> where (general search)?

  - c) Once he knows about the error...
  ```js
    // he could use try-catch + `instanceof` to react (retry, logging / rethrowing...)
    try {
      let x = whatEverFromCartoVL();
    } catch (err) {
      if (err instanceof CustomError) {
        showMessage("Whatever" + err.message);
      } else {
        console.error(err.message);
      }
    }
  ```

## 2. CARTO-VL dev point of view:
- Custom Error Types provide a "cleaner and more consistent error handling code"
- We want to have `error message` next to the `throw` (no error-list)
- We can create specific types by using an ES6 custom class, like this:
```js
  class MyError extends CartoError {
    constructor(message) {
      // whatever...
    }
  }
```
- How to return errors?
  - sync --> we'll usually go for a simple **throw**
  - async --> we would have different options (`callback(err, result)`, `promise.reject`, or even an `error event` using an EventEmmitter) but it seems it is not necessary for us now.


## 3. Notes on other libraries
### gmaps JS API
  - has a public error list here: [Google Maps Error Messages](https://developers.google.com/maps/documentation/javascript/error-messages), though it mostly has errors related to backend service.
  - writes errors messages to window.console, with something like: `Google Maps API error: [ERROR CODE] [Link to API document]`

### mapbox-gl-js
- They have an `error event` on map, and if no listener is registered, then they `console.log` [see map event](https://www.mapbox.com/mapbox-gl-js/api/#map.event:error)
- They internally use these types:
  - common ones such as `Error`, `RangeError`, `TypeError`... without a code, and just using string literals
  - some custom errors, like `RuntimeError`, `ParsingError` ... Example:
  ```js
    if (index >= array.length) {
      throw new RuntimeError(`Array index out of bounds: ${index} > ${array.length - 1}.`);
    }
  ```
- Examples at https://github.com/mapbox/mapbox-gl-js/search?q=%22throw+new%22&unscoped_q=%22throw+new%22

### ArcGIS API for JS
- They have a generic `esri / core / error` type.

### openlayers
- They have a url with a [list of errors](https://openlayers.org/en/latest/doc/errors/)

### Builder
- just `throw new Error('whatever')`
- It uses the external service TrackJS

### CARTO.JS
- `error-list.js` with `CartoValidationError` but also simple `throw new Error` [examples](https://github.com/CartoDB/carto.js/search?q=%22throw+new%22&unscoped_q=%22throw+new%22)

## 4. Some general references:
- [Error at MDN web docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)

- [What's a good way to extend Error in JavaScript?](https://stackoverflow.com/questions/1382107/whats-a-good-way-to-extend-error-in-javascript)

- [A Guide to Proper Error Handling in JavaScript](https://www.sitepoint.com/proper-error-handling-javascript/)

- [Exceptional exception handling in JS](https://www.sitepoint.com/exceptional-exception-handling-in-javascript/)

---

## 5. TASKS
- I like the general categories suggested by @davidmanzanares . I'm gonna keep its principles, but starting with a shorter list (just the parent type) to keep it simple... and I'll ask for your feedback later on code, fine?.

I'll work on:
[ ] **CartoMapsAPIError**
  - Expected from the user:
    * check the `credentials`
    * check the name of the `dataset` or `SQL syntaxis`
    * change the SQL as 'X' is not supported
    * read about general CARTO services (about oauth, maps API...)?
  - Comments:
    We can create 'SourceError' or more concrete errors later, if other subcategories are worthy

[ ] **CartoParsingError** (more concrete than generic JS `SyntaxError`)
  - Expected from the user:
    * check the code, correct wrong syntaxis and try again
    * read about the syntax on carto-vl guides / reference

[ ] **CartoValidationError** (our current type slightly adapted, which is pretty equivalent to `ParameterError`)
  - Expected from the user:
    * check the code and quickly fix it
    * check the reference for the object / expression
  - Comments:
    We will take advantage of current common checks (eg. `checkMaxArguments`), used in multiple expression constructors.

[ ] **CartoRuntimeError**
  - Expected from the user:
    * read the error and investigate the cause.
    * create an issue / contact support?
  - Comments:
    Instead of using detailed types, such as `WebGLError`, I'd go for a generic non-recoverable error, with a detailed message.

I'll derive all these errors from the current `CartoError` and I'll keep Carto-preffix, to avoid confusions, cause there are some similar error names on mapbox.gl

VictorVelarde:
I haved reset previous changes, cause revert presented some non trivial errors, and I have kept the original changes in another branch [743-error-list-original]

VictorVelarde:
Summary:

- Now there 4 error types, inheriting from a simpler `CartoError`: 
  - **CartoValidationError**
  - **CartoParsingError**
  - **CartoMapsAPIError** 
  - **CartoRuntimeError**

- We have removed `error-list` and used inline messages instead (during the process, we saw several messages not being used: nonValidServerURL, nonValidTemplateURL, metadataRequired...).

- We also have created some `enums` for error categories to help when building common messages (eg. `CartoValidationTypes.MISSING_REQUIRED` or `CartoRuntimeError.WEB_GL`). That avoids too nested error hierarchies and it works like a kind of 'pseudo-types'. It's easy to trace every use of each error type or category, and it also help us with the tests.

- Now we have more options for the **tests**:
```js
  const f = () => { whatEver(); };
  expect(f).toThrow();
  expect(f).toThrowError(CartoError); // parent type
  expect(f).toThrowError(CartoValidationError); // subtype
  expect(f).toThrowError('Exact error message'); // with exact message
  expect(f).toThrowError(CartoValidationError, 'Exact error message'); // type + exact message

  // looking for specific message patterns using enums...
  const regex = thatContains(cvt.MISSING_REQUIRED); // pattern using 'enum' (thatContais is an external function in utils to avoid dealing with complex regex at tests)
  expect(f).toThrowError(regex);

  const regex2 = thatContains(cvt.INCORRECT_VALUE + ' \'apiKey\'')
  expect(f).toThrowError(regex2); // more specific message pattern
```

- And finally, all our errors are now under one of our `CartoError` subtypes... (and probably we shouldn't use generic `Error` in the future, but let's see).

src/Layer.js
```diff
@@ -179,7 +181,7 @@ export default class Layer {
 
         await this._context;
         if (this._atomicChangeUID > uid) {
-            throw new Error('Another atomic change was done before this one committed');
+            throw new CartoRuntimeError('Another atomic change was done before this one committed (1)');
```
elenatorro: I would remove `(1)` and `(2)` even the error message is the same. I recognise I do not know if this is a convention, if it is go ahead. The rest looks ✨

davidmanzanares: I think the are useless now... But, if in the future we discover or introduce a bug, they could be highly useful. And this part of the code has been historically error-prone. I'd leave them.

VictorVelarde:
It all depends on how badly we want to enforce ourselves to define a "subtype"... do we? 
Maybe it would be interesting to add an overloaded constructor, where you can include the 'type' + message, keeping the one with just "message"

VictorVelarde:
After our meeting, I'm going to merge this in. We can add more details in future (new error types & categories) if we need them, and also create an error-list

rochoa:
VAMOS! 💪

Issue #825, opened 2018-08-08
Make unary and binary operators case insensitive


davidmanzanares:
`filter: $a >10 and $a < 10` works
`filter: $a >10 AND $a < 10` doesn't work

PR #931, opened 2018-09-05
Make operators case insensitive

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/825

### DIFF
--- a/src/renderer/viz/parser.js
+++ b/src/renderer/viz/parser.js
@@ -76,13 +76,17 @@ function parseVizNamedExpr (vizSpec, node) {
     if (node.operator !== ':') {
         throw new Error('Invalid syntax.');
     }
+
     if (node.left.name.length && node.left.name[0] === '@') {
         node.left.name = '__cartovl_variable_' + node.left.name.substr(1);
     }
+
     let name = node.left.name;
+
     if (!name) {
         throw new Error('Invalid syntax.');
     }
+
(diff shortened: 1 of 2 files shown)


src/renderer/viz/parser.js
```diff
@@ -115,7 +119,7 @@ function parseFunctionCall (node) {
 function parseBinaryOperation (node) {
     const left = parseNode(node.left);
     const right = parseNode(node.right);
-    switch (node.operator) {
+    switch (node.operator.toLowerCase()) {
```
jgoizueta: So now we accept `and` and `AND` but also `And`, `aNd`, `anD`, `ANd`, `AnD` and `aND`. Oh well, I guess that's case-insensitivity is all about...

elenatorro: hahaha yes 🤷‍♂️

VictorVelarde: I thought we were going just after 'and' & 'AND' variations, so maybe we could add this on the case side (and remove `toLowerCase`):

```
    case 'and':
    case 'AND':
        return expressions.and(left, right);
    case 'or':
    case 'OR':
        return expressions.or(left, right);
    case 'in':
    case 'IN':
        return expressions.in(left, right);
    case 'nin':
    case 'NIN':
        return expressions.nin(left, right);
```

Issue #934, opened 2018-09-05
Add syntax sugar for offset/placement


elenatorro:
Related with https://github.com/CartoDB/carto-vl/pull/919 and https://github.com/CartoDB/carto-vl/issues/361

Add `transform` property and `translate` expression as syntax sugar of `offset` and `placement` respectively. 

```
offset: placement(0, 1)
```

```
transform: translate(0, 1)
```

elenatorro:
I have some concerns  / question about this issue:

* What happens if there are both offset and transform property in the viz?
* Should we create a new `translate` expression, or only use it as 'syntax sugar' of placement?
* If we create a new expression, should `translate` expression type be 'transform' or 'transformation', so future transform expressions use the same type?
* Should we add in the documentation that both properties are now equivalent?

elenatorro:
After talking with @davidmanzanares, we have to do the following:

1. Create a new Translate expression with the same code of Placement (changing the 'type')
2. Change 'offset' name to 'transform'.

davidmanzanares:
Fixed by #938

PR #938, opened 2018-09-06
Rename offset and add translate expression

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/934


elenatorro:
@davidmanzanares we've been busy, _sorry not sorry_ 😂 🚀

Issue #674, opened 2018-07-10
Support to render text labels


jacknkandy:
It seems like there is a big feature missing from the Carto VL library - namely to be able to render labels alongside any geometry. This feature exists within Carto builder as well as the CARTO.js library.

VictorVelarde:
Hi @jacknkandy 

With the new CARTO VL v1.1.0 we have just released, we added a new method to features called `getRenderedCentroid`. That allows creating a dynamic geojson layer, which then can be used to apply the common Mapbox GL labeling capabilities.

We have prepared a few examples in our Dev Center, such as this one: https://carto.com/developers/carto-vl/examples/#example-point-labels

PR #940, opened 2018-09-07
Adding text labels [WIP]

elenatorro:
### Note: this is a PoC

Related with https://github.com/CartoDB/carto-vl/issues/674

This is a WIP. I started this first approach so we can discuss several things:

* Should we use the MapboxGL API for labels? Or should we create our own property / expressions?
* Discuss the naming for this new property. I currently used `label` as property and `text` expression:

```
label: text('example', x, y)
```

* Ideally, we must be able to pass a property and show the value of the property in the label.
* With this approach, we have the control of the label, but there are some limitations related with textures.

captura de pantalla 2018-09-07 a las 11 47 12


src/renderer/viz/expressions/Text.js
```diff
@@ -43,8 +58,18 @@ export default class Text extends Base {
...
+
+        if (!this.init) {
+            for (let i = 0; i < categories.length; i++) {
+                this[`_text${i}`] = categories[i];
+            }
+
             this.init = true;
+        }
+
+        if (this.init && !this.canvas) {
```
VictorVelarde: I guess this.init is always true here...

elenatorro:
Closing since this is a PoC

PR #941, opened 2018-09-07
Fix symbol override color

VictorVelarde:
### Context
When **using a symbol but not a color**, a default color was being created underneath. Later on, when using blend, real symbol color was being overriden by that default color (that's why in the issue a red-alike color was visible, and not the original yellow of the symbol). This special case has now been considered, modifying the shaders (thx @davidmanzanares for your help!)

### Tasks
- [x] fix implementation
- [x] render test
- [ ] CR


VictorVelarde:
Hi @davidmanzanares we would need to generate the reference image in a linux machine (yours or @jesusbotella right?)

examples/styling/image-external-no-color.html
```diff
@@ -0,0 +1,87 @@
+<!DOCTYPE html>
```
davidmanzanares: Do we need a test for this? I mean, the issue seems to be a bug. If there the example is really useful for the users (check with @makella ) we can add it. If the example is only useful for ourselves (cvl devs) we could add it to `debug/`, although we have a test

VictorVelarde: 👍 That's right, it is a bug, and nothing really special in the example itself for the users. 

I'm gonna move it to `debug/` (I didn't know about that possibility)

makella: the interactive-based style that we have in the dev center example is this one:
https://carto.com/developers/carto-vl/examples/#example-interactive-based-styling

for images, we don't have an example that shows style changes on hover/click so I think we are ok.

if there is something special that should be pointed out about the workflow for interactive based styling with images, we can definitely add.

VictorVelarde: Hi @makella ! This example helped us debugging the error, but it works just as the example you have posted, so I moved it to the "internal" debug folder. Is it fine?

makella: yep!! thank you!!

PR #953, opened 2018-09-14
Ramp GLSL refactor

davidmanzanares:
This refactor `ramp` GLSL code generation to avoid so many calls to `mix`. I thought this was going to improve performance, but it isn't, upon closer look,  it doesn't seem that the current `ramp` is impacting performance at all.

However, since I already done it, we can discuss if this code is easier or not.

The code still lacks the cielab transformation, it would be super easy to add, but I don't want to keep working on this if this is not going to get merged.


elenatorro:
It was failing in master --> https://github.com/CartoDB/carto-vl/pull/955, so the problem above it is not something introduced in this PR

PR #972, opened 2018-09-18
Fix filtered image

davidmanzanares:
Regression example: file:///home/dmanzanares/github/renderer-prototype/examples/editor/index.html#eyJhIjoibWVyZ2VkX21lcmdlZCIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoiLy9pbWFnZSgnaHR0cHM6Ly9saWJzLmNhcnRvY2RuLmNvbS9jYXJ0by12bC9hc3NldHMvaHVycmljYW5lLnN2ZycpXG5cbndpZHRoOiBzY2FsZWQoc3FydCgkd2luZF9zcGVlZCkvMilcbi8vc3ltYm9sUGxhY2VtZW50OiBwbGFjZW1lbnQoc2luKCR3aW5kX2RpciksY29zKCR3aW5kX2RpcikpXG5zeW1ib2w6IFxuXG5cbmltYWdlKCdodHRwczovL2xpYnMuY2FydG9jZG4uY29tL2NhcnRvLXZsL2Fzc2V0cy9odXJyaWNhbmUtMnB4LW91dGxpbmUuc3ZnJykqcmFtcChsaW5lYXIoJHdpbmRfc3BlZWQpLGFnX2dybnlsKVxuZmlsdGVyOiR3aW5kX3NwZWVkID41Ljlcbm9yZGVyOiBkZXNjKHdpZHRoKCkpXG50cmFuc2Zvcm06IHJvdGF0ZSgxMDAwKiR3aW5kX2Rpcipub3coKSlcbiIsImYiOnsibG5nIjotNjkuMDc4MTg0MzM1MzcwNzksImxhdCI6MzEuODc3NzIyOTg0ODA5NzA4fSwiZyI6My4zMjAyMzg3MzQ1OTQ1ODY3LCJoIjoiRGFya01hdHRlciIsImkiOiJkYXRhc2V0In0=

Filtering was being ignored in combination with "No override color".

Solution: defer the application of filtering until the overriding/no-overriding computations have been completed.


src/renderer/shaders/symbolizer/symbolizerFragmentShader.glsl
```diff
@@ -19,6 +20,7 @@ void main(void) {
     }else{
         c = symbolColor;
     }
+    c.a *= filtering;
```
VictorVelarde: ok, so now the filter is applied or not when dealing with colors, right?

src/renderer/shaders/symbolizer/symbolizerVertexShader.glsl
```diff
@@ -65,5 +65,7 @@ void main(void) {
     if (size==0. || (color.a==0. && color != noOverrideColor) || size<orderMinWidth || size>=orderMaxWidth){
         p.x=10000.;
```
VictorVelarde: I remembered when we added the noOverrideColor case, but why do we need to move this position so much? is there no other way to avoid the rendering in those cases?

src/renderer/shaders/symbolizer/symbolizerFragmentShader.glsl
davidmanzanares: The filtering affects the opacity everytime (instead of just when the `color:` was overriding the symbol color

src/renderer/shaders/symbolizer/symbolizerVertexShader.glsl
davidmanzanares: It's a way to ignore the vertex and stop processing it. There is no other way to discard a vertex, this makes sense because normally, it isn't safe to "discard" a vertex since the other 2 vertices may not be discarded. However, for the CVL case, the 3 vertices of a triangle will be discarded or not coherently (no triangles with discarded and not discarded vertices).

davidmanzanares:
I didn't create an issue for this (I discovered it), but I added this to the Kanban

src/renderer/shaders/symbolizer/symbolizerFragmentShader.glsl
VictorVelarde: thx for your explanation!

Issue #701, opened 2018-07-13
Support for pitch and bearing


rochoa:

davidmanzanares:
After talking with the rest of the team we have decided to deprioritize this issue due to its complexity and due to the fact that there are other important issues on hold.

Some of my super shallow research:

- We will need to create a perspective matrix and some maths. I saw this math utility library (https://github.com/toji/gl-matrix), seems good, but the minified bundle size goes up from 242KB to 259KB.
- https://github.com/CartoDB/carto-vl/issues/240 is a requirement for this
- We'll need to change the computation of getting the tile list for a particular zoom, center, pitch and bearing.
- We'll need to adapt the rendering shaders
- We'll need to test the performance and decide if we stick to a two-path rendering (with 2D and 3D shader versions)

davidmanzanares:
Integrating CARTO VL with the upstream custom layers proposal would make this easier

rochoa:
Any blocker holding this yet?

davidmanzanares:
No, #240 was blocking it, but it's already fixed

PR #984, opened 2018-09-19
Working PoC with rotation/tilt of the map

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/701

- [x] Rendering
- [x] Tile selection
- [ ] Fix rendering flickering issue
- [ ] Interactivity
- [ ] Add render test

### DIFF
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -371,6 +371,7 @@ export default class Layer {
         // Call request data if the matrix has changed
         if (!util.equalArrays(this._matrix, matrix)) {
             this._matrix = matrix;
+            this.renderer.matrix = matrix;
             this.requestData(matrix);
         }
     }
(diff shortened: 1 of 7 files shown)


src/utils/collision.js
```diff
@@ -4,13 +4,6 @@ const SEPARATING_LINE_FOUND = 'separatingLineFound';
...
 export function triangleCollides (triangle, viewportAABB) {
-    const viewport = [
-        { x: viewportAABB.minx, y: viewportAABB.miny },
-        { x: viewportAABB.minx, y: viewportAABB.maxy },
-        { x: viewportAABB.maxx, y: viewportAABB.miny },
-        { x: viewportAABB.maxx, y: viewportAABB.maxy }
-    ];
-
     /*
      * TODO
```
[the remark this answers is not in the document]
davidmanzanares: I don't think so, this was done some time ago to remark some possible ways to optimize the function. Since it is a performance intensive function, we should probably review this in the future.

src/Layer.js
```diff
@@ -512,29 +521,96 @@ export default class Layer {
...
+        zoom = util.computeMapZoom(this.map);
+        center = util.computeMapCenter(this.map);
 
         this.renderer.setZoom(zoom);
         this.renderer.setCenter(center);
     }
 
     _getViewport () {
-        return this.renderer.getBounds();
+        const inv = mat4.invert([], this._matrix);
```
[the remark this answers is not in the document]
davidmanzanares: I'll extract the `unproject` function

debug/line-interactivity.html
```diff
@@ -0,0 +1,164 @@
+<!DOCTYPE html>
```
VictorVelarde: @davidmanzanares I have seen something weird with this example, regarding to 'Interactivity'. See this gif:

interactivity-issue

It's like the coordinates have a strange offset

VictorVelarde: I didn't record it, but that also happens when using tilt

davidmanzanares: Could you re-check that? I'm not seeing the same behavior:
peek 2018-10-02 12-10

src/renderer/DummyDataframe.js
```diff
@@ -17,5 +17,9 @@ export default class DummyDataframe {
         this.metadata = metadata;
         this.propertyCount = 0;
         this._aabb = computeAABB(geom, type);
+        this.t1 = [0.1, 0.1, 0.1, 0.1];
```
VictorVelarde: How about extracting the array to a descriptive variable? I'd also consider extracting these initializations to its own little methods. A reader can find quite difficult to understand these operations objective... what's `_aabb`, what's `t1`...`t4`?

debug/line-interactivity.html
VictorVelarde: I'm still having issues. I'm using Chrome v 69.0.3497.100 on Mac OS X 10.13.6

src/renderer/DummyDataframe.js
davidmanzanares: `_aabb` was long before this PR, and I don't think it's that bad, it's just the array of AxisAlignedBoundingBox, which is a common term in the context of graphics.

Regarding `t1`... `t4`, I agree it's very ugly. But the thing is that `t1`...`t4` are different things depending on the method are used (in all cases they are vectors, but that can be seen in the line too). Of course, to have multiple meanings it's even uglier, but this is a performance optimization that was added after discovering performance problems. They are used to re-use the memory of the vectors without re-allocating memory for new arrays. They are temporal variables that never escape the scope of the public methods of this class.

VictorVelarde: Oh great, I didn't know AABB = AxisAlignedBoundingBox. Maybe we could add that somewhere in the comments, or even in the glossary.

Regarding to t1---t4, if they are just performance helpers, maybe we could add something like:
```js
ctor(){
   //... whatever
   this._initializeVectorsCache(); // or a similar name
}
```

and then something like
```js
/*
 * Initialize a set of variables to apply performance optimizations
 */
_initializeVectorsCache(){
  const seedValue = [0.1, 0.1, 0.1, 0.1];
  this.t1 = seedValue;
  this.t2 = seedValue;
  this.t3 = seedValue;
  this.t4 = seedValue;
}
```

That way the user is not distracted in the constructor but he gets a notion about them (t1...t4 are about performance, I won't worry about them for now)

debug/line-interactivity.html
davidmanzanares: Debugging it with @rochoa showed that this is probably caused by high-DPI screens with high devicePixelRatios. I'm looking for a fix now.

src/renderer/Dataframe.js
```diff
@@ -303,12 +330,28 @@ export default class Dataframe extends DummyDataframe {
...
-        });
+    _computePointRadius (feature, viz) {
+        const SATURATION_PX = 1024;
+        const diameter = Math.min(viz.width.eval(feature), SATURATION_PX) + Math.min(viz.strokeWidth.eval(feature), SATURATION_PX);
+
+        return diameter / 2;
+    }
+
+    _computeLineWidthScale (feature, viz) {
+        const SATURATION_PX = 1024;
```
VictorVelarde: It seems this constant is reused among the different methods. We could push it up to share it

src/renderer/Renderer.js
```diff
@@ -221,10 +220,25 @@ export default class Renderer {
     }
 
     renderLayer (renderLayer, drawMetadata) {
+        this.drawMetadata = drawMetadata;
         const dataframes = renderLayer.getActiveDataframes();
         const viz = renderLayer.viz;
         const gl = this.gl;
-        const aspect = this._getAspect();
+
+        dataframes.forEach(dataframe => {
```
VictorVelarde: We can extract this to a method like `this._updateMatrixFor(dataframes)` and avoid making the current method grow more in lines

src/renderer/Renderer.js
```diff
@@ -335,17 +347,8 @@ export default class Renderer {
...
-            gl.uniform2f(renderer.vertexOffsetUniformLocation,
-                (scale / aspect) * (this._center.x - dataframe.center.x),
-                scale * (this._center.y - dataframe.center.y));
-            gl.uniform2f(renderer.normalScale, 1 / gl.canvas.clientWidth, 1 / gl.canvas.clientHeight);
-
-            dataframe.vertexScale = [(scale / aspect) * dataframe.scale, scale * dataframe.scale];
-
-            dataframe.vertexOffset = [(scale / aspect) * (this._center.x - dataframe.center.x), scale * (this._center.y - dataframe.center.y)];
+            gl.uniform1f(renderer.normalScale, 1 / (Math.pow(2, drawMetadata.zoomLevel) * RESOLUTION_ZOOMLEVEL_ZERO * dataframe.scale));
+            console.log(1 / (Math.pow(2, drawMetadata.zoomLevel) * dataframe.scale));
```
VictorVelarde: I guess we can remove this console.log

src/renderer/shaders/geometry/point/pointVertexShader.glsl
```diff
@@ -72,6 +69,7 @@ void main(void) {
   if (size == 0. || (stroke.a == 0. && color.a == 0.) || size < orderMinWidth || size >= orderMaxWidth) {
     p.x = 10000.;
   }
+  p.z=0.9999;
```
VictorVelarde: What does this 0.9999 stands for? a very high z? can we add some explaining constant?

test/unit/renderer/dataframe.test.js
```diff
@@ -1,9 +1,18 @@
 import Dataframe from '../../../src/renderer/Dataframe';
 import Metadata from '../../../src/renderer/Metadata';
 import { pointInTriangle } from '../../../src/utils/geometry';
-import { translate } from '../../../src/renderer/viz/expressions';
+import { mat4 } from 'gl-matrix';
 
 describe('src/renderer/Dataframe', () => {
+    const m = [];
```
VictorVelarde: Here we could reuse `_createMatrix ` method from Map.js (after just a little refactor to receive ratio as a parameter).

examples/widgets/viewportAgg2.html
```diff
@@ -0,0 +1,74 @@
...
+    carto.setDefaultAuth({
+      user: 'cartovl',
+      apiKey: 'default_public'
+    });
+
+    // Define layer
+    const source = new carto.source.Dataset('mnmappluto');
+    const viz = new carto.Viz(`
+        @avg: viewportMax($numfloors)
+        transform: translate(100, 50)
```
VictorVelarde: Consider if we remove this translate and keep the example or remove the example itself

src/renderer/Dataframe.js
davidmanzanares: :+1:

src/renderer/Renderer.js

src/renderer/Renderer.js
davidmanzanares: :1+:

src/renderer/shaders/geometry/point/pointVertexShader.glsl
davidmanzanares: Actually, I can refactor that logic, hope it helps

src/renderer/Dataframe.js
```diff
@@ -286,8 +286,8 @@ export default class Dataframe extends DummyDataframe {
         const points = this.decodedGeom.vertices;
         const features = [];
 
-        const WIDTH = this.renderer.gl.canvas.width;
-        const HEIGHT = this.renderer.gl.canvas.height;
+        const WIDTH = this.renderer.gl.canvas.width / window.devicePixelRatio;
```
VictorVelarde: Good catch! 

As we are reusing the same expression 3 times in several places, we could also have:
```js
_getCanvasSize(){
    const WIDTH = this.renderer.gl.canvas.width / window.devicePixelRatio;
    const HEIGHT = this.renderer.gl.canvas.height / window.devicePixelRatio;
    return [WIDTH, HEIGHT]; // or maybe with an object {width, height} 
}
```

And then just call it like `const [WIDTH, HEIGHT] = this._getCanvasSize()` in every place.

test/unit/renderer/dataframe.test.js
davidmanzanares: Yes... but re-using that complex logic from the implementation in the tests... I'd prefer to keep this as is to be honest

src/renderer/Dataframe.js
VictorVelarde: I'm going to test that fix!

examples/widgets/viewportAgg2.html
davidmanzanares: :+1: removing the example

src/renderer/Dataframe.js
davidmanzanares: Yes... and it looks better. But I'm worried about the performance, I already suffered the cost of allocating new arrays. I don't know if this will be costly though, but it would require to re-check the performance :(

VictorVelarde: I've tested and it works like a charm after the fix

src/renderer/Renderer.js
```diff
@@ -470,6 +456,23 @@ export default class Renderer {
         gl.disable(gl.CULL_FACE);
     }
 
+    _updateDataframeMatrices (dataframes) {
```
VictorVelarde: ✨ Great!

src/renderer/shaders/geometry/point/pointVertexShader.glsl
```diff
@@ -67,9 +67,8 @@ void main(void) {
     }
 
   if (size == 0. || (stroke.a == 0. && color.a == 0.) || size < orderMinWidth || size >= orderMaxWidth) {
-    p.x = 10000.;
+    p.z = p.w; // Force clipping
```
VictorVelarde: 👍 Comments are really useful inside the shaders!

src/renderer/Dataframe.js
```diff
@@ -7,6 +7,8 @@ import { RESOLUTION_ZOOMLEVEL_ZERO } from '../constants/layer';
 // in a non-lazy manner
 const MAX_GPU_AUTO_UPLOAD_TEXTURE_LIMIT = 32;
 
+const SIZE_SATURATION_PX = 1024;
```
VictorVelarde: 👌

src/renderer/Dataframe.js
VictorVelarde: > Yes... and it looks better. But I'm worried about the performance, I already suffered the cost of allocating new arrays. I don't know if this will be costly though, but it would require to re-check the performance :(

Oh, I see, I was considering perfomance... 👍

test/unit/renderer/dataframe.test.js
VictorVelarde: No-worries man

PR #986, opened 2018-09-19
Add build configuration with browserlist + new CartoMap for tests

VictorVelarde:
### Related to https://github.com/CartoDB/carto-vl/issues/28
This PR adds configuration to Webpack so the bundling process makes transpilation using `babel-loader` and a list of supported browsers (**browserslist**). It also moves some references from dependencies to dev-dependencies.

- [x] Babel-loader config
- [x] Browserlist

---

### Related to https://github.com/CartoDB/carto-vl/issues/989 
This PR also removes carto.Map from the library. A renamed CartoMap has been created for the tests

- [x] Exclude carto.Map from bundle 
- [x] Adapt tests to new CartoMap


VictorVelarde:
**browserslist** has not been defined yet with more details. 
It currently includes just `defaults`

VictorVelarde:
Blocked by #984 (because of the `carto.Map` removal)

test/.eslintrc.json
```diff
@@ -1,5 +1,8 @@
 {
     "env": {
         "mocha": true
+    },
+    "globals": {
+        "mat4": true
```
IagoLast: why this is global?

VictorVelarde: mat4 = gl-matrix. It is a dependency of CartoMap, and it is loaded from node_modules with `<script src="<%- glmatrix %>"></script>`

test/CartoMap.js
```diff
@@ -1,28 +1,30 @@
-import { layerVisibility, RESOLUTION_ZOOMLEVEL_ZERO } from '../constants/layer';
-import CartoValidationError, { CartoValidationTypes as cvt } from '../errors/carto-validation-error';
-import { mat4 } from 'gl-matrix';
+// CartoMap depends on gl-matrix, which after imported is globally available as 'mat4'
+// Constants extracted from /carto-vl/src/constants/layer.js
```
elenatorro: This comment is not needed.

VictorVelarde: Do you mean the second one? the one on mat4 or both? 😄

VictorVelarde: Currently `carto.layerVisibility` is not available externally, nor it is `RESOLUTION_ZOOMLEVEL_ZERO`, so I'm kind of replicating them just for this utility class.

I'll keep the comments as a reminder of this.

Issue #977, opened 2018-09-18
Reference – review structure


rochoa:
Reference [WIP]: Proposed structure (1)

* Carto
  * Sources
      * Dataset
      * GeoJSON
      * SQL
      * MVT
        * MVTMetadata
        * MVTOptions
        * MVTProperty
  * Viz

  * Layer
    * LayerEvent

  * Expressions (+palette, +icons)

  * Expression Types
    * Color
    * Category
    * Date
    * Number
    * Image

  * Features
  * Interactivity
    * Events

* Errors

elenatorro:
🚀

PR #990, opened 2018-09-21
Reorganize reference documentation

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/977

Staging developers web: http://977-review-reference-structure.developers.carto-staging.com/developers/carto-vl/reference/


src/basemaps.js
```diff
@@ -1,26 +1,26 @@
 /**
- * Type of carto.basemaps.voyager
+ * Voyager vector basemap
```
davidmanzanares: We could link to https://carto.com/location-data-services/basemaps/

src/basemaps.js
```diff
@@ -1,26 +1,26 @@
...
  */
 export const darkmatter = 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json';
 
 /**
- * Type of carto.basemaps.positron
+ * Positron vector basemap
  *
- * @typedef {string} CARTO Positron style
  * @memberof carto.basemaps
+ * @name positron
```
davidmanzanares: Isn't the name extracted automatically?

src/renderer/viz/expressions.js
```diff
@@ -97,20 +108,26 @@
...
+ * @memberof Types
  * @api
  */
 
 /**
  * Type of Date Expressions.
  *
  * @typedef {} Date
+ * @name Date
+ * @memberof Types
```
davidmanzanares: I don't understand why the types are "memberof" `Types`. There is no constructor on `carto.Types. I find this confussing.

src/renderer/viz/expressions.js
```diff
@@ -120,9 +137,291 @@
...
+ * More information in {@link carto.expressions.image|carto.expressions.image}.
+ *
+ * @typedef {} Image
+ * @name Image
+ * @memberof Types
  * @api
  */
 
+/**
+ * Type of Translate Expressions.
```
davidmanzanares: Transformation type can be a translation, a rotation, or a chain of those two.

src/renderer/viz/expressions.js
```diff
@@ -120,9 +137,291 @@
...
+ * More information in {@link carto.expressions.translate|carto.expressions.translate}.
+ *
+ * @typedef {} Transformation
+ * @name Transformation
+ * @memberof Types
+ * @api
+ */
+
+/**
+ * @namespace Constants
```
davidmanzanares: Why do we have another namespace with the expressions duplicated?

src/renderer/viz/expressions/Animation.js
```diff
@@ -49,6 +49,9 @@ let waitingForOthers = new Set();
  *   color: ramp(linear(clusterAvg($temp), 0,30), tealrose)
  *   filter: animation(linear($date, time('2022-03-09T00:00:00Z'), time('2033-08-12T00:00:00Z')), 40, fade(0.1, 0.3))
  * `);
+ * Animation class
+ *
+ * This class is instanced automatically by using the `animation` function. It is documented for its methods.
```
davidmanzanares: I'd change the phrasing. I'd say something like: the `animation` function returns instances of the `Animation` class

src/renderer/viz/expressions/Animation.js
```diff
@@ -207,7 +208,7 @@ export class Animation extends BaseExpression {
      *   document.getElementById('timestamp').innerHTML = currTime;
      * });
      *
-     * @memberof carto.expressions.Animation
+     * @memberof s.Animation
```
davidmanzanares: memberof `s` works fine?

src/renderer/viz/expressions/Ramp.js
```diff
@@ -97,6 +97,20 @@ import { DEFAULT_RAMP_OTHERS } from './constants';
 * @function
 * @api
 */
+
+/**
+ * Ramp Class
+ *
+ * A mapping between an input (numeric or categorical) and an output (number, colors and/or images)
+ * This class is instanced automatically by using the `ramp` function. It is documented for its methods.
```
davidmanzanares: Same phrasing

src/renderer/viz/expressions/base.js
```diff
@@ -71,12 +69,25 @@ export default class Base {
         throw new CartRuntimeError('Unimplemented');
     }
 
+    /**
+     * Get the expression value
```
davidmanzanares: We should document the potential case of failure (the expression is feature dependant)

src/sources/Dataset.js
```diff
@@ -36,9 +36,8 @@ export default class Dataset extends BaseWindshaft {
      *
      * @fires CartoError
      *
-     * @constructor Dataset
-     * @extends carto.source.Base
```
davidmanzanares: the `@extends` indicated that this could be used in the place of `carto.source.Base`

src/Layer.js
```diff
@@ -18,24 +18,23 @@ const renderers = new WeakMap();
 *
 * A Layer is the primary way to visualize geospatial data.
 *
-* To create a layer a {@link carto.source.Base|source} and {@link carto.Viz|viz} are required:
+* To create a layer a {@link carto.source|source} and {@link carto.Viz|viz} are required:
```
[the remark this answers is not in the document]
VictorVelarde: The current baseclass for all is `carto.source.Base`, which is marked as @abstract, but it has no methods. Maybe it could be renamed to be `carto.source.Source`

src/Layer.js
```diff
@@ -18,24 +18,23 @@ const renderers = new WeakMap();
...
 *
 * @param {string} id - The ID of the layer. Can be used in the {@link addTo|addTo} function
-* @param {carto.source.Base} source - The source of the data
+* @param {carto.source} source - The source of the data
 * @param {carto.Viz} viz - The description of the visualization of the data
 *
 * @example
 * const layer = new carto.Layer('layer0', source, viz);
 *
-* @fires CartoError
```
VictorVelarde: Why did we remove the @fires?

src/basemaps.js
```diff
@@ -1,26 +1,26 @@
...
- * @typedef {string} CARTO Voyager style
  * @memberof carto.basemaps
+ * @name voyager
  * @api
  */
 export const voyager = 'https://basemaps.cartocdn.com/gl/voyager-gl-style/style.json';
 
 /**
- * Type of carto.basemaps.darkmatter
+ * Dark-Matter vector basemap
```
VictorVelarde: Thx for documenting these new constants properly

src/index.js
```diff
@@ -5,18 +5,35 @@
...
  * - {@link carto.Viz|carto.Viz}
+ *
+ * Expressions:
+ * - {@link carto.expressions|carto.expressions}
+ *
+ * Interactivity:
  * - {@link carto.Interactivity|carto.Interactivity}
- * - {@link carto.setDefaultAuth|carto.setDefaultAuth}
- * - {@link carto.setDefaultConfig|carto.setDefaultConfig}
  */
```
VictorVelarde: `on` & `off` functions are missing. They used to be under the old "Functions" header (with setDefaultAuth / setDefaultConfig). Maybe they could appear under a new one called "Events".

src/Layer.js
VictorVelarde: After reading http://usejsdoc.org/tags-throws.html, I think we should use @throws for Error derived classes, and @fires just for events

src/renderer/viz/expressions.js
VictorVelarde: 🔝

src/sources/Base.js
```diff
@@ -2,15 +2,15 @@ export default class Base {
     /**
      * Base data source object.
      *
-     * The methods listed in the {@link carto.source.Base} object are available in all source objects.
+     * The methods listed in the {@link carto.source} object are available in all source objects.
```
VictorVelarde: There are currently no common methods!

src/sources/Dataset.js
```diff
@@ -36,9 +36,8 @@ export default class Dataset extends BaseWindshaft {
      *
      * @fires CartoError
```
VictorVelarde: @throws instead of @fires...

src/sources/MVT.js
```diff
@@ -86,9 +86,8 @@ export default class MVT extends Base {
      *
      * @fires CartoError
      *
-     * @constructor MVT
-     * @extends carto.source.Base
      * @memberof carto.source
```
VictorVelarde: Currently, I can't see MVT class in https://carto.com/developers/carto-vl/reference/#cartosource category, and I think we should include it.

VictorVelarde:
We still have to fix some details (eg. Date or Fade look duplicated), but I think we are on the right path. 

Apart from several comments I have added, I see two possible improvements to be considered, and something that looks like a general devs.center issue.

**Possible improvements**:
* For the palettes, IMO we should find a way to document them visually, effectively showing the colors in the reference. Maybe in a table with the ramps or at least with a link to display them
* Do we have an example / guide to showcase all the included icons?

**Dev-center issue**:
* When navigating inside the reference, the internal 'links' are not properly displayed. See this gif: 
gotoreference. As you can see, the desired link is not visible, and you have to scroll the page up, so it finally appears.

src/Layer.js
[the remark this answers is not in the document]
elenatorro: As David says, carto.source is a namespace. I know that the base class is carto.source.Base, but I think we could mislead people to error if we link to this class directly in this line. What I did here is to link the source namespace, because once you click on it, you go directly to the sources section and you can see the different sources.

src/Layer.js
```diff
@@ -18,24 +18,23 @@ const renderers = new WeakMap();
...
-*
 * @constructor Layer
-* @memberof carto
+* @name carto.Layer
```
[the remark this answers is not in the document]
elenatorro: Because it is the only way, AFAIK, to have a unique section for the Layer in the documentation. We can leave the _@memerof_, though.

src/Viz.js
```diff
@@ -50,18 +50,9 @@ const SUPPORTED_PROPERTIES = [
 /**
  * A vizSpec object is used to create a {@link carto.Viz|Viz} and controlling multiple aspects.
  * For a better understanding we recommend reading the {@link https://carto.com/developers/carto-vl/guides/introduction-to-expressions/|Introduction to Expressions guide}
+ *
  * @typedef {object} VizSpec
- * @property {Color} color - fill color of points and polygons and color of lines, if used with `symbol` the color will override the original image RGB channels
```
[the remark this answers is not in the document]
elenatorro: Because it is duplicated

src/basemaps.js
elenatorro: We are using it to generate the documentation.

src/index.js
elenatorro: Hmmm yes, I have to take a look to this. Thanks!

elenatorro: After taking a look, they are defined in the interactivity section.

src/renderer/viz/expressions.js
elenatorro: Adding this also to the `Viz`.

src/renderer/viz/expressions.js
elenatorro: They are not duplicated, these are not expressions but types. This is meant to list the available types.

src/renderer/viz/expressions/Animation.js
elenatorro: As we saw, this is intended to place these expressions in the sidebar, I am thinking a better way to do so.

src/renderer/viz/expressions/Ramp.js
elenatorro: ?

src/sources/Base.js
elenatorro: good catch!

src/index.js
VictorVelarde: Ok, but we have to be sure that is not misleading, cause the are different funcions: carto.on / carto.off & interactivity.on / interactivity.off

elenatorro:
Hi :) could you please review again this PR? We have done several improvements. There could be a couple of things missing, such as having a visualization of the palettes. I think those are minor improvements, that can be really useful, but I think we need to move forward (we have added a link to the carto colors website)

The sidebar also needs to be improved, but we can handle this by opening a different issue in the developer center.

cc @davidmanzanares @VictorVelarde

src/renderer/viz/expressions/Ramp.js
davidmanzanares: " instanced automatically" sounds weird to me. Saying something like "you can use the `ramp` function to generate instances of the `Ramp` class" sounds better to me

elenatorro:
Yes, I totally agree with that, and I think it can be a good point to start working on after the public release.

Issue #551, opened 2018-06-13
Temporal aggregations


jgoizueta:
This originated in #130 
In that ticket, @davidmanzanares said:
> Since timestamps can have very high resolutions, but most of the times we don't need those, we should have a mechanism for aggregations. After implementing the naive version (without aggregations, setting the timestamp in dimensions), we could add new aggregation functions that only works with timestamps, for example: HOUR(), DAY(), WEEK(), MIN(), SEC(), MSEC())

We need 3 things here:
- [ ] Proper `dimensions` support in the Maps API for time columns
- [ ] A way for CARTO VL to define the granularity of time properties used in visualization (and/or a way to discretize time in animations into a number of buckets as the original torque did).
- [ ] Document and add error control to common date/time operations that are unsupported (see https://github.com/CartoDB/carto-vl/issues/523)

We should also considere a similar mechanism for quantization (in buckets, ranges, ...) of non-time properties to be sent as aggregation dimensions in the backend.

jgoizueta:
# Dimension Granularity Proposal

The idea is to allow defining the granularity (level of detail) of columns used as dimensions of the aggregation, i.e. columns which we don't want to summarize (place inside cluster aggregation functions such as clusterAvg()).

For example, given a timestamp column with millisecond resolutions, we may want to obtain it as years, months, weeks, days, hours, etc. to use for animation.

In torque manage the granularity by specifying a number of buckets, but when using time values it is much more convenient to emply the common units of time.

We'll need to expand the aggregation API in the tiler to support this and define a way to use this feature in Carto VL.

## Client Side (CARTO VL API)

I propose to implement expressions to define the granularity of a column, e.g.  `timeInHours($t)`, in a similar way to `clusterAvg` and the other cluster aggregations:

* Its argument must be a property
* Using it prevents the property from being used in *raw* form  (`$t`).

I'm not considering any alternative, such as variable/property declarations to define the granularity, because similar alternatives were discussed and dismissed for cluster aggregation function.

The use of a `timeInHours($t)` expresion introduces a new column `t_hours` defined as a SQL expression (which returns an epoch in hours) which would also be included in the GROUP BY of the aggregation.

We could have similar functions for numeric columns e.g. to classify by break points or number of buckets.

If the instantiation reports that data won't be aggregated and any of these expressions is used, it should fail (as for cluster aggr functions).

### Naming

We should define a consistent terminology for these expressions which makes its intent as clear as possible.

Some options I can think of:

[...]

PR #991, opened 2018-09-21
Temporal dimensions

jgoizueta:
See #551
~~This requires this [tiler branch](https://github.com/CartoDB/Windshaft-cartodb/tree/time-dimensions)~~ [update: the required tiler changes are in master and deployed]

Includes also fixes to use cluster aggregations in viewportFeatures & interactivity


jgoizueta:
I removed the DO NOT MERGE tag because now the production tilers are ready for this. But this PR's still missing tests!

jgoizueta:
There's a couple of things I'm still not confident about:

### 1. Start/End stats (min, max)

This is about the value of `globalMin(clusterTimeEnd($t))` and`globalMax(clusterTimeStar($t))`

In [this commit](https://github.com/CartoDB/carto-vl/pull/991/commits/7a3c8dd19feff5b372a4f573eed32ab93205e494) I changed my mind about it:

Let's say we have this situation:

```
 T1S      T2S      T3S           clusterTimeStart($t) values
/        /        /
|-- T1 --|-- T2 --|-- T3 --|     clusterTime($t) values
         \        \        \
           T1E     T2E      T3E  clusterTimeEnd($t) values
```

Then, originally (Situation **A**, before the commit) globalMin, Max was computed strictly for the start and end times:
```
globalMin(clusterTimeEnd($t)) = T1E
globalMax(clusterTimeStart($t)) = T3S

globalMin(clusterTimeStart($t))
|                 globalMax(clusterTimeStart($t))
|                 |
 T1S      T2S      T3S           clusterTImeStart($t) values
/        /        /
|-- T1 --|-- T2 --|-- T3 --|     clusterTime($t) values
         \        \        \
           T1E     T2E      T3E  clusterTimeEnd($t) values
         |                 |
         |                 globalMax(clusterTimeEnd($t))
         globalMin(clusterTimeEnd($t))
```

And after the commit (Situation **B**) we use commit limits for the pair of start/end properties:
```
globalMin(clusterTimeEnd($t)) = T1S = globalMin(clusterTimeStart($t))
globalMax(clusterTimeStart($t)) = T3E = globalMax(clusterTimeEnd($t))

globalMin(clusterTimeStart($t))
|                          globalMax(clusterTimeStart($t))
|                          |
 T1S      T2S      T3S           clusterTImeStart($t) values
/        /        /
|-- T1 --|-- T2 --|-- T3 --|     clusterTime($t) values
         \        \        \
           T1E     T2E      T3E  clusterTimeEnd($t) values
|                          |
|                          globalMax(clusterTimeEnd($t))
globalMin(clusterTimeEnd($t))
```

This doesn't affect result when `clusterTimeStart($t)` and `clusterTimeEnd($t)` are used within `linear` with explicit limits, e.g.:
```
linear(clusterTimeStart($t), time(@tstart), time(@end))
linear(clusterTimeEnd($t), time(@tstart), time(@end))
```

But when both are used with no explicit limits (using global min/max):
```
linear(clusterTimeStart($t))
linear(clusterTimeEnd($t))
```

Then, in case **A**, both expressions would yield the exact same, values, so if they're use for styling, the same styles would apply to start and end values.

In case **B**, start and end would differ consistently, so end values would get different styling than start, because they differ in their values.

I changed my mind from the strict perspective of **A** to **B** because it seems more consistent to treat both start and end expressions in combination, but I'm still not quite sure.

### 2. Start/End Date TZ interpretation

For expressions `clusterTimeStart` and `clusterTimeEnd` the backend computes the time  periods correctly in the specified time zone and then returns time-zone neutral representations of it. In the client then we use a numeric representation which works well because it is a continuous (no jumps or gaps) value of seconds. We use the UNIX epoch for that, but as if periods were UTC, so it is a value relative to the time zone specified.

But we expose this value, as we have been doing for dates in general as a JS Date object. What we need here would be the equivalent of PostgreSQL's `TIMESTAMP WITHOUT TIME ZONE`, but `Date` is more like `TIMESTAMP WITH TIME ZONE`. (go figure, maybe that can be useful after all)

So however we do it, the TZ information that Date shows won't be correct (because the date is actually in the requested time zone that, in general cannot be handled by Date that only knows about UTC and the local system time zone).

As it is now, the epoch value is preserved in Date as UTC epoch,
so externally the Date is shown as converted to an incorrect local date.
This means that only UTC representation methods (toUTCString, getUTCFullYear, getUTCMonth, getUTCDate, etc.) should be used to present the correct date.

The other alternatives I can think of:
* Convert to Date preserving the apparent structure (in local).
  Again the TZ information shown is not correct but apart from that the
  representation and all the non-UTC methods are alright.
* Drop the Date representation of dates and use just numbers
  (leave it to the users convert to Date when they wish).

debug/nytaxi.html
```diff
@@ -0,0 +1,122 @@
...
+            container: 'map',
+            style: carto.basemaps.darkmatter,
+            center: [-74.0032059, 40.7482066],
+            zoom: 12,
+        });
+
+
+    carto.setDefaultAuth({
+      apiKey: 'default_public',
+      user: 'cartovl'
```
VictorVelarde: We now use `username` as our preference (to keep the same name than in carto.js), though `user` still works

VictorVelarde:
Hi @jgoizueta 

I'm having a look at this branch, and it currently has *broken unit tests* (and probably the build itself). This is what I'm getting:
```
  errors/CartoValidationError
    ✔ shouldn't allow an error without a message
    ✔ should allow a simple error
    ✔ should allow an error message using several predefined categories
HeadlessChrome 0.0.0 (Mac OS X 10.14.0) ERROR
  {
    "message": "An error was thrown in afterAll\nUncaught ReferenceError: categoryId is not defined",
    "str": "An error was thrown in afterAll\nUncaught ReferenceError: categoryId is not defined"
  }

Finished in 0.341 secs / 0.169 secs @ 19:08:41 GMT+0200 (CEST)

SUMMARY:
✔ 118 tests completed
```

As you see, just 118 tests are executed, not the whole 900. This points to a circular reference problem somewhere we should have a look.

debug/nytaxi.html
```diff
@@ -0,0 +1,122 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Animation demo | CARTO VL</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta charset="UTF-8">
+  <script src="../dist/carto-vl.js"></script>
+  <script src='https://libs.cartocdn.com/mapbox-gl/v0.48.0-carto1/mapbox-gl.js'></script>
```
VictorVelarde: We should probably update this to v0.50, just in case

src/renderer/Metadata.js
```diff
@@ -23,7 +26,35 @@ export default class Metadata {
...
+                Object.values(property.aggregations).forEach(propName => {
+                    this._addProperty(baseName, propName);
+                });
+            } else if (property.dimension) {
+                if (property.dimension.range) {
+                    property.dimension.range.forEach(rangePropertyName => {
+                        this._addProperty(baseName, rangePropertyName, false);
+                    });
+                    // add source property too, for stats
+                    this._addProperty(baseName, property.dimension.propertyName);
```
VictorVelarde: This looks common for both branches of the IF

src/renderer/Metadata.js
```diff
@@ -39,24 +70,68 @@ export default class Metadata {
...
+        this.categoryToID.set(category, categoryId);
+        this.IDToCategory.set(categoryId, category);
+        this.numCategories++;
+        return categoryId;
     }
 
-    propertyNames (propertyName) {
-        const prop = this.properties[propertyName];
+    propertyNames (baseName) {
+        // TODO: aggregations and dimensions are specific of Windshaft;
```
VictorVelarde: Are we going to do this anytime soon or is there any blocking for this?

src/renderer/viz/expressions.js
```diff
@@ -481,6 +481,8 @@ import ClusterMode from './expressions/aggregation/cluster/ClusterMode';
 import ClusterSum from './expressions/aggregation/cluster/ClusterSum';
 import ClusterCount from './expressions/aggregation/cluster/ClusterCount';
 
+import ClusterTime from './expressions/aggregation/cluster/ClusterTime';
```
VictorVelarde: Nice!

src/renderer/viz/expressions/Animation.js
```diff
@@ -70,16 +73,17 @@ let waitingForOthers = new Set();
  * @class
  * @api
  */
-export class Animation extends BaseExpression {
+
+class ScalarAnimation extends BaseExpression {
```
VictorVelarde: Why are we renaming this?

src/renderer/viz/expressions/Animation.js
```diff
@@ -70,16 +73,17 @@ let waitingForOthers = new Set();
...
     constructor (input, duration = 10, fade = new Fade()) {
         checkMaxArguments(arguments, 3, 'animation');
 
         duration = implicitCast(duration);
         input = implicitCast(input);
         const originalInput = input;
 
-        if (input.isA(Property)) {
-            input = linear(input, globalMin(input), globalMax(input));
+        if (input.isA(Property) || (input.isA(ClusterTime) && input.type === 'timerange')) {
```
VictorVelarde: We should probably create some external & shared consts for this 'timerange', right?

src/renderer/viz/expressions/Animation.js
```diff
@@ -325,12 +317,83 @@ export class Animation extends BaseExpression {
...
+
         super._bindMetadata(meta);
 
         checkType('animation', 'input', 0, 'number', this._input);
         checkType('animation', 'fade', 2, 'fade', this.fade);
         checkFeatureIndependent('animation', 'duration', 1, this.duration);
     }
 }
+
+class TimeRangeAnimation extends And {
```
VictorVelarde: Oh, I see. IMO we should take this into its own file

jgoizueta:
Hey @davidmanzanares, the main problem I found after merging master wasn't related to the infamous `VariantExpression` as I had assumed after all. It was triggered by the lazy shader compilation (https://github.com/CartoDB/carto-vl/commit/24661c6230f26319294c76df033821923f67934b)

Now shaders aren't compiled the first time that `_runViewportAggregations` is called, and it needs to evaluate the filter expression in order to ignore filtered features.

If an animation (with a time range or a date) is used for the filter, it fails when trying to evaluate the min, max (non numeric is the case that requires evaluation from the bind metadata compilation) because metadata has not been bounded into that expression.

So I have triggered the compilation of the filter expression ahead of time here: https://github.com/CartoDB/carto-vl/pull/991/commits/5edd73a2f1eea9aca0f1bda3617c0fbf95668a79

src/renderer/Dataframe.js
```diff
@@ -277,7 +276,9 @@ export default class Dataframe extends DummyDataframe {
         const WIDTH = this.renderer.gl.canvas.width / window.devicePixelRatio;
         const HEIGHT = this.renderer.gl.canvas.height / window.devicePixelRatio;
 
-        for (let i = 0; i < points.length; i += 6) {
+        // FIXME: points.length includes rejected points (out of tile)
+        // so we use numFeatures here, but should fix the points size
+        for (let i = 0; i < this.numFeatures * 6; i += 6) {
```
jgoizueta: @davidmanzanares I had forgotten about this quickfix I made here, please look at that (points.length seems incorrect because it includes points rejected because they were out of the tile)

debug/nytaxi.html
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    const s = carto.expressions;
+    const source = new carto.source.SQL(`
+      select*from taxi_50k
+    `);
+
+    const viz = new carto.Viz(`
+      // @hour: clusterTime($tpep_pickup_datetime, 'hour')
+      @hour: clusterTime($tpep_pickup_datetime, 'hour', 'America/New_York')
```
VictorVelarde: Just a silly question here. If we are clustering per 'hour' the animation, how can we later get dates like "2016-05-01T06:17:29" not 'rounded' to hours?

debug/timedim.html
```diff
@@ -0,0 +1,131 @@
+<!DOCTYPE html>
```
VictorVelarde: In this example I'm currently getting this error: "CartoParsingError: Invalid function name 'clusterTimeStart'."

debug/timedim.html
```diff
@@ -0,0 +1,131 @@
...
+        }]
+      },
+      center: [0, 0],
+      zoom: 0,
+      dragRotate: false
+    });
+
+    carto.setDefaultAuth({
+      apiKey: 'default_public',
+      user: 'cartovl'
```
VictorVelarde: 'username' is preferred to 'user'

debug/timeweek.html
```diff
@@ -0,0 +1,131 @@
+<!DOCTYPE html>
```
VictorVelarde: It also throws a "CartoParsingError: Invalid function name 'clusterTimeStart'."

src/client/WindshaftMetadata.js
```diff
@@ -0,0 +1,178 @@
...
+            this.propertyKeys.push(propertyName);
+        }
+    }
+
+    _dimensionInfo (propertyName) {
+        const baseName = this.baseName(propertyName) || propertyName;
+        const column = this.properties[baseName];
+        let dimension = null;
+        if (baseName !== propertyName) {
+            if (column.dimension) {
```
VictorVelarde: No else in this nested if, so why not using...

```suggestion
            if (baseName !== propertyName && column.dimension) {
                // whatever
            }
```

src/client/WindshaftMetadata.js
```diff
@@ -0,0 +1,178 @@
...
+                dimension = column.dimension;
+            }
+        }
+        // type of the dataset/query column
+        const baseType = column.type;
```
VictorVelarde: Maybe we could rename some of these variables and save the comments:
```js
const datasetType = column.type:
const sourceType = // the same
const dataframeType = ... 
```

src/client/windshaft-filtering.js
```diff
@@ -169,6 +169,7 @@ class AggregationFiltering {
         if (f.isA(ClusterAvg) || f.isA(ClusterMax) || f.isA(ClusterMin) || f.isA(ClusterMode) || f.isA(ClusterSum)) {
             let p = this._property(f.property);
             if (p) {
+                // FIXME: should we just use the cluster aggr expression?
```
VictorVelarde: ?

src/client/windshaft.js
```diff
@@ -320,25 +331,49 @@ export default class Windshaft {
...
 
-    _adaptMetadata (meta, agg) {
+    _adaptMetadata (meta, agg, MNS) {
         meta.datesAsNumbers = meta.dates_as_numbers;
         const { stats, aggregation, datesAsNumbers } = meta;
         const featureCount = stats.hasOwnProperty('featureCount') ? stats.featureCount : stats.estimatedFeatureCount;
         const geomType = stats.geometryType && adaptGeometryType(stats.geometryType);
 
         const properties = stats.columns;
         Object.keys(agg.columns).forEach(aggName => {
```
VictorVelarde: We could extract both Object.keys blocks to little private functions, to simplify a bit this method, something like:
`_adaptAggregationColumns` and `_adaptAggregationDimensions`

debug/nytaxi.html
jgoizueta: Nope, the progress value is not discrete, it's always kinda continuous 😛
I haven't done any relevant change to `animation`, just added a special case for time range that uses the two combined animations trick internally. So we have an internal progress that it's mapped to the range of values, time in this case. So the progress is time interpolated from start to end, and then the feature values are chunks of this time (ranges), so they hold while the porgress value is inside them (and fade in/out is applied before/after the entire range is entered). I guess this warrants some explanation in the documentation 😅 

I think that when we tackle #1012 this should be reconsidered, and maybe we found it better to have some kind of discrete progress for time range animations. But I think that's out of the scope of this PR

src/codecs/Base.js
```diff
@@ -0,0 +1,67 @@
...
+// * The `internal` encoding is used in Dataframe properties
+//   apt for GPU consumption.
+// * The `external` encoding is the format in which feature
+//   properties are presented to the user
+//   (e.g. `property.eval(feature)` or `globalMin(property)`)
+// The Codec base class serves both as a base for derived codecs.
+// Codecs are associated to source properties.
+// There are two kind of codecs:
+// * Scalar (simple) encoder: source value is encoded as a single internal value
+// * Range encoder: a source value encodes as a pair [lo, hi] of values.
```
VictorVelarde: It would be nice to have the jsdoc already prepared for this group of codec classes

src/client/WindshaftMetadata.js
jgoizueta: :+1:

src/client/windshaft-filtering.js
jgoizueta: thanks for the heads-up, it's just a comment for myself I forgot about; I'll manage it

src/codecs/Base.js
jgoizueta: Of course! I've been really lazy about that 😞

src/codecs/Identity.js
```diff
@@ -0,0 +1,31 @@
+import BaseCodec from './Base';
+
+// The Identity codec class serves
+// as the default, identity encoding, in which all three
+// encodings are exactly the same.
+// TODO: null/NaN support?
```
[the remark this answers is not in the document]
jgoizueta: Oops, will remove (already done)

src/client/windshaft.js
jgoizueta: You mean something like:

```
this._adaptAggregationColumns(aggName => ...
```

Or do you think it would be better just to:

```
this._adaptAggregationColumns().forEach(aggName => ...
```

src/codecs/windshaft/TimeRange.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    inlineInternalMatch (thisValue, otherCodec) {
+        const offset = otherCodec._min.getTime() - this._min.getTime();
+        return `(${thisValue}-${offset.toFixed(20)})`;
+    }
+}
+
+function decodeModal (mode, propertyValue) {
+    // if (propertyValue instanceof Date) {
```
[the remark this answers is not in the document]
jgoizueta: yep

src/client/WindshaftMetadata.js
```diff
@@ -99,7 +99,9 @@ export default class WindshaftMetadata extends MVTMetadata {
         const column = this.properties[baseName];
         let dimension = null;
         if (baseName !== propertyName) {
-            if (column.dimension) {
+            if (baseName !== propertyName && column.dimension) {
+                // whatever
```
VictorVelarde: 👀 I'm afraid I didn't check my suggestion here <<

jgoizueta: :joy: I noticed after accepting the suggestion

PR #1000, opened 2018-09-26
RT3

davidmanzanares:
:frog:


src/sources/RT3.js
```diff
@@ -0,0 +1,130 @@
...
+                dataframe.removePoints(pointsToDelete);
+                dataframe.addPoints(points);
+                points = [];
+                pointsToDelete = [];
+                this._dataLoadedCallback();
+            }, 80);
+            this._rt3Client.setCallbacks({
+                onSet: point => {
+                    total++;
+                    if (point.lat > 180 || point.lat < -180 || point.lon > 190 || point.lon < -190) {
```
VictorVelarde: Why 190 for lon and 180 for lat? Should not these ranges be [90, -90] for lat and [180, -180] for lon?

davidmanzanares: Yes, I think there are multiple issues, but I don't know if we want to dedicate time to this :(

cc: @rochoa

PR #1002, opened 2018-09-29
Raster grids support

jgoizueta:
This is a basic implementation of raster grids visualization for CARTO VL created initially as a 3-day leapfrog project by @VictorVelarde & yours truly.

A generic Grid source has been created to handle raster grids. A popular format for raster grids, GeoTIFF, can be handled with this awesome library: [geotiff.js](https://geotiffjs.github.io/).

A grid occupies a rectangular area (in its reference projection), and comprises one or more data bands. Bands are 2-dimensional arrays of numeric values, all with the same dimensions (width and height). Each numeric value corresponds to a pixel; the grid rectangle is divided into `height` rows and `width` columns of pixels, and the data for each pixel is stored in the band data arrays from N to S and from W to E (by rows).

Two projections are supported: 4326, (WGS84 data in longitude, latitude as degrees), and 3857 (Webmercator coordinates in meters).

Conceptually a grid source consist of pixel features, each of one has a rectangular polygon geometry and properties corresponding to the band values. But the grid is treated internally for performance reasons as a single feature, with the whole rectangle as its geometry, so that we can use the bands as 2-dimensional textures mapped efficiently into it by WebGL.

So this source is quite different from any other an requires special treatment for its properties. Interactivity and expressions such as viewport aggregation as thus not supported at the moment.

The only graphical property that can be use in grid visualizations in the `color`. Each grid band is exposed as a numeric property named `$band0`, `$band1`, etc. Each cell (pixel) of the grid each band property will have the value corresponding to the cell, an those values can be used in CARTO VL expressions to compute the pixel color.

Due to the leapfrog nature of the initial implementation we've incurred into some technical debt. 
* No styling shaders are use for the grid, only the geometry (renderer) shader, so some operations usually done for styling shaders have been replicated in the rendering section.
* No tests yet!
* So, many things might be broken. In particular no attention was paid to providing FeatureIds for the new source.

### Technical notes

The support for longitude-latitude grids has been done by transforming the v texture coordinates at the fragment level. Since both webmercator and the long-lat implicit projection are cylindrical,
they're aligned and there's a linear relation between both in the horizontal (x/longitude) direction.
So we adjust the vertical texture coordinate v to take values at each rendered pixel from the
appropiate texel (grid cell).

Currently the texture mapping uses NEAREST rather than LINEAR filtering and no MIP-mapping.
This is because we use the original, single value band data arrays as textures.
To use LINEAR we'd need to use a RGBA texture (so well need to wast 3 of the components).
To use MIPMAPPING we'd need to pad the arrays to make the texture width and height powers of two.
This would enhance the the graphical quality (and the performance), and it would probably be most relevant when using pitch/bearing.

### Pending stuff

* Eliminate technical debt
* Source code documentation, documentation in general, examples
* Support for no-data values (so they're always transparent/ignored)
* Filtering
* Interactivity
* Viewport aggregation expressions
* LINEAR texture filtering (needs to use RGBA textures)
* MIPMAPPING (needs texture sizes to be powers of two )
* Source to use raster data from PostGIS (CARTO accounts)


jgoizueta:
Note: the tests are broken because they still use geotiff; geotiff usage should be replaced by generated grids like [here](https://github.com/CartoDB/carto-vl/blob/099d2e44cdfb1533483e6089c14fde0ec98b244a/debug/gridtest.html#L71-L120)

jgoizueta:
The rotation/tilt wasn't hard at all. It is available in PR #1003 which is based on the PR #984 for rotation/tilt

So, when #984 is merged, this should be replaced by #1003 By the way, I think that if we just add some tests this is mergeable (and the rest of Pending stuff can be handled later in other tickets).

jgoizueta:
Rotation/tilt changes has been brought here.

PR #1003, opened 2018-10-01
Raster rotation/tilt transformations

jgoizueta:
This adapts the raster rendering for the new matrix transformations.

It was easy! which is good indicator of the quality of https://github.com/CartoDB/carto-vl/pull/984

There was some confusion because we were setting the grid dataframe center to the center of the grid, but it should actually always be 0,0 (the grid is now positioned with `gridOffset`. Maybe we could change this and remove the `gridOffset` usage in the shader, restore the dataframe center and let the transformation matrix do all the work.


jgoizueta:
See #1002 for context

jgoizueta:
continue in #1002

Issue #969, opened 2018-09-18
Guides – Styling your visualization using Expressions


rochoa:
Recommendation: if possible, reuse content from _Introduction to expressions, interpolation, styling and image files_.
Note: add a visual example for each section.

## Proposed structure 
1. What's a visualization property? (link to reference)
1. What's an expression? (link to reference)
1. Playing with colors
    * Chaning color
        - Color spaces
    * Stroke property  
1. Understanding Opacity
    (link to advanced section)
    * Adding opacity
        - Opacity vs RGBA color
1. Using images as markers (link to advanced)
    * Images as symbols
    * Built-in icons
1. Transformation and Placement (better name) related: https://github.com/CartoDB/carto-vl/issues/822
    * Translating the geometry (link to examples)
    * Placing images
1. Changing the resolution
    (only mention the property, link to advanced) 
1. Filtering data
    (only mention the property, link to advanced)
1. Defining variables
    (only mention the property, link to advanced)
1. Ordering
    (only mention the property, link to advanced)

davidmanzanares:
If this is one of the first guides, I think that images and transformations are too advanced for it. Although I agree we should mention and link them.

davidmanzanares:
Fixed by #1005

Issue #970, opened 2018-09-18
Guides – Data-driven visualizations


rochoa:
## Proposed structure

1. What's a ramp? (example with a simple categorical property)
1. Using input expressions
    * Using another expression in a ramp
    * Nesting expressions (link to advanced "the power of nested expressions")
1. Thematic visualizations (all the examples must use only _colors_ as output)
    * Visualizing Numeric Data
      * Classification
      * Interpolation
    * Visualizing Categorical Data
      * Top
      * Buckets
1. Using color palettes, numbers and images
    * CartoColor and custom colors
    * Numbers
    * Images
1. Getting ramp information
    * Using legends## Proposed structure
1. What's a ramp? (example with a simple categorical property)
1. Using input expressions
    * Using another expression in a ramp
    * Nesting expressions (link to advanced "the power of nested expressions")
1. Thematic visualizations (all the examples must use only _colors_ as output)
    * Visualizing Numeric Data
      * Classification
      * Interpolation
    * Visualizing Categorical Data
      * Top
      * Buckets
1. Using color palettes, numbers and images
    * CartoColor and custom colors
    * Numbers
    * Images

~~Getting ramp information~~
    ~~* Using legends~~ --> Moved to a separated guide

davidmanzanares:
Fixed by https://github.com/CartoDB/carto-vl/pull/1005

PR #1004, opened 2018-10-01
WIP: Guide viz expr

davidmanzanares:
https://github.com/CartoDB/carto-vl/issues/969

@makella this is just a stub, but I prefer to get some feedback before continuing. Feel free to add new stuff or talk to me directly.


makella:
ok! taking a look now

makella:
@davidmanzanares ok, looked through, here are a couple of comments:

- in the outline, I don't see an item related to `width` so we should add that
- `strokeWidth` is inside of color, we probably want to take that out from color 
- what I don't see is what is outlined more in the reference which is using attributes and expressions. - also the expression types in the reference: numbers, categories, colors aren't outlined here, but I'm guessing that is covered in "What is an expression" (?)
- another thing that could be useful is to know about any caveats (for example, `order` is only for points that are sized by a numeric attribute...)

I know we have a guide for data-driven styling (https://github.com/CartoDB/carto-vl/issues/970) but the topics covered there are more about `ramp`.

In terms of a map-guided example, what are you thinking of using?

makella:
@davidmanzanares sorry! i didn't actually read your guide... I only referred to the outline linked. :(

Reading now!

examples/guides/viz-properties/geom-types.html
```diff
@@ -0,0 +1,104 @@
...
+    map.addControl(nav, 'top-left');
+
+    // Define user
+    carto.setDefaultAuth({
+      user: 'cartovl',
+      apiKey: 'default_public'
+    });
+
+    // Define layer
+    const source = new carto.source.Dataset('ne_10m_populated_places_simple');
```
elenatorro: use `populated_places` instead

examples/guides/viz-properties/accessing-viz-properties.html
```diff
@@ -0,0 +1,85 @@
...
+    map.addControl(nav, 'top-left');
+
+    // Define user
+    carto.setDefaultAuth({
+      user: 'cartovl',
+      apiKey: 'default_public'
+    });
+
+    // Define layer
+    const source = new carto.source.Dataset('ne_10m_populated_places_simple');
```
elenatorro: use `populated_places` instead

davidmanzanares:
Merged into https://github.com/CartoDB/carto-vl/pull/1004

PR #1010, opened 2018-10-02
Fix ramp others

davidmanzanares:
As discovered by @ztephm `ramp` `others` parameter doesn't work well in some circumstances.

One tested example is: `ramp(..., ..., 5)`, which fails due to "invalid types".

This happens because `5` is not implicitly cast to `number(5)`.

While adding regression tests for it I discovered another bug: `others` don't work with feature-dependent expressions because `this.others._bindMetadata(metadata);` was called after the type check, which makes the check fail.


elenatorro:
🚀

elenatorro:
Good catch @ztephm! Thanks ✨

PR #1013, opened 2018-10-03
Enable pitch

davidmanzanares:
- Enable pitch & bearing on examples
- Fix Changelog, mentioning pitch & bearing

### DIFF
--- a/debug/MVT.html
+++ b/debug/MVT.html
@@ -40,8 +40,8 @@
       style: carto.basemaps.voyager,
       center: [-73.97720677029952, 40.77784846220871],
       zoom: 13,
-      dragRotate: false,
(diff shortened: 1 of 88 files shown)


debug/MVT.html
```diff
@@ -40,8 +40,8 @@ <h1>Add layer</h1>
       style: carto.basemaps.voyager,
       center: [-73.97720677029952, 40.77784846220871],
       zoom: 13,
-      dragRotate: false,
-      touchZoomRotate: false,
+
```
jgoizueta: Shouldn't we be explicit and have `dragRotate: true`?

davidmanzanares: I'd prefer to reduce noise

PR #1015, opened 2018-10-03
Fix request tiles & Fix interactivity corner case & Fix perf regression

davidmanzanares:
- Layer.update() was re-instancing, but it wasn't requesting new tiles unless the user moved the map
- Interactivity could throw an error (ignored) if a tile loaded and a user interactivity event was fired before it was rendered for the first time.
- Fix performance regression of dataframe.js with viewport agg. I don't know what happened since this was tested and optimized in the original rotation branch. But when I re-tested it, the performance had regressed.
- Fix interactivity with polygons with offset

### DIFF
--- a/debug/viewportAggPluto.html
+++ b/debug/viewportAggPluto.html
@@ -1,0 +1,79 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Add layer | CARTO</title>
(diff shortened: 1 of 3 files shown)


elenatorro:
At a first glance, the code seems fine to me. Please update the CHANGELOG :)

Issue #1022, opened 2018-10-05
Allow username as parameter when using auth objects


VictorVelarde:
Current CARTO VL uses 
```js
{
  apiKey: 'YOUR_API_KEY_HERE',
  user: 'YOUR_USERNAME_HERE'
}
```

But in CARTO.js we use `username` not `user`.
```
{
  apiKey: 'YOUR_API_KEY_HERE',
  username: 'YOUR_USERNAME_HERE'
```

So let's add support for **username** for consistency (without removing current `user` for backwards compatibility)

PR #1024, opened 2018-10-05
Allow `username` as parameter when using auth objects

VictorVelarde:
Related with #1022 

This PR adds `.username` as param in auth (the same as in CARTO.js). Current `.user` is kept for backwards compatibility.

I have also updated the current guides and examples in master to display the preferred option: **username**


src/setup/auth-service.js
```diff
@@ -45,7 +45,7 @@ function checkAuth (auth) {
     if (!util.isObject(auth)) {
         throw new CartoValidationError(`${cvt.INCORRECT_TYPE} 'auth' property must be an object.`);
     }
-    auth.username = auth.user; // API adapter
+    auth.username = util.isDefined(auth.username) ? auth.username : auth.user; // backwards compatibility
```
davidmanzanares: `util.isDefined(` could we deprecate it?

VictorVelarde: I have just added it to utils... as I was using `!util.isUndefined` ... but I can remove it if you don't like it very much! don't you like that sugar 😄 ?

davidmanzanares: Well... I don't like it much. I think it is more verbose than directly comparing it, and, if the language provides a reasonable to do it, I prefer that over an util

VictorVelarde: I see your point, you prefer not using a new util. I prefer `isDefined` to `!isUndefined`, cause IMO negatives are a bit harder to understand than positives. 

But it is not a big deal, so I'm gonna remove the new method. 
I'll also flip the if, cause I always try to express conditionals as positives.

Thx for your review!

Issue #1026, opened 2018-10-05
Basemaps render tests fail randomly


VictorVelarde:
Some basemaps tests are randomly failing

As discussed with @davidmanzanares , it looks like MGL loaded event is fired when the map loads, but it doesn't wait till the label fade in is ended

VictorVelarde:
Merged 🚀

PR #1027, opened 2018-10-05
Deal with failing basemaps render tests

VictorVelarde:
Related to https://github.com/CartoDB/carto-vl/issues/1026


VictorVelarde:
I think the problem with the current test-render error is that the original images where generated without a delay, so they don't have the labels. They should be regenerated

davidmanzanares:
I don't have a problem to regenerate the images, but... Maybe we are testing MGL too much. The covered functionality is important, but at the same time, errors in the basemaps would be caught by our "smoke" tests.

Issue #898, opened 2018-08-28
Make documentation available for different versions


rochoa:
Steps:

* [x] Have different versions of the documentation available at https://carto.com/developers/carto-vl/
* [x] Add hook to have a CDN per branch (we would need to modify the `copy` task in the developer center)

elenatorro:
🚀special thanks to @csubira, awesome work! 💪

rochoa:
Thanks, @elenatorro and @csubira. ❤️ 

Having documentation for different versions is the way to go. 💪

PR #1029, opened 2018-10-08
Add script to publish branches to cdn

elenatorro:
Related with **Make documentation available for different versions** https://github.com/CartoDB/carto-vl/issues/898

In order to be able to test different branches both locally and in the developer center, we have added a script to publish a branch in our CDN.


scripts/cdn/branch/publish.js
```diff
@@ -0,0 +1,54 @@
...
+let secrets = JSON.parse(fs.readFileSync('secrets.json'));
+if (!secrets ||
+    !secrets.AWS_USER_S3_KEY ||
+    !secrets.AWS_USER_S3_SECRET ||
+    !secrets.AWS_S3_BUCKET) {
+    throw Error('secrets.json content is not valid');
+}
+
+const branchName = require('current-git-branch');
+const branch = JSON.parse(fs.readFileSync('package.json')).branch || branchName();
```
davidmanzanares: I don't understand `JSON.parse(fs.readFileSync('package.json')).branch`, we never define the branch there, right?

scripts/cdn/branch/publish.js
```diff
@@ -0,0 +1,54 @@
...
+});
+
+uploadFiles(branch);
+
+function uploadFiles (branch) {
+    console.log('Publishing branch:', branch);
+
+    let uploader = client.uploadDir({
+        localDir: 'dist',
+        deleteRemoved: true,
```
davidmanzanares: Is this safe? I think this flag was blamed before for removing the wrong files

scripts/cdn/branch/unpublish.js
```diff
@@ -0,0 +1,49 @@
...
+        secretAccessKey: secrets.AWS_USER_S3_SECRET
+    }
+});
+
+deleteFiles(branch);
+
+function deleteFiles (branch) {
+    console.log('Unpublishing branch:', branch);
+
+    let uploader = client.deleteDir({
```
davidmanzanares: I'm super scared by this

scripts/cdn/branch/publish.js
elenatorro: Nope, we don't. That's why I added a package to get the current branch. The purpose of this is to allow us to use the branch name we want by adding it in the package json, if you prefer I can add it as a script parameter/option instead.

scripts/cdn/branch/publish.js
elenatorro: didn't know that it was removed, I can remove it again.

scripts/cdn/branch/unpublish.js
elenatorro: Why? We can add a condition checking the `branch` parameter is not a version. What do you think?

scripts/cdn/branch/publish.js
davidmanzanares: To be honest, I wouldn't bother in supporting it in any way

scripts/cdn/branch/publish.js
davidmanzanares: :+1:

scripts/cdn/branch/unpublish.js
davidmanzanares: I would just leave the published branches in S3. If we don't want to put so much noise there, we could use a subfolder in which we put all the branches (but not the real versions)

elenatorro: In my opinion it doesn't add any noise and it's very useful to keep things clean. If we don't want to have this because it's dangerous, for me it's ok, but if it's because it adds 'noise' I don't agree.

scripts/cdn/branch/publish.js
```diff
@@ -0,0 +1,52 @@
...
+
+function uploadFiles (branch) {
+    console.log('Publishing branch:', branch);
+
+    let uploader = client.uploadDir({
+        localDir: 'dist',
+        s3Params: {
+            ACL: 'public-read',
+            Bucket: secrets.AWS_S3_BUCKET,
+            Prefix: 'carto-vl/' + branch + '/'
```
rochoa: What about using a specific `branches` subdirectory?

scripts/cdn/branch/publish.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Publish packages to our CDN
+ */
+
+// Load secrets file
+
+let fs = require('fs');
+
+let secrets = JSON.parse(fs.readFileSync('secrets.json'));
```
rochoa: For JSON files, you can use `require`.

Issue #1038, opened 2018-10-09
clusterMode is broken


davidmanzanares:
See http://127.0.0.1:8080/examples/editor/index.html#eyJhIjoibmVfMTBtX3BvcHVsYXRlZF9wbGFjZXNfc2ltcGxlIiwiYiI6IiIsImMiOiJjYXJ0b3ZsIiwiZCI6Imh0dHBzOi8ve3VzZXJ9LmNhcnRvLmNvbSIsImUiOiJjb2xvcjogcmFtcChjbHVzdGVyTW9kZSgkbmFtZSksIGJvbGQpIiwiZiI6eyJsbmciOjMzLjMyNzMxNTM1NDgyOTU4LCJsYXQiOjIuODQyMTcwOTQzMDQwNDAxZS0xNH0sImciOjEuMDU1MjgyNDE1MDQzMTk4NSwiaCI6IkRhcmtNYXR0ZXIiLCJpIjoiZGF0YXNldCJ9

VictorVelarde:
@davidmanzanares It looks like this should be closed, right?

davidmanzanares:
Yes, sorry

PR #1040, opened 2018-10-09
Fix master

davidmanzanares:
Fix:
- Examples
- blendToViz
- ramp(clusterMode(.... https://github.com/CartoDB/carto-vl/issues/1038


examples/guides/animation/step-1.html
```diff
@@ -3,19 +3,13 @@
 
 <head>
     <!-- Include CARTO VL JS -->
-    <script src="../../../../dist/carto-vl.js"></script>
+    <script src="../../../dist/carto-vl.js"></script>
```
VictorVelarde: It's fine to change this to adapt to local carto-vl path. At dev-center should be automatically replaced by the corresponding version

examples/guides/animation/step-1.html
```diff
@@ -3,19 +3,13 @@
...
     <link href="https://libs.cartocdn.com/mapbox-gl/v0.48.0-carto1/mapbox-gl.css" rel="stylesheet" />
     <!-- Make the map visible -->
-    <style>
-        #map {
-            position: absolute;
-            height: 100%;
-            width: 100%;
-        }
-    </style>
+    <link rel="stylesheet" type="text/css" href="../../style.css">
```
VictorVelarde: But I won't change this!

Examples inside guides are meant to be reproduced, step by step, by the users, and adding style.css dependency is more complex than just having a style rule for the map. In the guides I've worked that's enough.

Ping to @elenatorro here to agree on this before changing it

davidmanzanares: Most maps I saw shown visual problems due to incorrect CSS (when the CSS was the inlined one), and I don't think it's possible to have a small CSS for many examples. At least in this way, the CSS doesn't add noise. We are trying to help people with CARTO VL, not with CSS.

VictorVelarde: I would be very careful when changing `step-N.html` examples for the guides, as it usually also requires sync fixes in the content of the .md files

VictorVelarde: We have agreed on going on with this fix, and delay changes & possible refactoring on css. We will try that in #1041

Issue #1039, opened 2018-10-09
Don't force babel usage everywhere


davidmanzanares:
For debugging, babel is a problem since the generated code is sometimes not relatable to the original one and makes debugging very difficult.

![babel](https://user-images.githubusercontent.com/1692175/46662800-ac4fc800-cbbc-11e8-962b-0a55f9d4b0df.png)

rochoa:
How easy is to disable it in your local environment?

davidmanzanares:
On top of that, unit test times have increased from 3.7 seconds to 13.6 seconds

VictorVelarde:
We have removed babel from build:dev

davidmanzanares:
This is still happening when using `build:watch` and in `gh-pages`

VictorVelarde:
@davidmanzanares:
* `build:watch` just uses the development mode, without babel, as it was before (PR https://github.com/CartoDB/carto-vl/pull/1044 reverted that AND also fixed `devtool` param with 'source-map'). what's the error it presents?
* with "gh-pages" do you refer to `ghpublish` task? it uses inside `yarn build` which launches both dev and min builds... what is expected from the build:dev?

davidmanzanares:
Hmu...

Maybe I was in an outdated branch that didn't get your commits yet, sorry.

PR #1044, opened 2018-10-09
Remove babel-loader from development build

VictorVelarde:
### Related to https://github.com/CartoDB/carto-vl/issues/1039

Remove babel-loader from webpack configuration in `build:dev`

### DIFF
--- a/webpack/webpack.config.js
+++ b/webpack/webpack.config.js
@@ -10,22 +10,10 @@ module.exports = {
         library: 'carto',
         libraryTarget: 'umd'
     },
-    devtool: 'sourcemap',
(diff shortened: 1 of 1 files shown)


webpack/webpack.config.js
```diff
@@ -10,22 +10,10 @@ module.exports = {
         library: 'carto',
         libraryTarget: 'umd'
     },
-    devtool: 'sourcemap',
+    devtool: 'source-map',
```
VictorVelarde: This seems to be the correct value, as described in https://webpack.js.org/configuration/devtool

PR #1050, opened 2018-10-10
Take clusterCount() into viewport calculations

davidmanzanares:
This fixes the `viewportSum` problem, but requires further effort:
- `ViewportPercentile.` will be super slow after this when the aggregations kick in a lot. This probable can be solved by using something like viewportHistogram implementation.
- `viewportFeatures` may be renamed to accommodate the fact that each "feature" is now a cluster. Obviously, this was the case earlier, but earlier, all viewport functions assume 1 cluster 1 feature. But now, viewportFeatures is the only viewport function that doesn't take the clustering into account directly (it is left to the developer). And we shouldn't change that (by making a super long list) since it will be a performance issue.

cc: @jgoizueta

### DIFF
--- a/src/renderer/viz/expressions/aggregation/viewport/ViewportAvg.js
+++ b/src/renderer/viz/expressions/aggregation/viewport/ViewportAvg.js
@@ -46,7 +46,7 @@ export default class ViewportAvg extends ViewportAggregation {
         const propertyValue = this.property.eval(feature);
 
         if (!Number.isNaN(propertyValue)) {
-            const clusterCount = feature._cdb_feature_count;
+            const clusterCount = feature._cdb_feature_count || 1;
             this._count += clusterCount;
             this._sum += clusterCount * propertyValue;
         }
--- a/src/renderer/viz/expressions/aggregation/viewport/ViewportAvg.js
+++ b/src/renderer/viz/expressions/aggregation/viewport/ViewportAvg.js
@@ -46,8 +46,9 @@ export default class ViewportAvg extends ViewportAggregation {
         const propertyValue = this.property.eval(feature);
 
         if (!Number.isNaN(propertyValue)) {
-            this._count++;
-            this._sum += propertyValue;
+            const clusterCount = feature._cdb_feature_count;
+            this._count += clusterCount;
+            this._sum += clusterCount * propertyValue;
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/renderer/viz/expressions/aggregation/viewport/ViewportAvg.js
```diff
@@ -46,8 +46,9 @@ export default class ViewportAvg extends ViewportAggregation {
         const propertyValue = this.property.eval(feature);
 
         if (!Number.isNaN(propertyValue)) {
-            this._count++;
-            this._sum += propertyValue;
+            const clusterCount = feature._cdb_feature_count;
```
VictorVelarde: We have a convenient constant CLUSTER_FEATURE_COUNT available at schema.js

src/renderer/viz/expressions/aggregation/viewport/ViewportPercentile.js
```diff
@@ -32,52 +32,65 @@ export default class ViewportPercentile extends ViewportAggregation {
...
+            const unclampedIndex = Math.ceil(this.percentile.eval(feature) / 100 * this._total) - 1;
+            const index = clamp(unclampedIndex, 0, this._total - 1);
+            const array = [...this._map.entries()];
+            array.sort((a, b) => a[0] - b[0]);
+            let accum = 0;
+            for (let i = 0; i < array.length; i++) {
+                accum += array[i][1];
+                array[i][1] = accum;
+            }
+            this._value = binarySearch(array, index, 0, array.length - 1);
```
VictorVelarde: Nice refactoring! Did it improve the perf notably?

src/renderer/viz/expressions/aggregation/viewport/ViewportPercentile.js
```diff
@@ -32,52 +32,65 @@ export default class ViewportPercentile extends ViewportAggregation {
...
-        super._preDraw(...args);
+        const clusterCount = feature[CLUSTER_FEATURE_COUNT] || 1;
+        this._map.set(v, (this._map.get(v) || 0) + clusterCount);
+        this._total += clusterCount;
     }
 }
 
-function _getPercentile (value, length) {
-    return Math.floor(value / 100 * length);
+function binarySearch (array, index, begin, end) {
```
VictorVelarde: If this can be reused, it could be inside some utils.js

Issue #1058, opened 2018-10-15
Start using Mapbox GL v0.50.0


rochoa:
The team at Mapbox has already released a new [v0.50.0](https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.50.0) release, this is the first official release that supports Custom Layers.

We need to:
 - [ ] Test it works as expected with our current CARTO VL version.
 - [ ] Update the documentation to reflect you can use CARTO VL with the standalone Mapbox GL library from version v0.50.0.
 - [ ] Update all guides to use the v0.50.0 version.
 - [ ] Update all examples to use the v0.50.0 version.

PR #1064, opened 2018-10-16
Use unpatched MGL 0.50.0

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/1058


davidmanzanares:
`<!-- Include Mapbox GL -->`

Totally disagree with that. I don't think that repeating the same comment over a hundred examples improves things, particularly since the script tag is pretty self-explanatory.

> It's quite obvious for a dev,

It is, and for non-devs, they will copy and paste it. We should not try to teach HTML/JS in our examples.

Even on the first Google Maps API example, they don't explain the script tag, see https://developers.google.com/maps/documentation/javascript/tutorial

VictorVelarde:
The GMaps is a good example!

I'm ok with simplifying / removing basic comments (like import and the one for div-map). Maybe we could just keep them in the "first" getting-started guide, and that's it. 

I just wanted to point out that @makella is working in the examples with that premise, of being very-explicit in the comments, and that we should be homogeneous.

Issue #1059, opened 2018-10-15
Legend title is broken on minified versions


davidmanzanares:
As @VictorVelarde pointed out in the ramp guides the examples legend title is broken:
` e($dn, e(20), e(254))`

This happens due to minification, the same code generates the following title with the unminified bundle;
`linear($dn, globalMin($dn), globalMax($dn))`

The reason for this is on `base.js`, line:
`        this.expressionName = _toCamelCase(this.constructor.name);`
`this.constructor.name` suffers the minification unfortunately.

VictorVelarde:
This looks related to https://github.com/webpack/webpack/issues/6841
I'll have a look at it

VictorVelarde:
The situation here is this:
* webpack 4 uses by default Uglifyjs when running in **production** mode (which we are using for our carto-vl.min.js). See https://webpack.js.org/configuration/optimization/
* we use this.constructor.name at runtime, which has been mangled by the minify process.
* it seems this is pretty common and three options are usually recommended:
    - *disable* minify (in our case that doesn't seem feasible, as the min.js grows to 1MB!)
    - change how minify works: configure Uglify with *keep_fnames* --> 422 KB (It seems there was an ongoing work on a more detailed *keep_classnames* param that never got into Uglify:master)
    - change the use of constructor.name with a string const added to each class

davidmanzanares:
422 KB from how many KBs originally?

PR #1066, opened 2018-10-16
Keep function names in build for production mode

VictorVelarde:
Fix https://github.com/CartoDB/carto-vl/issues/1059


debug/legendDataName.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
```
davidmanzanares: I wouldn't add another debug example. It is already horrible to do the smoke tests. This issue is indirectly covered by the guide examples, so I don't think this is necessary

VictorVelarde: Done!

Issue #994, opened 2018-09-24
Clustering with viewport agg doesn't work


davidmanzanares:
A quick research showed that it may be related to the lack of a name (of the clustered property) when generating the feature that is passed in the `accum` methods

davidmanzanares:
You can take a look at `file:///home/dv/github/renderer-prototype/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoiQGE6IGNsdXN0ZXJBdmcoJGFtb3VudClcbkBjOiBjbHVzdGVyTWF4KCRhbW91bnQpXG53aWR0aDogQGEvMTAwXG5jb2xvcjogcmdiKDIwNCwwLDApXG5AYjogdmlld3BvcnRBdmcoY2x1c3RlckF2ZygkYW1vdW50KSkiLCJmIjp7ImxuZyI6Mi4xNjA4NzY4MTg3ODQ0NTksImxhdCI6NDEuMzk1OTIzNDMxNzQwNDR9LCJnIjoxMi43MzI2OTIzNzA4MjM5NTcsImgiOiJEYXJrTWF0dGVyIiwiaSI6ImRhdGFzZXQifQ==`

and put `layer.viz.variables.b.value` on Chrome's console to debug it.

The issue happens when things like `viewportAvg(clusterAvg($amount))` are used, but this is not the only problem, for example, Interactivity with `@b: (clusterAvg($amount))` is broken too, and for the same reason.

The problem happens on Dataframe.js, method `_genFeatureClass`. In it, there is a loop over the **Metadata** properties to retrieve (later) the properties from the Dataframe. This logic is wrong because a Metadata property may have multiple Dataframe properties due to multiple aggregations (avg and min for example).

This also fails when there is only one clustering in use because the name on the Dataframe includes the aggregation, but the name on the Metadata is just the original name of the column.

`viewportFeatures` suffers a similar problem, and in this case I think that an API change is the only possible way to solve it.

davidmanzanares:
`viewportFeatures` should be fixed with the time aggregation PR of @jgoizueta 

The rest should be fixed by #1071

VictorVelarde:
[...]

PR #1071, opened 2018-10-17
Fix viewport + clustering

davidmanzanares:
Partially address https://github.com/CartoDB/carto-vl/issues/994

This doesn't fix `viewportFeatures`, but it should fix the original issue


src/renderer/Dataframe.js
```diff
@@ -497,6 +497,20 @@ export default class Dataframe extends DummyDataframe {
         const getters = {};
         for (let i = 0; i < this.metadata.propertyKeys.length; i++) {
             const propertyName = this.metadata.propertyKeys[i];
+            if (this.metadata.properties[propertyName].aggregations) {
```
VictorVelarde: I think this could be refactored to a new function called **_buildGetters**, with an objective of populating the `getters` object.

It could be something like:
```suggestion
 const metadata = this.metadata;
 const dataframe = this._dataframe;
 const getters = this._buildGetters(metadata, dataframe);
 
_buildGetters (metadata, dataframe) {
    // code with the for...
}
```

It could also have no input params, as both are available through _this_.

test/acceptance/e2e/pop_density_points/scenario.js
```diff
@@ -14,15 +14,15 @@ carto.setDefaultConfig({
...
-const s = carto.expressions;
-const $dn = s.property('dn');
-const viz = new carto.Viz({
-    width: s.scaled(10, 4),
-    color: s.ramp(s.linear($dn, 1, 300), s.palettes.PRISM),
-    strokeColor: s.rgba(0, 0, 0, 0.2),
-    strokeWidth: 1,
-    filter: s.between($dn, 100, 140)
-});
+const viz = new carto.Viz(`
```
VictorVelarde: Nice, I prefer this syntax!

src/renderer/Dataframe.js
davidmanzanares: :+1:

PR #1074, opened 2018-10-17
Refactor Layer.js and related

davidmanzanares:
- [x] Remove dead code (mainly stuff related to pitch & bearing and the new MGL integration)
- [x] Remove complexity:
  - [x] Flags nightmare
  - [x] Update and blendTo nightmare
  - [x] Stablish clear expectations around update/blendTo/blendToViz when racing between each other
- [x] Reduce unneeded shader compilations
- [x] Improve tests
- [x] Review and document the new API details


davidmanzanares:
## The good

I think that at this particular moment (commit) this branch is ready to be merged.

In particular, complexity and lines of code have been removed:
```
git diff --stat master -- src/
 src/Layer.js                         | 385 +++++++++++++++++++++++++-------------------------------
 src/Viz.js                           |  83 +++++++-----
 src/constants/layer.js               |   4 -
 src/renderer/Dataframe.js            |   9 --
 src/renderer/Renderer.js             |  28 -----
 src/renderer/viz/expressions/base.js |   7 +-
 src/utils/util.js                    |  18 ---
 7 files changed, 229 insertions(+), 305 deletions(-)
```

While specifying, documenting and implementing the behavior on some corner cases regarding concurrent calls to *update-like* methods.

Tests have also been improved and expanded:
```
git diff --stat master -- test/
 test/CartoMap.js                                                 |  22 -----------
 test/integration/render/scenarios/belongs/nin/scenario.js        |   1 +
 test/integration/render/scenarios/blendto/rejects/reference.png  | Bin 0 -> 2065 bytes
 test/integration/render/scenarios/blendto/rejects/scenario.js    |  17 +++++++++
 test/integration/render/scenarios/blendto/resolves/reference.png | Bin 0 -> 2081 bytes
 test/integration/render/scenarios/blendto/resolves/scenario.js   |  19 ++++++++++
 .../render/scenarios/blendtoviz/changes-viz/reference.png        | Bin 0 -> 2081 bytes
 .../render/scenarios/blendtoviz/changes-viz/scenario.js          |  18 +++++++++
 .../render/scenarios/blendtoviz/rejected-by-newer/reference.png  | Bin 0 -> 2030 bytes
 .../render/scenarios/blendtoviz/rejected-by-newer/scenario.js    |  47 ++++++++++++++++++++++++
 .../render/scenarios/layer-update/changes-source/reference.png   | Bin 0 -> 1697 bytes
 .../render/scenarios/layer-update/changes-source/scenario.js     |  15 ++++++++
 .../render/scenarios/layer-update/changes-viz/reference.png      | Bin 0 -> 2030 bytes
 .../render/scenarios/layer-update/changes-viz/scenario.js        |  13 +++++++
 .../scenarios/layer-update/rejected-by-newer/reference.png       | Bin 0 -> 1903 bytes
 .../render/scenarios/layer-update/rejected-by-newer/scenario.js  |  51 ++++++++++++++++++++++++++
 .../layer-update/uninterruptible-by-blendto/reference.png        | Bin 0 -> 2030 bytes
 .../layer-update/uninterruptible-by-blendto/scenario.js          |  42 +++++++++++++++++++++
 .../render/scenarios/regression/cartocolor-top/reference.png     | Bin 5289 -> 8307 bytes
 .../render/scenarios/regression/cartocolor-top/scenario.js       |   6 +--
 test/integration/user/test/layer.test.js                         |  55 ++++++++++++++--------------
 test/unit/layer.test.js                                          |   5 ++-
 test/unit/viz.test.js                                            |  11 ------
 23 files changed, 256 insertions(+), 66 deletions(-)
```

In general, I consider that with these changes, usage of `update` and `blendToViz` is much safer and its implementation more robust.

## The bad

However, there is still one piece that I feel it is quite fragile in its current state: `blendTo`. The reason for this is that the implementation makes the API particularly unsafe (in contrast with `blendToViz` and `update`).

The important detail of the implementation is that `blendTo` modifies the `Viz` tree **in-place** instead of using a new `Viz` and taking the general `update` path in which things are validated and then, if everything went ok, committed (resolving a promise too).

This is not easily fixable by generating a new Viz at `blendTo` to re-use the `update` path. Since this will have important un-wanted consequences: users will still have access to the previous `Viz` object and modifications on that one wouldn't affect the current Viz.

## The question

I think that, at some point, we should research possible solutions for `blendTo`. However, it may not be a priority right now.

I think that usage of `blendTo` is particularly useful for custom applications with a limited set of "paths" instead of more general applications like editors in which the end-users can provide its own styles.

In this context, invalid usage of `blendTo` is typically due to a programmer's error and the ability to catch the error and recover from it (something supported with `update` and `blendToViz`) is not that important.

I think that, for the time being, we can just document `blendTo` limitations.

Do you agree, @rochoa ?

rochoa:
Yes, we could work and improve `blendTo` in the future. As far as I see, the `blendTo` doesn't return anything. I don't know if returning the _new_ viz would be enough to cover the problem you are mentioning. The good thing is that we can decide that yet. For now, let's :ship: this one.

davidmanzanares:
:+1:

Right now it is returning a promise, but that is only rejected on parser errors.

src/Layer.js
```diff
@@ -43,56 +48,78 @@ export default class Layer {
...
+    set visible (visible) {
+        this.map.setLayoutProperty(this.id, 'visibility', visible ? 'visible' : 'none');
+        this._visible = visible;
     }
 
     /**
-     * Get layer visibility. Can be true or false.
-     * @readonly
+     * Change layer visibility to visible
+     *
```
VictorVelarde: Maybe we could add a small note about what's a mayor and a minor change / UID

src/Layer.js
```diff
@@ -154,71 +181,113 @@ export default class Layer {
...
-
-        if (this._source !== source) {
-            this._freeSource();
-        }
-
-        this._source = source;
-        this._noFirstRequestData = false;
-        this.requestData();
-
         viz.setDefaultsIfRequired(this.metadata.geomType);
```
VictorVelarde: This block of sentences regarding to updating the viz can be easily extracted to a new `_updateViz()` to reduce a bit the size of this method

src/Layer.js
```diff
@@ -457,41 +458,12 @@ export default class Layer {
         return feature;
     }
```
VictorVelarde: Yeah, it looks like a pure function

src/Viz.js
```diff
@@ -310,39 +310,60 @@ export default class Viz {
...
+    get lineShader () {
+        return this._compileShader('lineShader', { vertexShader: lineVertexShaderGLSL, fragmentShader: lineFragmentShaderGLSL },
+            { transform: this.transform });
+    }
+    get polygonShader () {
+        return this._compileShader('polygonShader', { vertexShader: polygonVertexShaderGLSL, fragmentShader: polygonFragmentShaderGLSL },
+            { transform: this.transform });
+    }
+
+    _compileShader (shaderName, GLSL, expr) {
```
VictorVelarde: Nice use of properties

src/Layer.js

src/Viz.js
```diff
@@ -310,39 +310,60 @@ export default class Viz {
...
+        if (!this['_' + shaderName]) {
+            this._getRootExpressions().forEach(expr => expr._bindMetadata(this.metadata));
+            checkVizPropertyTypes(this);
+            this['_' + shaderName] = compileShader(this.gl, GLSL, expr, this);
         }
+        return this['_' + shaderName];
     }
 
     replaceChild (toReplace, replacer) {
```
VictorVelarde: This could be refactored to something like (I haven't tested it, though):
```suggestion
	replaceChild (toReplace, replacer) {
        if (Object.values(this.variables).includes(toReplace)) {
            const varName = Object.keys(this.variables).find(varName => this.variables[varName] === toReplace);
            this.variables[varName] = replacer;
            replacer.parent = this;
            replacer.notify = toReplace.notify;
        } else {
        	const properties = ['color', 'width', 'strokeColor', 'strokeWidth', 'filter', 'symbol', 'symbolPlacement', 'transform'];
            properties.forEach((property) => {
            	if (toReplace === this[property]){
                	this[property] = replacer;
                    replacer.parent = this;
                    replacer.notify = toReplace.notify;
                }
            });
            
            if (!properties.includes(toReplace)){
            	throw new CartoRuntimeError('No child found');
            }
        }
  	}
```

src/Layer.js
davidmanzanares: I think that would suggest that the new method can be used without the `_update` method when the source doesn't change, but that is no true since a change in the `viz` may require new re-instantiations in the source.

src/Layer.js
```diff
@@ -274,62 +333,15 @@ export default class Layer {
...
-    }
-
-    async requestData () {
-        if (!this.metadata || !this._visible) {
-            return;
-        }
-        this._needRefresh();
-    }
-
     hasDataframes () {
```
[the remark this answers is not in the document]

src/renderer/Renderer.js
```diff
@@ -62,34 +62,6 @@ export default class Renderer {
         this._initGL(gl);
     }
 
-    /**
```
VictorVelarde: this wasn't being used...? or is this somehow provided now by the layer?

test/integration/render/scenarios/belongs/nin/scenario.js
```diff
@@ -12,5 +12,6 @@ const layer = new carto.Layer('layer', source, viz);
 
 layer.addTo(map);
 layer.on('loaded', () => {
+    console.log('wad');
```
VictorVelarde: 'wad' stands for...?

src/Layer.js
```diff
@@ -371,84 +383,73 @@ export default class Layer {
...
-            }
-            this._fireUpdateOnNextRender = true;
-        } else if (!this._noFirstRequestData && this._source && this._visible) {
-            this._source.requestData(this._getZoom(), this._getViewport());
-            this._noFirstRequestData = true;
+        this.renderer.matrix = matrix;
+        if (this._source && this.visible) {
+            this._source.requestData(this._getZoom(), this._getViewport(matrix));
         }
     }
```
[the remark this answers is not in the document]
davidmanzanares: There are at least 2 reasons (this was simplified step by step).

One reason is dead code like `_setRendererZoomCenter` which did things, but those things were never used.

The other reason is that before there was a lot of effort into trying to minimize calls to `requestData`. However, I think that wasn't good because it added a lot of code and the `requestData` function should be optimized (and it is) anyway because of a simple thing like panning will modify the view matrix at every frame during the panning

src/Viz.js
davidmanzanares: Yes! :+1:

src/renderer/Renderer.js
davidmanzanares: not being used, now it only uses the matrix

test/integration/render/scenarios/belongs/nin/scenario.js
davidmanzanares: hahaha, stands for `wadus`, I forgot to remove it

src/Layer.js
VictorVelarde: Even with just one proper use from _update, I would extract it to reduce the number of lines, but it's up to you

test/integration/render/scenarios/layer-update/uninterruptible-by-blendto/scenario.js
```diff
@@ -0,0 +1,42 @@
...
+    container: 'map',
+    background: 'black'
+});
+
+const source1 = new carto.source.GeoJSON(sources['points']);
+const layer = new carto.Layer('layer', source1, new carto.Viz());
+
+layer.addTo(map);
+layer.on('loaded', async () => {
+    // TODO
```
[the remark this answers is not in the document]
davidmanzanares: ups, I forgot to remove it

davidmanzanares:
hahaha, yes, the old code wasn't removed when applying your suggestion, but I refactored it a little bit afterwards

src/Viz.js
```diff
@@ -366,67 +366,23 @@ export default class Viz {
...
-            	if (toReplace === this[property]){
-                	this[property] = replacer;
-                    replacer.parent = this;
-                    replacer.notify = toReplace.notify;
-                }
-            });
-            
-            if (!properties.includes(toReplace)){
-            	throw new CartoRuntimeError('No child found');
+            const properties = ['color', 'width', 'strokeColor', 'strokeWidth', 'filter', 'symbol', 'symbolPlacement', 'transform'];
```
VictorVelarde: This could be a shared collection of properties, in its own module. See `featureVizProperties.js` in PR #1085

PR #1075, opened 2018-10-18
Set bundle size warning over 425KB

VictorVelarde:
### Related with: https://github.com/CartoDB/carto-vl/pull/1066

After that PR, the bundle size for the min has grown, so we need to adjust our warning size

### DIFF
--- a/webpack/webpack.min.config.js
+++ b/webpack/webpack.min.config.js
@@ -52,7 +52,7 @@ module.exports = {
         new webpack.BannerPlugin(banner)
     ],
     performance: {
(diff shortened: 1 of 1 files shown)


rochoa:
Why we didn't consider the option 3 in https://github.com/CartoDB/carto-vl/issues/1059?

VictorVelarde:
@rochoa After talking to @davidmanzanares about this, we just decided to go for option 2 being pragmatic, to keep the current code.

rochoa:
OK, fair enough.

In the future, we will work on improving the size of the bundle.

PR #1076, opened 2018-10-18
Update deps (major versions)

davidmanzanares:
- `yarn upgrade --latest` with the exception of "karma-webpack" (which breaks the build of the bundle for the tests)
- Fix an order dependency problem that was detected after updating the testing framework
- Update the linter to  use `2018` and run `yarn lint --fix`


VictorVelarde:
@davidmanzanares Can you try after these commits? (karma-webpack updated)

VictorVelarde:
The error with "karma-webpack": "3.0.5" was:
```
ERROR in chunk main [entry]
carto-vl.js
Conflict: Multiple chunks emit assets to the same filename carto-vl.js (chunks index.test and main)
```

`main` is here the default entry name for our build, when using `entry: './src/index.js'` (equivalent to `entry: { main: './src/index.js'}`). Until now, we were using the default entry point name and an static output filename for it.

And this situation looks like a problem with this webpack configuration (`webpack.config.js`) when it's used by the new karma-webpack for preparing the unit tests (`karma.conf.js`).

> We use the karma configuration recommended in https://github.com/webpack-contrib/karma-webpack, where a single bundle is created for all the tests. In the 3.0.5 version of that plugin it seems that they have fixed a bug (https://github.com/webpack-contrib/karma-webpack/releases/tag/v3.0.5) when dealing with multiple outputs. As a result, it seems that now the plugin tries to use the 'carto-vl.js' filename also for the index.text chunk, thus generating in our case that error in webpack (https://github.com/webpack/webpack/blob/a8687890f72b48d5135315f86bca788407d30b68/lib/Compilation.js#L2371).

If we use an explicit name and the [dynamic option](https://webpack.js.org/configuration/output/#output-filename), with `filename: [name].js`, that gives each bundle a different name and avoids the conflict.

davidmanzanares:
Thanks for the explanation!

PR #1082, opened 2018-10-19
Run headless render tests also in macOS

rochoa:
AFAIK, we don't run the tests in CI for macOS. Most of the time we run it locally, so this should be OK. See https://bugs.chromium.org/p/chromium/issues/detail?id=617551#c21.


rochoa:
Good catch. I can take care of that.

Issue #1078, opened 2018-10-18
Tests for devicePixelRatio


rochoa:
This bit us in the past a few times, we are still lacking some tests for it.

PR #1083, opened 2018-10-19
Add tests for devicePixelRatio

rochoa:
This test should catch the basic case when we mess up with the device pixel ratio. This should avoid regressions as in https://github.com/CartoDB/carto-vl/pull/1042.

Closes https://github.com/CartoDB/carto-vl/issues/1078.


rochoa:
I introduced the same regression we fixed in #1042 and the test failed. I'm gonna revert the change now to go green again.

Issue #954, opened 2018-09-14
The expression viewportFeatures should return features as in interactivity


Jesus89:
The name of this expression is confusing because you expect to obtain an array of features, like this one from the interactivity: https://carto.com/developers/carto-vl/reference/#feature.

However, the usage of this expression indicates that you get a list of metadata associated with a feature instead.

```js
viewportFeatures($value, $category) -> array of {value, category}
```

So IMO it should be renamed to something like `viewportMetadata` or `viewportFeaturesData` to clarify what this expression does.

elenatorro:
IMO, `viewportFeaturesData` or even `viewportFeaturesMetadata` sounds great. Should keep having a `viewportFeatures` expression that does what you expect, obtaining an array of features?

davidmanzanares:
We could return the same "smart" features we are returning on Interactivity, I think the implementation could re-use most parts of the Interactivity one.

VictorVelarde:
We have just merged the new approach to master 🚀

PR #1085, opened 2018-10-23
Use interactivity-features in viewportFeatures expression

VictorVelarde:
### Related to https://github.com/CartoDB/carto-vl/issues/954

This PR ensures that `viewportFeatures` expression returns the same type as `Interactivity` object.
Those features are defined in _interactivity/feature.js_ (**Feature** class) and they have:
- `id`
- `FeatureVizProperty` properties: `color`, `width`, `colorStroke`, `widthStroke`, `symbol` and `symbolPlacement` (+ `reset()` method). + missing `filter` and `transform`
- access to `variables`


src/renderer/viz/utils/featureVizProperties.js
```diff
@@ -0,0 +1,2 @@
+const FEATURE_VIZ_PROPERTIES = ['color', 'width', 'strokeColor', 'strokeWidth', 'symbol', 'symbolPlacement'];
```
davidmanzanares: Some properties are missing here (at least `filter` and `transform`), I don't know if adding them will have unexpected consequences though

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Eliminate the parameter to specify a property in `globalCount()` and `viewportCount()`
+- Change type from `viewportFeatures` expression, adding vizProperties to features as in `Interactivity`
```
davidmanzanares: I would put it in `Added` because as far as I know, this is not a breaking change, this just adds new functionality

src/interactivity/feature.js
```diff
@@ -72,34 +73,52 @@ import FeatureVizProperty from './featureVizProperty';
...
+        const variables = {};
+        Object.keys(vizVariables).map(varName => {
+            const name = `__cartovl_variable_${varName}`;
+            variables[varName] = this._buildFeatureVizProperty(name);
+        });
         this.variables = variables;
     }
 
+    get rawFeature () {
+        return this._rawFeature;
```
davidmanzanares: I don't know if we want to expose this. If the user doesn't use a property anywhere, then the property won't be here anyway.

If the user uses the property somewhere but not on the `viewportFeatures` call then the function is no longer pure, it is returning values for properties that are unknown for the function, i.e.: CVL magic is exposed in the open.

src/interactivity/feature.js
```diff
@@ -72,34 +73,52 @@ import FeatureVizProperty from './featureVizProperty';
...
+
 export default class Feature {
     constructor (rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty) {
-        const variables = {};
-        Object.keys(viz.variables).map(varName => {
-            variables[varName] = new FeatureVizProperty(`__cartovl_variable_${varName}`, rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty);
+        this.id = rawFeature[idProperty];
+
+        this._rawFeature = rawFeature;
+        this._featureVizParams = { rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty };
```
davidmanzanares: I think it is better to pass the parameters to `_buildFeatureVizProperty` directly, they are not used anymore after that

src/interactivity/feature.js
```diff
@@ -72,34 +73,52 @@ import FeatureVizProperty from './featureVizProperty';
...
-        this.symbol = new FeatureVizProperty('symbol', rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty);
-        this.symbolPlacement = new FeatureVizProperty('symbolPlacement', rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty);
+    _buildFeatureVizProperty (name) {
+        const { rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty } = this._featureVizParams;
+        return new FeatureVizProperty(name, rawFeature, viz, customizedFeatures, trackFeatureViz, idProperty);
+    }
+
+    _defineVariables (vizVariables) {
+        const variables = {};
+        Object.keys(vizVariables).map(varName => {
```
davidmanzanares: `forEach`?

src/renderer/viz/expressions/viewportFeatures.js
```diff
@@ -54,6 +56,7 @@ import CartoRuntimeError from '../../../errors/carto-runtime-error';
 export default class ViewportFeatures extends BaseExpression {
     constructor (...properties) {
         properties = properties.map(p => implicitCast(p));
+        // TODO validate here properties don't collide with interactivity.Feature predefined props
```
davidmanzanares: Having properties and viz properties under the same namespace may bite us in the future. But changing this would be a breaking change to add another namespace... I don't know what should we do.

cc: @rochoa

src/renderer/viz/expressions/viewportFeatures.js
```diff
@@ -64,7 +67,7 @@ export default class ViewportFeatures extends BaseExpression {
         this.type = 'featureList';
         this._isViewport = true;
         this._requiredProperties = properties;
-        this._FeatureProxy = null;
+        this._columns = null;
```
davidmanzanares: columns is SQL terminology, I don't think it should be named like that here

src/renderer/viz/expressions/viewportFeatures.js
```diff
@@ -84,34 +87,31 @@ export default class ViewportFeatures extends BaseExpression {
...
-        };
-        properties.forEach(prop => {
-            Object.defineProperty(cls.prototype, prop, {
-                get: function () {
-                    return this._feature[prop];
-                }
-            });
+    _addRequiredPropertiesTo (interactivityFeature) {
+        this._columns.forEach((column) => {
+            interactivityFeature[column] = interactivityFeature.rawFeature[column];
```
davidmanzanares: This refactor is not the first time it is proposed. However, there is a reason for `genViewportFeatureClass`. 

`genViewportFeatureClass` defines a class with getters which are able to retrieve the properties without storing the properties within the class instance.

The previous code, which is similar to yours, adds a lot of overhead in memory allocation and in GC since each feature is being created one property at a time.

See the PR that optimized this: https://github.com/CartoDB/carto-vl/pull/663

test/integration/user/test/viewportFeatures.test.js
```diff
@@ -45,7 +46,7 @@ function checkFeatures (list, expectedList) {
     }
 }
 
-describe('viewportFeatures', () => {
+fdescribe('viewportFeatures', () => {
```
davidmanzanares: `f`

test/integration/user/test/viewportFeatures.test.js
```diff
@@ -97,6 +98,27 @@ describe('viewportFeatures', () => {
...
+                { value: 10, category: 'a' },
+                { value: 1000, category: 'b' }
+            ];
+            const featureList = viz1.variables.list.eval();
+            checkFeatures(featureList, expected);
+
+            // 2. and featureVizProperties / variables are available
+            featureList.forEach(feature => {
+                FEATURE_VIZ_PROPERTIES.forEach(propertyName => {
+                    expect(feature.hasOwnProperty(propertyName)).toBeTruthy();
```
davidmanzanares: This checks that the viz properties exist, but it doesn't test they work. I think that a `render` test would be better: wait for `loaded`, read viewportFeatures variable, use a `blendTo` in one of the features, wait, mark the test as done

src/renderer/Renderer.js
```diff
@@ -187,13 +188,30 @@ export default class Renderer {
...
         });
     }
 
+    _accumViewportAggregations (viewportExpressions, rawFeature, renderLayer) {
+        const viewportExpressionsLength = viewportExpressions.length;
+
+        for (let j = 0; j < viewportExpressionsLength; j++) {
+            const currentViewportExp = viewportExpressions[j];
+            if (currentViewportExp instanceof ViewportFeatures) {
+                const interactivityFeature = new Feature(rawFeature,
```
davidmanzanares: I think the place for this is in `viewportFeatures`, although I guess that some `renderLayer` parameter will need to be passed to the expression (maybe in the reset function?).

On top of that, I think this code will perform poorly since `viewportFeatures` may return many features and the `Feature` constructor is not lightweight. I hope that the same trick that is being used in `viewportFeatures` to mitigate memory allocation overhead can be used for this too.

src/renderer/viz/utils/featureVizProperties.js
VictorVelarde: Good catch. I have added them both and the tests pass

CHANGELOG.md
VictorVelarde: 👍 Yeah, it is an addition, cause I'm finally keeping the "extended properties" for these features as they are, the ones required in the constructor (I had some doubts at the beginning...)

src/interactivity/feature.js
VictorVelarde: Yeah, I had some doubts on this. Would you prefer just a `feature._rawFeature` call from outside?. I thought this might be useful in some use cases, like preparing popups, grids... but I don't have a strong opinion

davidmanzanares: It is definitely useful, and if we expose it users will probably use it. However, I think that the problems outweigh the advantages.

src/interactivity/feature.js
VictorVelarde: Fine!

src/interactivity/feature.js
VictorVelarde: Better! I had not noticed that one

src/interactivity/feature.js
VictorVelarde: Then let's remove it!

src/renderer/viz/expressions/viewportFeatures.js
VictorVelarde: I thought about a breaking change here. 

IMO it's more clear if we move the vizProperties one level down (as the current `variables`), and keep the "feature properties" there. For now I added a validation to catch this and raise an error (e.g. when trying to use a 'color' column from a feature)

src/renderer/viz/expressions/viewportFeatures.js
VictorVelarde: Thx! I was keeping the original name from LIN 91, but I think I'm gonna rename it to "_propertyNames"

test/integration/user/test/viewportFeatures.test.js
VictorVelarde: Good catch

test/integration/user/test/viewportFeatures.test.js
VictorVelarde: Yeah, that's true. I'll keep this one and add another render test

src/renderer/Renderer.js
VictorVelarde: @davidmanzanares That was my first intention, as it feels more natural. 

I was using `accumViewportAgg (interactivityFeature, renderLayer)` and it looked well, but bringing `Feature` into `viewportFeatures` breaks the current build (and tests). It seems to be causing a circular dependency problem, regarding to `expressions`. You can reproduce it just by adding this to viewportFeatures code:
`import Feature from '../../../interactivity/feature';`
We can try to dig more into that or just keep it in `Renderer` for now.

On the other hand, I'll try to mitigate memory allocation by transforming object construction.

src/renderer/viz/expressions/viewportFeatures.js
VictorVelarde: Fair enough, thx!

src/interactivity/feature.js
VictorVelarde: I have changed my mind here after the last refactor, as I was using it more in other methods, and I found it convenient to keep it as a private variable

src/renderer/viz/expressions/viewportFeatures.js
VictorVelarde: I have applied the same dynamic creation of getters, but not with a proxy in the class, but within the Feature class itself, with an optional parameter. See ⬇️

VictorVelarde:
@davidmanzanares The last commit applies a dynamic creation of getters, exposing rawFeature properties and thus eliminating the overhead of new copied variables. It is not a proxy in the viewportFeatures class, but a change made to the Feature class itself, with an optional parameter

VictorVelarde:
The last commit adds a much lighter version of the Feature class (LigthweightFeature) with all the original properties deferred behind getters and built around the original rawFeature

VictorVelarde:
@rochoa after merging your fix to master, e2e is working again here!

@davidmanzanares I've used the same pattern than you in other render scenarios to fix mine (adding `await sleep` after `blendTo`)

test/integration/render/scenarios/viewport/viewportfeatures-viz-properties/scenario.js
```diff
@@ -0,0 +1,34 @@
...
+
+    // Using viz properties, viz variables (and also feature props)
+    const duration = feature.numeric;
+    const augmentedSize = feature.variables.augmentedSize.value;
+
+    await feature.color.blendTo('blue', duration);
+    await sleep(duration);
+
+    await feature.width.blendTo(augmentedSize, 10);
+    await sleep(10);
```
rochoa: What about using the `updated` event to decide when to mark it as loaded? We are already doing that for the e2e tests: https://github.com/CartoDB/carto-vl/blob/3eaf73bf8d930640a646a708fcc6d9c5ad0e76b6/test/acceptance/e2e/blendTo/scenario.js#L27

debug/viewportfeatures-perf.html
```diff
@@ -0,0 +1,90 @@
+ok
```
davidmanzanares: ops

debug/viewportfeatures-perf.html
```diff
@@ -0,0 +1,90 @@
...
+     @wadus: viewportAvg($trip_distance)
+     @wadus2: viewportFeatures($fare_amount)
+   `);
+    const layer = new carto.Layer('layer', source, viz);
+
+    layer.addTo(map, 'watername_ocean');
+    layer.on('loaded', hideLoader);
+    layer.on('updated', () => {
+      if (!window.feature) {
+        window.feature = viz.variables.wadus2.value[0];
```
davidmanzanares: calling `window.feature.width.blendTo(100);` in the console throws. The reason is that `renderLayer.idProperty` is null. PR to fix it: https://github.com/CartoDB/carto-vl/pull/1098

davidmanzanares: Apart from that I experienced a change in width in all the features the first time (as we discussed earlier), I'm going to research that

debug/viewportfeatures-perf.html
VictorVelarde: 😄

debug/viewportfeatures-perf.html
VictorVelarde: 👍

VictorVelarde:
@davidmanzanares we can merge this in, if you agree!

davidmanzanares:
Further manual testing revealed that calling `.width.reset()` ends up with
`carto-error.js:23 Uncaught (in promise) CartoRuntimeError: Another `viz change` finished before this one. Commit 0.7 overridden by commit 0.9`

davidmanzanares:
I also found that the API is not very consistent with the Interactivity one.

The interactivity one provides top-level `blendTo` and `reset` methods to allow setting all styling properties at the same time, this functionality is not added by this PR.

We should discuss this, and the need to add another namespace with the properties.

davidmanzanares:
Regarding the bug, with that info, it is probably related to the Layer refactor, so I will try to debug it on Monday.

Regarding the top-level methods I rechecked and only a `reset` exists, see https://github.com/CartoDB/carto-vl/blob/master/src/interactivity/feature.js#L96

VictorVelarde:
Yes, I've checked it out!. 

First I discarded my `blendUtils` refactor and later on I just tried with the current **master**, where http://127.0.0.1:8080/examples/guides/interactivity/step-4.html is broken. This example is unrelated to this PR, but it is also using `blendTo`, thus getting the same errors as you detected here.

Regarding to `reset()` at root-level, I hope that the existing implementation in `LightweightFeature` will work after dealing with the `Layer` issue. 

I think we should also add the root-level `blendTo`, cause I find it quite interesting and it would give some "simmetry" to the class. The call could be something like:
```js
feature.blendTo(
  {
    color: "yellow",
    width: 20,
    strokeColor: "red",
    strokeWidth: 2
  },
  100
);
```

We'll talk on Monday, have a nice weekend!

VictorVelarde:
After the fix in #1103, the current behaviour looks fine. 
IMO just one question is left: should we add the `blendTo` at root-level for all features? 

See proposal 🔝

davidmanzanares:
I think that the proposed `blendTo` could be not trivial to implement, so I would try to add that later.

Reviewing the API, particularly, if the feature properties should go under a new namespace is in my opinion much more important (and the future addition of `blendTo` should be taken into account here too).

Another thing I saw is that the features don't include their geometry.

VictorVelarde:
Last changes: 
* `feature.blendTo` (convenience method to allow several combined blendTo)
* `feature.properties` namespace (**breaking change**, but much needed)

rochoa:
VAMOS! 💪

PR #1086, opened 2018-10-23
Improve e2e environment setup

rochoa:
## Use loopback IP instead of hostname
This uses a subdomain-less configuration in the Maps API to avoid resolving the subdomain. See https://github.com/CartoDB/carto-vl/pull/1091 for more details.

## Pass extra arguments from the parent script
Allowing, for instance, to grep specific tests.

## Future improvements
 - Being able to use a different Maps API branch than `master`.
 - Bind to an unused port to avoid conflicts when starting the Maps API application.


VictorVelarde:
@rochoa not sure why, but even rebuilding the docker image, I get this error  with 'undefined-geometry-type' test

image

rochoa:
Part of this PR is superseded by #1091. Specifically, we will no longer validate the hostname as it will use a lookback IP instead of a hostname.

rochoa:
This is now the continuation of https://github.com/CartoDB/carto-vl/pull/1091. Instead of validating the hostname, it uses an IP and a different configuration mechanism for the Maps API.

Issue #878, opened 2018-08-22
Remove deprecated checkLooseType


davidmanzanares:
`checkLooseType` is complex, error-prone (there have been more bugs with it than I can count) and adds additional complexity.

We are already not adding more checks like that, but the codebase still contains a huge amount of them.

Removing them will increase consistency, reduce the code base and reduce the likelihood of false positive type errors.

cc: @rochoa

davidmanzanares:
`opacity()` with variables is broken due to this

elenatorro:
Good catch 🎉

davidmanzanares:
I reviewed every expression constructor and I found several things:
- Some expressions check for **string/number literals**. Sometimes it's due to clear implementation limitations, but other times it would be easy to remove that limitation.
- **reverse** checks the type at constructor time and it is broken with palettes under variables
- **animation** has a `isA` check at constructor time. The automatic `linear` don't work with variables. This was also discovered by @jgoizueta 
- **Binary** has a `getSignatureLoose` which I think was used before for a check, right now I think it is safe, but better to remove it.
- **blend** has one "checkLooseType" check, I think it is correct, but better to remove it
- **Asc/Desc** have problematic checks and are broken with variables
- **SVG** assumes a string literal as the first parameter, but it is not checked. So it is broken under variables (in this case this is probably an understandable limitation), but the error on non-string parameters is probably a nightmare
- **Time** has literal-like limitations and it has undocumented functionality
- **Top** has a  checkInstance that breaks it with variables

I'm going to priotize them and I'll try to fix the most important ones. However, I don't think I'll remove every "no-variable input" limitation since I don't think it may make sense every time.

davidmanzanares:
Waiting for time aggregations to close this

PR #1090, opened 2018-10-24
Improve expression constructors

davidmanzanares:
Address https://github.com/CartoDB/carto-vl/issues/878

Fixing:
- Top has a checkInstance that breaks it with variables
- reverse checks the type at constructor time and it is broken with palettes under variables
- Time has undocumented functionality and no checks
-  Asc/Desc have problematic checks and are broken with variables
- Binary has a getSignatureLoose which I think was used before for a check, right now I think it is safe, but better to remove it.
-  blend has one "checkLooseType" check, I think it is correct, but better to remove it
- globalPercentile improvements (some limitations were discovered with @makella )


src/utils/util.js
```diff
@@ -44,7 +44,11 @@ export function castDate (date) {
         date.setUTCMilliseconds(msEpoch);
         return date;
     }
-    return new Date(date);
+    if (isString(date)) {
+        return new Date(date);
+    } else {
+        throw new Error('invalid date type');
```
VictorVelarde: Here we could raise a new CartoValidationError

PR #1091, opened 2018-10-24
E2e subdomainless docker

rochoa:
This uses subdomainless configuration in the Maps API to avoid requiring a new hostname configured in `/etc/hosts`.

It also `checkout`s and `pull`s latest changes from the master branch.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -102,8 +102,6 @@ This end to end tests cover the entire library by performing tests against real 
 
 To install Docker, follow the instructions on https://docs.docker.com/install/
 
-You'll also need to add `127.0.0.1 localhost.localhost.lan` your `/etc/hosts/` file.
-
 ```bash
 # Running the tests
 $ yarn test:e2e
--- a/test/acceptance/docker/deploy.sh
+++ b/test/acceptance/docker/deploy.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 git clone https://github.com/CartoDB/Windshaft-cartodb.git ;
+git checkout master
+git pull origin master
 
 cp /srv/config/environments/test.js Windshaft-cartodb/config/environments/development.js
 cd ./Windshaft-cartodb/
--- a/test/acceptance/docker/deploy.sh
+++ b/test/acceptance/docker/deploy.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
 
 git clone https://github.com/CartoDB/Windshaft-cartodb.git ;
-git checkout master
-git pull origin master
 
 cp /srv/config/environments/test.js Windshaft-cartodb/config/environments/development.js
 cd ./Windshaft-cartodb/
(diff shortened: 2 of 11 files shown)


DEVELOPERS.md
```diff
@@ -102,8 +102,6 @@ This end to end tests cover the entire library by performing tests against real
 
 To install Docker, follow the instructions on https://docs.docker.com/install/
 
-You'll also need to add `127.0.0.1 localhost.localhost.lan` your `/etc/hosts/` file.
```
davidmanzanares: Vamos!

test/acceptance/docker/deploy.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 git clone https://github.com/CartoDB/Windshaft-cartodb.git ;
+git checkout master
```
davidmanzanares: Are these really needed? A clean `clone` just was performed

rochoa: I'm not sure. 🤷‍♂️ 

What happens when you already have a local copy?

DEVELOPERS.md
VictorVelarde: 👍 nice!

test/acceptance/docker/deploy.sh
VictorVelarde: (from docs) It seems that clones checks out currently active branch, by default
https://git-scm.com/docs/git-clone

davidmanzanares: I see the intention now, but, the clone will create a new directory and the checkout and pull will be performed in the parent directory, right?

Also, do we want to enforce the checkout+pull? Isn't it useful to be able to work with a local Windshaft?

rochoa: Actually, I just tested and it doesn't work as I was expecting. :\

If I check out an older commit than master:HEAD, expecting that the deploy script would check out master:HEAD again, but it doesn't.

rochoa: I think it's not a problem for CI. I'm gonna remove this and push a new docker image.

I was trying to fix the situation with outdated Windshaft-cartodb repositories in local/dev environments.

rochoa:
About the `git clone`. Do you know if that gets executed every time the image starts or it's executed and materialized when you build the image?

davidmanzanares:
When the image starts, since we want to test Windshaft `master` every time and not a stale version. Since the `deploy.sh` is the entry point it will be called by default when running the image.

The ´clone´ would have been on the Dockerfile if it was on the build phase.

rochoa:
Oh, stupid me! I know why it wasn't having any effect. I wasn't changing the directory where I was executing the `git checkout && git pull`. If only you would get errors. 💔

davidmanzanares:
> the clone will create a new directory and the checkout and pull will be performed in the parent directory, right?

:joy::joy:

davidmanzanares:
You can also open a bash session in Docker to run the commands interactively and see all the output

VictorVelarde:
💪 vamos!

PR #1095, opened 2018-10-25
Prevent constant triggering of the layer `updated` event

rochoa:
We introduced this regression in https://github.com/CartoDB/carto-vl/commit/3ab6087171fb079be1e12c335dddb2b51191d6c0.

This takes advantage of the fact that we are already repainting when a dataframe is added.

As an enhancement, we only trigger that event when the dataframe gets activated. This also means we don't need anymore the onDataLoaded callback all across the place.

Profiling before the changes
before

Profiling after the changes
after

### DIFF
--- a/src/renderer/Dataframe.js
+++ b/src/renderer/Dataframe.js
@@ -109,6 +109,21 @@ export default class Dataframe extends DummyDataframe {
         }
     }
 
+    onActive (callback) {
+        this._onActive = callback;
+    }
+
+    get active () {
+        return this._active;
+    }
+
+    set active (value) {
+        if (this._active !== value) {
+            this._active = value;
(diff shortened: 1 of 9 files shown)


rochoa:
The failing e2e test is the same that is failing in the `master` branch.

src/renderer/Dataframe.js
```diff
@@ -109,6 +109,24 @@ export default class Dataframe extends DummyDataframe {
         }
     }
 
+    onActive (callback) {
+        this._onActive = callback;
+        if (this.active) {
```
rochoa: I don't know if this makes sense. I added it because a test was failing, but the test itself is testing the implementation and not the API.

PR #1100, opened 2018-10-26
Fix shader textureIDs with shader cache

davidmanzanares:
### The bug

You can reproduce the bug by opening http://127.0.0.1:8080/debug/viewportfeatures-perf.html and putting in the console `window.feature.width.blendTo(100);`. You'll see that every point changes in width, even though we are targeting a single feature.

Of course, I thought this was related to the per-feature blending, but it wasn't that at all.

After a lot of debugging, I discovered the root cause: **the shader cache** and **textureIDs**.

When the per-feature `blendTo` is called that has to regenerate a new shader for the `width`. This new shader is completely different from the original one, and therefore, it triggers a clean shader recompilation. The key thing is that the end result (with all the features changed in width) is the correct result. The wrong result is the initial one.

Why is the initial result wrong?

Well, this example is funny, if you look at the style of `width` and the style of `strokeWidth` you'll see that both expression trees have the same patterns, numbers changed, but the same expressions are used in the same order (same tree).

This similarity is not ignored by CVL shader caching mechanism. Firstly, a clean compilation is performed to generate a valid shader for `width`. After that, a shader compilation is requested for `strokeWidth`, but that correctly hits the shader cache because the code is identical, avoiding the recompilation.

However, even though the shaders for both expression trees are identical, its JS metadata (particularly `textureIDs`) is not. 

The `textureIDs` property is a simple object that matches property names with texture IDs. This is used to bind Dataframe property textures into the shader uniforms.

The previous code was assuming one `textureIDs` table per Viz object, which makes the shader cache between multiple Vizs (layers). However, it assumed that only one table is needed per Viz, which is WRONG since two expression trees may have identical code but refer to different properties within the same Viz.

Bottom line:

Visualizations displayed wrong attributes when 2 styling expressions within the same Viz shared the same expression pattern but different properties.

### The solution

I tried several solutions. The first ones based on keeping the existing `Map` and improving the key. This is much more difficult than expected because:
- you need to keep discriminating based on Viz
- you need to discriminate based on expressions, but there are shaders that combine multiple root-level expressions
- blendTo details that I haven't fully understand

The final solution is based on removing the `Map` and creating a `MetaShader` object that is **not cached** and that stores the `textureIds` object (uncached, copied for each metashader) and the cached Shader.

TODO:
- [x] This is still broken!!! Fixed for different styleProperty, but broken by different Vizs!
- [x] Add regression test


src/renderer/Renderer.js
```diff
@@ -216,8 +216,9 @@ export default class Renderer {
         gl.depthMask(false);
         gl.bindFramebuffer(gl.FRAMEBUFFER, this.auxFB);
 
-        const styleDataframe = (dataframe, dataframeTexture, shader, vizExpr) => {
-            const textureId = shader.textureIds.get(viz);
+        const styleDataframe = (dataframe, dataframeTexture, metashader, vizExpr) => {
```
VictorVelarde: This function only uses `this.gl`, so it can be extracted with a refactor to something like this:

```js
_styleDataframe (dataframe, dataframeTexture, metashader, vizExpr) => {
            const gl = this.gl;

            const shader = metashader.shader;
            const textureId = metashader.textureIds;

            gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, dataframeTexture, 0);
            gl.viewport(0, 0, RTT_WIDTH, dataframe.height);
            gl.clear(gl.COLOR_BUFFER_BIT);

            gl.useProgram(shader.program);
            // Enforce that property texture TextureUnit don't clash with auxiliar ones
            drawMetadata.freeTexUnit = Object.keys(textureId).length;
            vizExpr._setTimestamp((Date.now() - INITIAL_TIMESTAMP) / 1000.0);
            vizExpr._preDraw(shader.program, drawMetadata, gl);

            Object.keys(textureId).forEach((name, i) => {
                gl.activeTexture(gl.TEXTURE0 + i);
                gl.bindTexture(gl.TEXTURE_2D, dataframe.getPropertyTexture(name));
                gl.uniform1i(textureId[name], i);
            });

            gl.enableVertexAttribArray(shader.vertexAttribute);
            gl.bindBuffer(gl.ARRAY_BUFFER, this.bigTriangleVBO);
            gl.vertexAttribPointer(shader.vertexAttribute, 2, gl.FLOAT, false, 0, 0);

            gl.drawArrays(gl.TRIANGLES, 0, 3);
            gl.disableVertexAttribArray(shader.vertexAttribute);
        };
```

src/renderer/Renderer.js
```diff
@@ -243,11 +244,15 @@ export default class Renderer {
             gl.disableVertexAttribArray(shader.vertexAttribute);
         };
 
-        dataframes.map(dataframe => styleDataframe(dataframe, dataframe.texColor, viz.colorShader, viz.color));
-        dataframes.map(dataframe => styleDataframe(dataframe, dataframe.texWidth, viz.widthShader, viz.width));
-        dataframes.map(dataframe => styleDataframe(dataframe, dataframe.texStrokeColor, viz.strokeColorShader, viz.strokeColor));
-        dataframes.map(dataframe => styleDataframe(dataframe, dataframe.texStrokeWidth, viz.strokeWidthShader, viz.strokeWidth));
-        dataframes.map(dataframe => styleDataframe(dataframe, dataframe.texFilter, viz.filterShader, viz.filter));
+        dataframes.map(dataframe => styleDataframe(dataframe, dataframe.texColor, viz.colorMetaShader, viz.color));
+        if (dataframes[0].type !== 'polygon') {
```
VictorVelarde: It's nice to avoid it if not needed

src/renderer/Renderer.js

davidmanzanares: I was already doing the refactor when I remembered why this was left there, increasing considerably the already long render method.

The reason is that the code relies a lot in the current state of WebGL, therefore the function is completely tied to that place. It depends on a lot of state set to WebGL before its call, and it probably lefts the WebGL state in a particular state.

Therefore, even though extracting it is trivial, there are a lot of dependencies (WebGL side effects) that are needed.

PR #1103, opened 2018-10-29
Fix feature `reset()` by ignoring rejected changes

davidmanzanares:
Fix the issue described https://github.com/CartoDB/carto-vl/pull/1085#issuecomment-433468670

Uncaught exceptions were being thrown when using the feature's `reset` method.

This was due to our recent changes in `Layer`. Basically, we informed of "overridden" updates by rejecting the update promise. This doesn't make sense from the `blendTo` point of view since that changes the viz synchronously.

Therefore, this PR ignores those rejections when the call to update has been the result of a `blendTo`.

I don't know if this will be easier to test within #1085 , maybe we could merge this to that instead of `master`.

@VictorVelarde What do you think? Could you help with the regression test?

TODO:
- [x] Add regression test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix `viewportPercentile` (styling by it was broken)
 - Fix `viewport*` functions to take clustering into account
 - Fix wrong `expressionName` due to mangled `this.constructor.name` in the minified version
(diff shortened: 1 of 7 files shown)


VictorVelarde:
@davidmanzanares I'm gonna have a look at it

davidmanzanares:
Thanks @VictorVelarde!

davidmanzanares:
Super good work finding that function!

PR #1109, opened 2018-10-30
Fix Legend data (geomType) with no data (still metadata)

davidmanzanares:
As reported by @makella

### DIFF
--- a/src/Viz.js
+++ b/src/Viz.js
@@ -241,7 +241,7 @@ export default class Viz {
     }
 
     setDefaultsIfRequired (geomType) {
-        this._geomType = geomType;
+        this._geomType = this._geomType || geomType;
         if (this._appliedDefaults) {
             return;
         }


VictorVelarde:
@makella Do you have an example to get back to this issue?

makella:
hmmmm unfortunately, no.

makella:
was likely opened after a discussion w/ David and I honestly can't remember! (sorry)

VictorVelarde:
Ok, thx I'll try to find the related issue...

davidmanzanares:
Hey! Just saw this little thing left open. I don't know if this still happens, but if it does it could be reproduced by a map from an SQL/Dataset source with a view that returns only empty MVTs and a legend that uses the geometry type (I don't remember how to do this exactly, but @elenatorro probably knows :D)

Issue #1045, opened 2018-10-09
Filter breaks when using global & viewport functions


VictorVelarde:
When using **populated_places** to build the guide on Aggregations (#1023), I found this:

This viz works fine:
```js
 const citiesViz = new carto.Viz(`
            color: grey
            width: 15
            @g_max: globalMax($pop_max)
            @g_avg: globalAvg($pop_max)
            @g_p95: globalPercentile($pop_max, 95)
            filter: ($pop_max > @g_p95)
`);
```

But adding another variable with a viewport aggr. function seems to break it. This doesn't work:
```js
        const citiesViz = new carto.Viz(`
            color: grey
            width: 15
            @g_max: globalMax($pop_max)
            @g_avg: globalAvg($pop_max)
            @g_p95: globalPercentile($pop_max, 95)
            filter: ($pop_max > @g_p95)
            @v_sum: viewportSum($pop_max)
`);
```

The error is `CartoRuntimeError: Unimplemented` at `GlobalPercentile.eval`
![image](https://user-images.githubusercontent.com/458196/46685244-4c294800-cbf5-11e8-805a-d6d8638fdb89.png)

VictorVelarde:
Fix merged to master 🚀

PR #1113, opened 2018-11-02
Fix animations

davidmanzanares:
`master` has regular animations broken (WWI ships for example). It was difficult to fix that while keeping `/debug/timerange.html` working.

To be honest, I don't know why exactly was this broken. However, `Variant` seems suspicious and that was already inconsistent with how we resolve a similar problem for `ramp` and for `reverse`. Therefore, I remove it.

I think this works, but at the same time, I really think this code should be refactored at some point.


src/renderer/viz/expressions/Animation.js
```diff
@@ -68,21 +67,24 @@ import AnimationRange from './AnimationRange';
...
-    _choose () {
-        if (this.input.type || !(this.input instanceof BaseExpression)) {
-            if (this.input.type === 'timerange') {
-                return new AnimationRange(this.input, this.duration, this.fade);
-            } else {
-                return new AnimationGeneral(this.input, this.duration, this.fade);
-            }
+    _resolveAliases (aliases) {
+        this._getChildren().map(child => child._resolveAliases(aliases));
+        if (this.input.type === 'timerange') {
```
rochoa: I still don't like this: it stinks. 😄 

Let's move on and review it later.

debug/timerange.html
```diff
@@ -110,7 +110,7 @@
         console.log(viz);
         console.log("Min Month", viz.variables.minMonth.value);
         console.log("Max Month", viz.variables.maxMonth.value);
-        console.log('Months:',viz.variables.list.value.map(f => viz.variables.month.eval(f).text));
+        console.log('Months:',viz.variables.list.value.map(f => viz.variables.month.eval(f.properties).text));
```
VictorVelarde: Yeah, this looks like the new 'properties' namespace breaking change

src/renderer/viz/expressions/base.js
```diff
@@ -66,10 +66,7 @@ export default class Base {
      *
      */
     eval (feature) {
-        if (this.isFeatureDependent()) {
-            throw new CartoRuntimeError('Unimplemented');
-        }
-        return this.value;
+        throw new CartoRuntimeError('Unimplemented');
```
VictorVelarde: This was a fix for https://github.com/CartoDB/carto-vl/issues/1045

davidmanzanares: See all the code:
```
eval (feature) {
        if (this.isFeatureDependent()) {
            throw new CartoRuntimeError('Unimplemented');
        }
        return this.value;
    }
    get value () {
        return this.eval();
    }
```

If you don't override any of those methods you'll get a stack overflow since they both try to rely on the other one.

If this introduces a regression for #1045 then we should fix it in other way.

Could you check if this will actually re open #1045 ?

VictorVelarde: I'll check it, but IMO this takes priority, so go ahead with this and I'll reopen that one if needed

davidmanzanares: :+1:

VictorVelarde: 👌 I have checked it and the original error at 1045 doesn't happen anymore, even adding more viewport functions, such as viewportFeatures

davidmanzanares: Great!

Issue #1114, opened 2018-11-02
Categorical properties with MVT sources are broken


davidmanzanares:
As reported off-channel by @padawannn the problem seems caused by
```
_clone () {
        return new MVT(this._templateURL, JSON.parse(JSON.stringify(this._metadata)), this._options);
    }
```
```Converting circular structure to JSON```

PR #1115, opened 2018-11-02
Fix MVT with categories

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/1114


src/sources/MVT.js
```diff
@@ -42,7 +42,6 @@ export default class MVT extends Base {
             metadata = new MVTMetadata(metadata);
         }
         this._metadata = metadata;
-        this._metadata.setCodecs();
```
VictorVelarde: What problem was causing this?

davidmanzanares: The MVT category codec object has a reference to the metadata, and the metadata has a reference to the codec.

On the `_clone` method, JSON.stringify throw due to a circular structure.

PR #1116, opened 2018-11-02
Revert "Fix polygon decoding corner case with NaN normals"

### LINKED ISSUE #1107: Clipping/resolution issue with polygon stroke
To test the issue open the jsfiddle and play with the zoom.

https://jsfiddle.net/qk13vn2d/

rochoa:
This reverts commit 07a3052d0587beff5345a4e7dfc2bd200dad60ba which seems to be the responsible for https://github.com/CartoDB/carto-vl/issues/1107.

It's not clear what that commit was trying to prevent but definitively causes this regression.

Let's discuss what we should try to cover apart from reverting the regression.

Fixes #1107.


davidmanzanares:
I can confirm that this breaks the interactivity on these corner cases, I'll see how to keep both things fixed.

rochoa:
#1121 supersedes this one.

PR #1118, opened 2018-11-02
Upgrade MGL version from v0.48.0-carto1 to v0.50.0 in several examples

VictorVelarde:
Some examples were still using our previous fork, instead of the now supported version (official MGL 0.50.0)


VictorVelarde:
I have seen some 'auto-formatting' in some html paragraphs... let me fix that

PR #1120, opened 2018-11-02
Fix example

davidmanzanares:


debug/animation.html
```diff
@@ -4,7 +4,7 @@
   <title>Animation demo | CARTO VL</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta charset="UTF-8">
-  <script src="../../dist/carto-vl.js"></script>
+  <script src="../dist/carto-vl.js"></script>
```
VictorVelarde: This works fine!

davidmanzanares:
> Url change works fine. When reviewing animation examples, I fixed a small error, when trying to convert value toISOString from '0' (first animation.getProgressValue())

Yes, I was commenting this morning off channel about that, I don't know if it is a deprecated example or a bug

davidmanzanares:
Ah, I see the problem now, the problem is that `updated` is being fired before loading.

davidmanzanares:
Should be fixed now

src/Layer.js
```diff
@@ -417,7 +417,9 @@ export default class Layer {
             this._needRefresh();
         }
 
-        this._fire('updated');
+        if (this._state !== states.INIT) {
```
VictorVelarde: Good catch!

VictorVelarde: I guess we should remove the check

VictorVelarde: Give 1 min

PR #1128, opened 2018-11-06
Fix anim sync

davidmanzanares:
Fix https://github.com/CartoDB/carto-vl/issues/599


src/renderer/viz/expressions/AnimationGeneral.js
```diff
@@ -133,6 +138,7 @@ export default class AnimationGeneral extends BaseExpression {
         }
 
         this.progress.expr = (this.progress.expr + speed * deltaTime) % 1;
+        console.log(timestamp, this.progress.expr);
```
rochoa:
```suggestion
```

Let's remove that console.log.

davidmanzanares: Ups :joy:

PR #1129, opened 2018-11-06
Change signature in `viewportHistogram`

rochoa:
From `(x, weight = 1, size = 1000)` to `(x, size = 20, weight = 1)`.

### DIFF
--- a/examples/guides/add-widgets/step-2.html
+++ b/examples/guides/add-widgets/step-2.html
@@ -52,7 +52,7 @@
 
         const source = new carto.source.Dataset("railroad_accidents");
         const viz = new carto.Viz(`
-            @histogram: viewportHistogram($total_damage, 1, 6)
+            @histogram: viewportHistogram($total_damage, 6)
 `);
         const layer = new carto.Layer('layer', source, viz);
 
--- a/src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
+++ b/src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
@@ -13,15 +13,15 @@ import { CLUSTER_FEATURE_COUNT } from '../../../../schema';
  * Histograms are useful to get insights and create widgets outside the scope of CARTO VL, see the following example for more info.
  *
  * @param {Number} input - expression to base the histogram
- * @param {Number} weight - Weight each occurrence differently based on this weight, defaults to `1`, which will generate a simple, non-weighted count.
- * @param {Number} size - Optional (defaults to 1000). Number of bars to use if `x` is a numeric expression
+ * @param {Number} size - Optional (defaults to 20). Number of bars to use if `x` is a numeric expression
+ * @param {Number} weight - Optional. Weight each occurrence differently based on this weight, defaults to `1`, which will generate a simple, non-weighted count.
  * @return {Histogram} Histogram
  *
  * @example <caption>Create and use an histogram. (String)</caption>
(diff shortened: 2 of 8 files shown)


examples/guides/add-widgets/step-2.html
```diff
@@ -52,7 +52,7 @@ <h1>Damage distribution (logarithmic scale)</h1>
 
         const source = new carto.source.Dataset("railroad_accidents");
         const viz = new carto.Viz(`
-            @histogram: viewportHistogram($total_damage, 1, 6)
+            @histogram: viewportHistogram($total_damage, 6)
```
VictorVelarde: 👍 This looks more intuitive

src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
```diff
@@ -13,15 +13,15 @@ import { CLUSTER_FEATURE_COUNT } from '../../../../schema';
  * Histograms are useful to get insights and create widgets outside the scope of CARTO VL, see the following example for more info.
  *
  * @param {Number} input - expression to base the histogram
- * @param {Number} weight - Weight each occurrence differently based on this weight, defaults to `1`, which will generate a simple, non-weighted count.
- * @param {Number} size - Optional (defaults to 1000). Number of bars to use if `x` is a numeric expression
+ * @param {Number} size - Optional (defaults to 20). Number of bars to use if `x` is a numeric expression
```
VictorVelarde: 👌

src/renderer/viz/expressions/classification/ViewportStandardDev.js
```diff
@@ -58,7 +58,7 @@ export default class ViewportStandardDev extends Classifier {
             throw new RangeError(`The 'classSize' must be > 0.0, but '${classSize}' was used.`);
         }
 
-        const children = { input, _histogram: viewportHistogram(input) };
+        const children = { input, _histogram: viewportHistogram(input, 1000) };
```
VictorVelarde: Maybe we could add an explaining constant here. 
In the future we could even consider exposing this as a ViewportStandardDev parameter, using 1000 as default value

rochoa: To be honest, I changed it because tests were failing. I guess we were relying on the distribution over 1000 bins.

rochoa:
Haha. I'm actually PR to `guides-public-copy-edits`, not to master.

src/renderer/viz/expressions/classification/ViewportStandardDev.js
VictorVelarde: Even for the stdDev calculation?

davidmanzanares: I don't know what do you mean about `stdDev calculation` :confused:

VictorVelarde: I was just wondering about the effect of having a lower number in the precision of the calculation made at `_standardDevFrom(histogram, average)` function

davidmanzanares: ah, yes, but we can/should override the default for that, right?

VictorVelarde: 👍 That's it. We should add it as an external parameter and then choose a good default

PR #1148, opened 2018-11-16
Increase simplify extent

rochoa:
Use extent from MVT metadata: defaults to 4096 but overrides to 2048 in Windshaft.

This will improve some artefacts, especially, in polygons.


Algunenano:
There is another hardcoded 4096 in https://github.com/CartoDB/carto-vl/blob/00ebe5674f6f0058b3a6cbd69629ab822b2c0bed/src/renderer/viz/expressions/ListImage.js#L47.

Could you check if that one should be replaced too?

src/client/windshaft.js
```diff
@@ -292,6 +294,8 @@ export default class Windshaft {
                     options: {
                         sql: aggSQL,
                         aggregation: agg,
+                        vector_extent: TILE_EXTENT,
+                        vector_simplify_extent: TILE_EXTENT,
```
Algunenano: Not necessary (by default if you change `vector_extent`, `vector_simplify_extent` will be the same) , but I like that you added it here as it makes the intent clearer.

rochoa:
Thanks for raising that textureAtlasSize thing. It has nothing to do. 😄

src/renderer/Metadata.js
```diff
@@ -18,6 +18,7 @@ export default class Metadata {
         this.categoryToID = new Map();
         this.IDToCategory = new Map();
         this.numCategories = 0;
+        this.extent = extent || 4096;
```
VictorVelarde: Maybe we could extract 4096 to its own explaining constant (as in windshaft.js), using `DEFAULT_MVT_EXTENT`

rochoa: Done. Thx!

PR #1152, opened 2018-11-19
Avoid pushing the metadata object back and forth in the Worker

rochoa:


VictorVelarde:
@rochoa Will we make further work with this one?

rochoa:
Yes, I would love to remove the back and forth of the metadata in the worker. However, this requires more work than I anticipated.

Issue #1166, opened 2018-11-27
viewportQuantiles doesn't seem to be behaving properly


makella:
I know this has come up a few times in the past and it seems to be happening again where viewportQuantiles doesn't seem right.

For example, I would expect [this map](https://cartodb.github.io/carto-vl/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IDNcbmNvbG9yOiByYW1wKHZpZXdwb3J0UXVhbnRpbGVzKCRhbW91bnQsNSksc3Vuc2V0KVxuc3Ryb2tlV2lkdGg6IDAiLCJmIjp7ImxuZyI6Mi4yMDAwOTI5ODg3Mjk5MjA4LCJsYXQiOjQxLjM5NTIxMzYwNzIxNTkxfSwiZyI6MTAuMjc3MTI3NzUyMzE4MzMyLCJoIjoiRGFya01hdHRlciIsImkiOiJkYXRhc2V0In0=) which uses `viewportQuantiles` to be the same as [this map](https://cartodb.github.io/carto-vl/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IDNcbmNvbG9yOiByYW1wKGdsb2JhbFF1YW50aWxlcygkYW1vdW50LDUpLHN1bnNldClcbnN0cm9rZVdpZHRoOiAwIiwiZiI6eyJsbmciOjIuMjAwMDkyOTg4NzI5OTIwOCwibGF0Ijo0MS4zOTUyMTM2MDcyMTU5MX0sImciOjEwLjI3NzEyNzc1MjMxODMzMiwiaCI6IkRhcmtNYXR0ZXIiLCJpIjoiZGF0YXNldCJ9) which uses `globalQuantiles` since in both maps, all of the data points are in the view.

[...]

PR #1167, opened 2018-11-28
Fix viewportQuantiles

VictorVelarde:
### Fixes https://github.com/CartoDB/carto-vl/issues/1166

`viewportQuantiles` uses `viewportHistogram` internally for its calculations.

This issue is a regression, caused by our last change in the defaults size for `viewportHistogram`, from 1000 to 20 here: https://github.com/CartoDB/carto-vl/pull/1129

This PR recovers the original default value, which seems to work pretty fine for both examples used by @makella


VictorVelarde:
@rochoa We could also expose this as an optional parameter for the `viewportQuantiles` and `viewportStandardDev` classifiers, instead of fixing it.

makella:
💥

PR #1165, opened 2018-11-27
Use only cartovl account for examples

VictorVelarde:
###Fixes #1035

PR #1168, opened 2018-11-28
Store config objects instead of encoded ones

rochoa:
This will help with #1165.


VictorVelarde:
I'm gonna merge this one as it is, and leave the extraction to a 'just data' file outside of this PR

rochoa:
OK. 👌 

Thanks!

Issue #1134, opened 2018-11-07
Update / review examples in editor


VictorVelarde:
There are some internal examples not very well displayed in editor, such as this one:
http://127.0.0.1:8080/examples/editor/index.html#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoid2lkdGg6IHNxcnQoJGFtb3VudC81MDAwKSoyMCooem9vbSgpLzQwMDArMC4wMSkqMS41XG5jb2xvcjogcmFtcCgkY2F0ZWdvcnksIFBSSVNNKVxuc3Ryb2tlV2lkdGg6IDAiLCJmIjp7ImxuZyI6Mi4xNjk5OTk5OTk5OTcxMTcsImxhdCI6NDEuMzgwMDAwMDAwMDA0NTd9LCJnIjoxNCwiaCI6IkRhcmtNYXR0ZXIiLCJpIjoiZGF0YXNldCJ9

IMO it is just a matter of tuning a bit the styles (e.g. those using `zoom()`)

VictorVelarde:
Fixed in #1169

PR #1169, opened 2018-11-28
Examples in editor using `cartovl` user

### LINKED ISSUE #1035: Use the `cartovl` account only
copy `river_directions_upper_co` from Mamata's account into the `cartovl` account

rochoa:
This requires #1168.

This will fix #1035.


VictorVelarde:
I have extracted the 'examples' to its own external file: `examples.js` so it can be easily replaced with other, for example for a demo (ping --> @makella)

I also had modified a bit some examples that were using the 'old' zoom(), so they look better. That part fixes this https://github.com/CartoDB/carto-vl/issues/1134

makella:
ahhh i like this! much easier to "see".

PR #1170, opened 2018-11-28
Change when 'updated' event is raised in `Layer`

### LINKED ISSUE #1159: Intertwined layer updates
It seems that `Layer:updated` event is emitted for a Layer even if the changes have been generated by another, unrelated layer.

This example, prepared by @rjimenezda , helps to reproduce the issue:

```js
const map = new mapboxgl.Map({
            container: 'map',
            style: carto.basemaps.darkmatter,
            center: [-4.77, 37.88],
            zoom: 12
        });

        carto.setDefaultAuth({
            username: 'roman-carto',
            apiKey: 'default_public'
        });

        const POP_MIN = 0;
        const POP_MAX = 2.5e6;

        const viz = new carto.Viz(`
            @duration: 30
            @animation: animation(linear($beginning, time('1910-01-01T00:00:07Z'), time('2018-01-01T00:00:07Z')), @duration, fade(0.1, ${Number.MAX_SAFE_INTEGER}))

            strokeWidth: 0.5
            width: 8
            color: red
            strokeColor: red
  
            filter: @animation
        `);

        const dataViz = new carto.Viz(`
          @histogram: viewportHistogram($cartodb_id, 10)

          strokeWidth: 0
          width: 8
          filter: 1
        `)

        const source = new carto.source.Dataset('cordoba_catastro_simple');
        const dataSource = new carto.source.Dataset('cordoba_catastro_simple');
        const vizLayer = new carto.Layer('layer', source, viz);
        const dataLayer = new carto.Layer('dataLayer', dataSource, dataViz);

        vizLayer.on('updated', function updateWidgets() {
            console.log('animation progress', viz.variables.animation.getProgressPct());
        });

        dataLayer.on('updated', () => {
          console.log('histogram');
        });

        dataLayer.addTo(map, 'watername_ocean');
        vizLayer.addTo(map, 'watername_ocean');
        
        let previousMin = null;
        let previousMax = null;
```

VictorVelarde:
### Fixes #1159 

The first step is to review the set of cases when an 'updated' event has to be emitted:
- [x] a) layer visibility has changed
- [x] b) viz has changed
- [x] c) source has changed
- [x] d) viewport has changed (zoom, center, bearing or pitch) and that requires new dataframes.
- [x] e) an animation has changed
Note: b), c), d) and e) require the layer to be visible to raise an 'updated' event

There is a new **/debug/layer-updated.html** example to test those cases 🔝 

Current situations to fix:
- [x] two layers, with separated handlers, must not received crossed updates
- [x] label-examples: creating and adding another layer inside an update handler must not rise an infinite update loop, as describe in https://github.com/CartoDB/carto-vl/pull/1146#issuecomment-440627161
- [x] an animation should be updated (if no other events occur), just when it is playing


src/Layer.js
```diff
@@ -446,11 +454,11 @@ export default class Layer {
...
 
         if (this.isAnimated()) {
-            this._needRefresh();
-        }
-
-        if (this._state !== states.INIT) {
-            this._fire('updated');
+            this._needRefresh().then(() => {
+                if (this.isPlaying()) {
+                    this._fire('updated'); // this should be restricted to animation.isPlaying?
```
jgoizueta: What do you mean with the comment? (`this should be restricted to animation.isPlaying?`): to only fire update here if `this` is an Animation and isPlaying? (in that case since all this executes only if the Layer `isAnimated()` I see no further need to check `this`)

VictorVelarde: The comment was old, before adding isPlaying.... Now it is not relevant, but the if(isPlaying)... it is!

VictorVelarde:
I'm quite confident on the use of `isAnimated / isPlaying` in the expressions. I have run all the examples to check for any issue, without finding anyone.

`isAnimated` was already been invoked in the main render method, so I just followed the same pattern and replicated it with the new `isPlaying` logic (from the animation expressions up to the Viz & Layer).

> Note: the dataset of the example (`cordoba_catastro`), with Cordoba buildings, comes from @rjimenezda, so thx to him!. Maybe we could even use this dataset (already migrated to cartovl account) in some other animated visualizations...

VictorVelarde:
@rochoa could you have a look at this?

test/integration/user/test/interactivity.test.js
```diff
@@ -515,7 +515,7 @@ describe('regression with a category filter', () => {
     });
 });
 
-describe('regression with blendTo', () => {
+fdescribe('regression with blendTo', () => {
```
rochoa: 😈

test/integration/user/test/layer.test.js
```diff
@@ -34,53 +34,74 @@ describe('Layer', () => {
...
-            let update = jasmine.createSpy('update');
-            await layer.update(source, new carto.Viz('width: now()'));
-            layer.on('updated', update);
-            layer.on('loaded', () => {
-                layer.render();
-                expect(update).toHaveBeenCalled();
-                layer.render();
-                expect(update).toHaveBeenCalled();
-                done();
+                await layer.update(source, new carto.Viz('width: now()'));
```
rochoa: 😍

src/renderer/viz/expressions/transition.js
```diff
@@ -28,23 +28,32 @@ export default class Transition extends BaseExpression {
...
         this.mix = (time - this.aTime) / (this.bTime - this.aTime);
         return Math.min(this.mix, 1.0);
     }
+
     isAnimated () {
         return !this.mix || this.mix <= 1.0;
     }
+
+    isPlaying () {
+        return this.isAnimated();
```
rochoa: While I understand the difference between isAnimated and isPlaying, I think exposing both methods at the API level might be confusing now. Most of the time the end-user will be interested only in the isPlaying one. Thoughts?

Furthermore, seeing the implementation, what's the difference between isPlaying and isAnimated in most of the expressions? Because it's very subtle, isn't it?

VictorVelarde: As a general pattern I would interpret `isAnimated` as "it has an animation", whether it is active now or not, while `isPlaying` is more expressive: "a paused layer, with an animation inside, is not playing, but it is animated (for example, it might have some special controls in the app)".

In this particular case, it is true I had some doubts, cause `isPlaying` is much more clear in an animation context (where `isAnimated` keeps being always true; see xxx), so I'm open to suggestions.

rochoa: Fair enough. Your reasoning about having one for determining if you should have the controls proves that having both is good.

Let's :shipit:!

jgoizueta: I think the difference is clear... but slightly confusing when explained: isAnimated means is not a static visualization/expression, but that contains animations and can be *played* (or be stopped, etc.). IsPlaying is an *is-animated* expression that is not stopped.

jgoizueta: And, yes, I think users should only be interested in `isPlaying`, so `isAnimated` could be renamed `_isAnimated` and kept private.

Issue #1140, opened 2018-11-12
GZIP bundle for CDN


rochoa:
We have to gzip the bundles before uploading them to the CDN.

VictorVelarde:
We have already merged the build improvements, including the gzip of .js & .map files at CDN

PR #1180, opened 2018-12-05
Improve build system

### LINKED ISSUE #1171: Import error v1.0.0 in a webpack & react project
As indicated by @mbforr:

### Description
Using `import carto from '@carto/carto-vl'` in a project (https://github.com/mbforr/isthmus-react/)
(added with `yarn add @carto/carto-vl`) --> `"@carto/carto-vl": "^1.0.0"` in package.json file

It **won't compile** when running with webpack producing this:

### The error is 
```
./node_modules/@carto/carto-vl/dist/carto-vl.js
Module parse failed: /Users/mbforr/Documents/matt/react/isthmus-react/node_modules/@carto/carto-vl/dist/carto-vl.js Unexpected token (15870:32)
You may need an appropriate loader to handle this file type.
|                 const dimensionStats = stats.dimensions[dimName];
|                 const dimType = adaptColumnType(dimensionStats.type);
|                 const { column, ...params } = dimension;
|                 if (properties[column].dimension) {
|                     throw new _errors_carto_maps_api_error__WEBPACK_IMPORTED_MODULE_6__["default"](`${_errors_carto_maps_api_error__WEBPACK_IMPORTED_MODULE_6__["CartoMapsAPITypes"].NOT_SUPPORTED} Multiple dimensions based on same column '${column}'.`);
 @ ./src/components/CARTOVLMap.js 35:15-41
 @ ./src/components/Page.js
 @ ./src/routers/AppRouter.js
 @ ./src/app.js
 @ multi (webpack)-dev-server/client?http://localhost:8081 ./src/app.js
```

VictorVelarde:
### Fixes #1171 and #1140 

Tasks
- [x] Review *main* and *module* in package. See this as previous context:
    - https://github.com/CartoDB/carto-vl/pull/717 on raw vs bundled versions for npm
    - https://github.com/CartoDB/carto-vl/commit/4b795263bf20aa61300c27b2101ba9f7c5e3745f removing babel from 'development' version.
- [x] Fix *main* and *module*. After discussion, this is the proposal:
    - stick to dist versions, cause we are bundling svg + glsl + workers with babel.
    - remove 'module' (with precedence over main in tools like webpack), cause dist versions are not using es6 modules (they are umd bundles).
    - use production .min.js in *main* (development .js is not babelified anymore).
    - add a small note on getting-started guide regarding to the previous point.
    - keep distributing the full src, so a raw import could be fully customized by an advanced user in the future (importing from node_modules)
- [x] rename worker: `carto-vl-worker-[hash].min.js`
- [x] clean tasks. New **build:clean** task, to be run in every build process
- [x] add bundle analyzer: new task **build:min:report** to run on demand
- [x] **gzip**: minified contents deployed to CDN (released versions, not branches)
- [x] refactor publish to be used when deploying a 'release' or the 'current_branch' to our cdn + --dry-run mode


VictorVelarde:
This should be ready for review now.

The only "pending point" is the *temp* folder. We were already using a whitelist in our package.json, just including **src** & **dist**, and *temp^ was already added to '.npmignore' so not sure how it got there. Maybe we could add a `rm -rf` in the new `build:clean` for that folder, but I'm not sure.

docs/guides/01-getting-started.md
```diff
@@ -29,7 +29,10 @@ The most straight-forward way to use CARTO VL is to include the required files f
 </head>
 ```
 **Note:**
-CARTO VL is not compatible with every Mapbox GL version. We recommend using the same version that we use in the [examples](/developers/carto-vl/examples/). However, every version from **`0.50.0`** should work. Historically, we provided patched MGL bundles, but this is no longer required.
+**Mapbox GL**: CARTO VL is not compatible with every version. We recommend using the same version that we use in the [examples](/developers/carto-vl/examples/). However, every version from **`0.50.0`** should work. Historically, we provided patched MGL bundles, but this is no longer required.
+
+**Note:**
+**Developers**: if you have experience with `npm` and a build system in your project (*webpack*, *rollup*...), you can install CARTO VL library with `npm install @carto/carto-vl`. You can import it with `import carto from '@carto/carto-vl'` and then you will have access to an already babelified version of the library, ready to be used.
```
rochoa: 👏

package.json
```diff
@@ -25,7 +25,7 @@
   "license": "BSD-3-Clause",
   "files": [
     "src",
-    "dist"
+    "dist/carto-vl*.*"
```
rochoa: CLEAN! ❤️

package.json
```diff
@@ -43,6 +43,7 @@
     "@babel/core": "^7.0.0",
     "@babel/plugin-transform-runtime": "^7.1.0",
     "@babel/preset-env": "^7.0.0",
+    "aws-sdk": "^2.372.0",
```
rochoa: We should remove the `"s3": "^4.4.0"` dependency if we no longer use it.

scripts/cdn/publish.js
```diff
@@ -3,82 +3,132 @@
...
+    try {
+        files = await readdir('dist');
+    } catch (e) {
+        console.error(e);
+        return;
+    }
+
+    const uploadExtensions = ['.JS', '.MAP'];
+    for (const fileName of files) {
+        const extension = path.extname(fileName).toUpperCase();
```
rochoa: I'm curious, why do you go with uppercase?

scripts/cdn/publish.js
```diff
@@ -3,82 +3,132 @@
...
+}
 
-    uploader.on('error', function (err) {
-        console.error('Error: unable to upload:', err.stack);
-    });
+async function compressAndUploadFile (version, fileName) {
+    // gzip file
+    const localFile = `dist/${fileName}`;
+    let fileContent = await readFile(localFile);
+    fileContent = await gzip(fileContent);
```
rochoa: 😍

package.json
VictorVelarde: It seems it is still being used in `scripts/branch/publish.js` (vs scripts/publish.js)... That is a quite similar script (not currently gzipped) just for branches, so I'm gonna refactor them both and try to remove the dependency

rochoa: Sorry, I didn't check it. 😄 

Much better if you do so. 💪

scripts/cdn/publish.js
VictorVelarde: I was going to say "just in case... 😄 " but it's true, we don't need it, as we control the naming when generating the bundles.

scripts/cdn/publish.js
```diff
@@ -80,54 +81,109 @@ function configureS3Connection () {
...
+    } else {
+        throw new Error('Not a valid publishing mode, use RELEASE or CURRENT_BRANCH');
+    }
 }
 
-publish();
+// main -----------------------------------------------
+const args = process.argv;
+const mode = args[2];
+const SMOKE_MODE = args[3] === 'SMOKE' || false; // to simulate a publication without hitting the CDN
```
rochoa: What about using `--dry-run`?

VictorVelarde: Done! 

```
node scripts/cdn/publish.js release --dry-run
node scripts/cdn/publish.js current_branch --dry-run
```

rochoa:
Everything is 👍 from my side.

PR #1182, opened 2018-12-05
MVT source should handle undefined/null values in Categories

### LINKED ISSUE #1181: Regression: MVT source doesn't work with missing properties
When an MVT source contains an undefined/null value for a Category property, it throws a mismatching type error: `Metadata property is of type 'category' but the MVT tile contained a feature property of type 'undefined': 'undefined'`.

This was working in v0.9.1, I guess we introduced the regression while working on the new codecs for temporal aggregations.

rochoa:
This will fix #1181.

### DIFF
--- a/test/unit/source/MVT.test.js
+++ b/test/unit/source/MVT.test.js
@@ -1,5 +1,6 @@
 import MVT from '../../../src/sources/MVT';
 import MVTMetadata from '../../../src/sources/MVTMetadata';
+import { FP32_DESIGNATED_NULL_VALUE } from '../../../src/renderer/viz/expressions/constants';
 
 describe('sources/MVT', () => {
     describe('constructor', () => {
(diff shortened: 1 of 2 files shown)


test/unit/source/MVT.test.js
```diff
@@ -61,5 +62,43 @@ describe('sources/MVT', () => {
                 metadata.codec('wadus').sourceToInternal(123);
             }).toThrowError(/MVT decoding error. Metadata property is of type \'category\' but the MVT tile contained a feature property of type 'number': \'123\'/);
         });
+        it('Category property should work with undefined value', () => {
```
VictorVelarde: Nice TDD approach!

test/unit/source/MVT.test.js
```diff
@@ -61,5 +62,43 @@ describe('sources/MVT', () => {
...
+            const metadata = new MVTMetadata({
+                properties: {
+                    wadus: {
+                        type: 'category'
+                    }
+                },
+                idProperty: 'id'
+            });
+            metadata.setCodecs();
+            expect(metadata.codec('wadus').sourceToInternal(null)).toEqual(FP32_DESIGNATED_NULL_VALUE);
```
VictorVelarde: We have 'undefined' and 'null'... how about considering **empty string** (with trim)?

rochoa: An empty string will resolve to the same internal ID and the `typeof` won't be undefined. I think it's safe.

PR #1183, opened 2018-12-05
Modify style by number example to have more meaningful data

makella:
@VictorVelarde a quick fix for the `viewport` vs `global` classification of features in the style by numbers example as discussed here: https://github.com/CartoDB/carto-vl/issues/1166

### DIFF
--- a/examples/styling/color-schemes.html
+++ b/examples/styling/color-schemes.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html>
+
 <head>
   <title>Color schemes | CARTO</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
@@ -9,6 +10,7 @@
   <link href="https://api.tiles.mapbox.com/mapbox-gl-js/v0.50.0/mapbox-gl.css" rel="stylesheet" />
   <link rel="stylesheet" type="text/css" href="../style.css">
 </head>
+
 <body>
   <div id="map"></div>
     <aside class="toolbox">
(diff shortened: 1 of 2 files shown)


makella:
@VictorVelarde looking a little closer, doesn't seem like the StDev example is showing above and below the mean... I haven't used it much myself (really at all yet!!) but is that syntax right? should there be more to the syntax?

VictorVelarde:
Regarding to the `viewportStandardDev`, it also LGTM. The logic should be ok, but it would be great if you could have a look at this small test: https://github.com/CartoDB/carto-vl/blob/5a51cff7e91eeebbcb2ddf8d349de9a35be52800/test/unit/renderer/viz/expressions/classification/classifier.test.js#L223

standardDev methods have another interesting parameter (*classSize*), but it is optional. It defaults to 1.0 (standard deviation), but sometimes it could be interesting to use 0.5 stdDevs or even 0.25.

Another comment. If we are changing viewportQuantiles for globalQuantiles, we should probably adapt the rest of the viewport methods (viewportEqIntervals, viewportStandardDev) to their corresponding global versions

makella:
@VictorVelarde yep! i changed all of them...

makella:
i think the biggest issue with the example, is that it isn't the best data for these different classification methods. when we made them at first, we didn't have classification method for polygons so... before merging this, let me try out some other datasets that demonstrate this better! [for example, st dev more informative in this map](https://cartodb.github.io/carto-vl/examples/editor/index.html#eyJhIjoiZWR1Y2F0aW9uX2J5X3JhY2VfY291bnR5X2NvcHlfY29weV8yIiwiYiI6IiIsImMiOiJtYW1hdGFha2VsbGEiLCJkIjoiaHR0cHM6Ly97dXNlcn0uY2FydG8uY29tIiwiZSI6ImNvbG9yOiByYW1wKGdsb2JhbFN0YW5kYXJkRGV2KCRoaWdoZXJfZWRfdG90YWxfcGN0LDUpLHRyb3BpYylcbnN0cm9rZUNvbG9yOiByYW1wKGdsb2JhbFN0YW5kYXJkRGV2KCRoaWdoZXJfZWRfdG90YWxfcGN0LDcpLHRyb3BpYykiLCJmIjp7ImxuZyI6LTkxLjIzNDEzOTU1ODM1OTE3LCJsYXQiOjM2LjUwMjkyNjg0MDc1ODc1fSwiZyI6MywiaCI6IkRhcmtNYXR0ZXIiLCJpIjoiZGF0YXNldCJ9)

VictorVelarde:
Much better example. That one also works well with 5 buckets for stDev

makella:
@VictorVelarde ok! I adjusted the example and I think it is a lot better/more informative now.

I did the following:
- found a better dataset (percent of pop with higher edu by US County)
- re-symbolized each styling method
- made the names of `linear` and `buckets` viz objects more meaningful to the example (`interpolate` and `manual`)
- re-organized the order of them in the radio-button selector

Maps:
screen shot 2018-12-05 at 12 20 22 pm
screen shot 2018-12-05 at 12 20 31 pm
screen shot 2018-12-05 at 12 20 42 pm
screen shot 2018-12-05 at 12 20 50 pm
screen shot 2018-12-05 at 12 21 08 pm

VictorVelarde:
Great job @makella I have checked them both locally and they look much better now!
IMO it's ready to be merged

Maybe we could add a pair of links in the color-schemes 'box', to the CARTOColors and ColorBrewer if you find it useful too, like:
```
<label for="Cartocolor">CARTOColors <a href="https://carto.com/carto-colors/" title="CARTOColors" target="_blank">more?</a></label>
```

makella:
@VictorVelarde thanks for looking through!

ok, added the links for CC and CB (actually, for CB, I could link directly to the scheme that is used in the map which is nice).

tell me what you think

makella:
ok, think i'll merge this!

thanks @VictorVelarde

Issue #995, opened 2018-09-25
viewport histogram crashes with null values


davidmanzanares:
Uncaught exception:
```
TypeError: Cannot read property 'localeCompare' of null
    at _sortNumerically (ViewportHistogram.js:232)
```

VictorVelarde:
@rochoa This should probably be developed after https://github.com/CartoDB/carto-vl/pull/1162
Do you want me to take it or will you do it?

rochoa:
I can check this one after we merge #1162.

PR #1187, opened 2018-12-10
Fix viewportHistogram with null values

VictorVelarde:
### Fixes https://github.com/CartoDB/carto-vl/issues/995


src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
```diff
@@ -272,9 +272,17 @@ function _getCategoryValue (histogram) {
...
 
 function _sortNumerically (a, b) {
-    if (b.y - a.y === 0) {
-        return a.x.localeCompare(b.x);
+    const frequencyDifference = (b.y - a.y);
+    if (frequencyDifference === 0) {
+        const categoryA = a.x;
+        const categoryB = b.x;
+
+        if (!categoryA && !categoryB) { return 0; } // both null or undefined
```
rochoa: I think it's not critical at all, but this will not sort properly `null` and `undefined` values.

I was also thinking if doing:

```js
if (!categoryA || !categoryB) { return (!categoryA) - (!categoryB); }
```

Makes things simpler or more complicated. 🤔

VictorVelarde: Hi @rochoa I think the implementation works well with these three cases: null, undefined and stringEmpty. Would you please check my test, to see if I'm missing anything (link here)?

The test goes like this. 6 features, with these values for 'category':
- 1 --> 'a'
- 2 --> 'a'
- 3 --> 'b'
- 4 --> null
- 5 --> '' (stringEmpty)
- 6 --> undefined

The frequencies calculated (no changes here, I think they were fine):
- 'a' --> 2
- 'b' --> 1
- '' --> 1
- null --> 2 (null and undefined have been merged, which IMO it's ok)


And this is the ordering part (where it was failing), after the fix. It is important to notice that entries must be first order by the frequency (and then by the category name):

1.  { x: 'a', y: 2 },
2.  { x: null, y: 2 },
3.  { x: 'b', y: 1 },
4.  { x: '', y: 1 }

Isn't this ok? 🔝 

Regarding to your suggestion, I really like the compactness but I would prefer to be more verbose in that part, specially as it was one with a bug.

rochoa: Agree on using the more verbose approach for the code.

About the sorting, if you have several `null` and `undefined` keys, you will see they mix. Let me add a test to describe the behaviour. It will be simpler than explaining it.

VictorVelarde: Ok, I see, you would prefer to keep them separated, having both keys independently: null & undefined. 

That wasn't causing any error, and it is more like a 'feature', but it is a good moment to implement it if that's what we need.

test/integration/user/test/viewportHistogram.test.js
```diff
@@ -0,0 +1,154 @@
...
+            features: [feature1, feature2, feature3,
+                featureWithNullCategory, featureWithEmptyStringCategory, featureWithUndefinedCategory]
+        };
+
+        const histogramViz = '@histogram: viewportHistogram($category)';
+        createMapWith(histogramViz, featuresWithSomeNulls);
+
+        layer.on('loaded', () => {
+            const histogram = viz.variables.histogram.value;
+            expect(histogram).toEqual([
```
rochoa: BTW, aren't we missing the _`undefined`_ feature here?

VictorVelarde: That's what I was saying in my comments 🔝 
Both features are been counted under the 'null' category, which I think it's pretty fine.

src/renderer/viz/expressions/aggregation/viewport/ViewportHistogram.js
rochoa: Well, now that I look closer, the histogram itself is preventing the behaviour I was describing, as the histogram will group the same `y` values.

My initial concern was about having something like:

```js
const input = [null, 'd', 'b', 'c', '', 'a', null, ''].map(x => ({ x, y: 0 }));
const result = input.sort(_sortNumerically);
```

Which will result in:

screen shot 2018-12-11 at 13 17 50


For now, I think we can live with the _non-pure_ sorting of `null`, `undefined` and empty strings.

test/integration/user/test/viewportHistogram.test.js
rochoa: Hahaha. See my latest https://github.com/CartoDB/carto-vl/pull/1187#discussion_r240586723.

Sorry for the noise, but I wanted to understand better what was going on.

rochoa:
Let's :shipit: !

PR #1188, opened 2018-12-11
Fix the use of variables in a classification

VictorVelarde:
### Context
* All expressions should allow the use of 'variables', as per https://github.com/CartoDB/carto-vl/pull/650
* ...but **Classifiers** functions included several type assumptions + early type checks, throwing exceptions when using variables at constructors (even when evaluated those expressions have eventually correct types). 

In some other classes, not classifiers, it was enough to just remove some checks, but here it has needed a bigger refactor (including *Classifier* itself), to allow the use of 'aliases' without breaking the proper initialization.

### Done in this PR
Allow the use of a variable for **input** & **buckets** for all:
- [x] globalEqIntervals
- [x] globalQuantiles
- [x] globalStandardDev
- [x] viewportEqIntervals
- [x] viewportQuantiles
- [x] viewportStandardDev

Allow the use of a variable for **classSize** and **histogramSize** (required parameters for certain expressions, such as viewportStandardDev)


src/renderer/viz/expressions/classification/ViewportQuantiles.js
```diff
@@ -68,8 +93,23 @@ export default class ViewportQuantiles extends Classifier {
...
         });
     }
+
+    _getAccumHistogramFrom (histogram) {
+        let prev = 0;
+        const accumHistogram = histogram.map(({ y }) => {
+            prev += y;
+            return prev;
+        });
+        return accumHistogram;
```
rochoa: Why not "reduce"?

```
return histogram.reduce((acc, { y }) => {
    return acc + y;
}, 0);
```

test/unit/renderer/viz/expressions/classification/classifier.test.js
```diff
@@ -69,78 +68,100 @@ describe('src/renderer/viz/expressions/classifier', () => {
     });
 
     describe('eval', () => {
```
rochoa: 👏 I see the connections with the changes you made.

src/renderer/viz/expressions/classification/ViewportQuantiles.js
VictorVelarde: I had a look, and the map preserves the original array disposition, which is needed. Using reduce here would require accumulating but also returning the array on every iteration, so IMO map is nice in this particular case.

src/errors/carto-validation-error.js
```diff
@@ -16,5 +16,5 @@ export const CartoValidationTypes = {
     MISSING_REQUIRED: '[Missing required property]:',
     INCORRECT_TYPE: '[Property with an incorrect type]:',
     INCORRECT_VALUE: '[Incorrect value]:',
-    TOO_MANY_ARGS: '[Too many arguments]:'
+    WRONG_NUMBER_ARGS: '[Wrong number of arguments]:'
```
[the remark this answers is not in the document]
VictorVelarde: Good point!. I have refactored the types, to keep the compatibility while adding new cases.

src/renderer/viz/expressions/classification/ViewportQuantiles.js
rochoa: Got it! Thanks for the offline explanation. 😅

Issue #1102, opened 2018-10-27
Animations never sleep!


jgoizueta:
For an `Animation`, [`isAnimated`](https://github.com/CartoDB/carto-vl/blob/master/src/renderer/viz/expressions/Animation.js#L115) returns always true, even while `_paused`.

If we fix that (`this.paused` -> `this._paused`), then paused animations don't resume (with `play`) until some other change occurs, because in the `Layer` `render`, the [checking of `isAnimated()`](https://github.com/CartoDB/carto-vl/blob/master/src/Layer.js#L414-L416) for refreshing is ineffective, as `render` stops being called when animation is paused.

Also, when the update is forced by changing the map zoom or calling `map.triggerRepaint()`, the animation jumps to where it would have been if not paused. It seems the progress adjustment fails if animation hasn't been  receiving `_setTimestamp` while paused.

davidmanzanares:
I think that technically this is not a bug, it is something that it is pretty close to becoming a bug.

Right now, with the `this.paused -> this._paused` the animation never sets the render into the `idle` state. This is not desired behavior, and I agree this could be changed.

Once we change that, we will need to inform the `Layer` on `Animation.play()` to request a new animation frame. This can be done by calling the Animation inherited method `notify()`. This should work, but I don't know if it is overkill and if it will create a performance regression.

For all of these reasons, I would change `Animation.isAnimated` to return always true. And in the future fix this properly (calling notify and checking it works as expected, or adding a new lightweight notify-like method).

jgoizueta:
OK, I already [set it to true](https://github.com/CartoDB/carto-vl/pull/991/files#diff-94085fcffacdd5d83bb78d4682a11e2bR56) so what I'd do know is:

* clarify the documentation of isAnimated
* add new public methos `paused` or `isPaused` (`return this._paused`)

VictorVelarde:
Just for completeness, a new isPlaying() method has been added here: https://github.com/CartoDB/carto-vl/pull/1161

VictorVelarde:
IMO This can be closed after merging the new approach in https://github.com/CartoDB/carto-vl/pull/1170:
* `isAnimated` will keep returning always 'true'
* 'isPlaying' will be dynamic
* a paused animation will not raise an 'updated event' 
* no 'jump' will happen on playing again a paused animation.

VictorVelarde:
Animations will definitively have a rest after the work in #1190... 💤

PR #1190, opened 2018-12-11
Fix regressions on Layer:updated

### LINKED ISSUE #1189: Incorrect layer updated events
With the changes in https://github.com/CartoDB/carto-vl/pull/1170, we have introduced a few regressions. The ones I'm aware:

1. The `updated` event is not fired when the viewport changes—it fires when a dataframe is added/removed.
1. The `updated` event is not properly fired after a `blendTo` operation. It looks like it's missing the last `updated` event. If you rely on that event to access the value of a histogram, you might not get final representation after the blendTo operation finishes.

VictorVelarde:
### Regressions introduced in #1170 

* **regression 1**: the updated event is not fired when the `viewport` changes but no dataframes are added/removed. It should fire always on that kind of change.
* **regression 2**: the updated event is not properly fired after a `blendTo` operation.

---
### Analysis and tasks

- [x] **regression 1** 
In this case, it was a misunderstanding on the need to restrict the `Layer:updated` event to just *dataframe updates*. 
To fix it the tasks are:
   - [x] failing test to illustrate the problem
   - [x] fire event on any 'viewport' change (based on cameraMatrix changes)

---

- [x] **regression 2**
Before #1170 there was an "updated" event emitted on any render call (except on init). See the changes [here](https://github.com/CartoDB/carto-vl/pull/1170/files#diff-b97124864856432df2f3ab3a5fa12871). In that point in time there were some buggy situations, like two layers receiving crossed updates and always-emitting-animations (although *blendTo* operations were working fine). So we removed that event to fix those situations and that has caused some operations updating the viz (eg. `blendTo`) to lose some of their corresponding "updated" events. We also restricted the animation updated events to `isPlaying` (but it was kept "awaken" to keep its timestamp in sync).
Todo:
   - [x] make animations sleep (and resume correctly). See #1102 as a reference.
   - [x] better identification of each 'updated' event and its cause
   - [x] recover events caused by minor viz changes (e.g blendTo)
   - [x] minor refactors in Layer for more clarity

--- 

### Event triggers
I'm also gonna extend / modify here the catalog of 'updated' _events triggers_, indicating if they are implemented or not, in case we need to discuss on them:

- [x] a) layer visibility has changed
- [x] b) viz has been updated
- [x] c) source has been updated
- [x] d) viewport has changed (zoom, center, bearing or pitch) and that requires new dataframes or NOT (**regression 1**)
- [x] e) an animation isPlaying
- [x] f) blendTo updates (**regression 2**).

Note: All cases but a) require the layer to be visible to raise an 'updated' event

### DIFF
--- a/test/integration/user/test/layer.test.js
+++ b/test/integration/user/test/layer.test.js
@@ -42,6 +42,12 @@ describe('Layer', () => {
                 spyOn(layer, '_needRefresh').and.callFake(() => Promise.resolve());
             }
 
+            function resetMocks () {
+                layer._getZoom.calls.reset();
+                layer._getViewport.calls.reset();
+                layer._needRefresh.calls.reset();
+            }
+
             it('should fire an "updated" event when ready', (done) => {
                 layer.on('updated', done);
             });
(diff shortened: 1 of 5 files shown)


test/integration/user/test/layer.test.js
```diff
@@ -67,23 +73,51 @@ describe('Layer', () => {
...
+                    const currentMatrix = layer.renderer.matrix;
+                    layer.prerender(null, currentMatrix);
 
                     expect(update).toHaveBeenCalledTimes(1);
+                    resetMocks();
                     done();
                 });
             });
 
+            it('should fire a new "updated" after moving the map (even if not requiring new dataframes)', (done) => {
```
rochoa: 👏

PR #1194, opened 2018-12-19
labeling tests 

makella:
@VictorVelarde put what we worked on yesterday in this branch debug folder... `geojson-label-centroid.html` and we can use this branch for our testing... :)

### DIFF
--- a/debug/geojson-label-centroid.html
+++ b/debug/geojson-label-centroid.html
@@ -1,0 +1,150 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <title>Labels & points | CARTO</title>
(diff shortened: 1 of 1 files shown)


VictorVelarde:
I'm adding this link as a reference to the previous conversations on the topic: https://github.com/CartoDB/product/issues/150. 

It has some interesting notes, such as how we can currently access the feature.size property, and use it in relation to a label-size

VictorVelarde:
@makella IMO we should probably delete this branch, now that we have the labeling examples at: https://github.com/CartoDB/carto-vl/pull/1209

We can always open a new one in the future if we need more advanced labeling examples...

makella:
👍

PR #1197, opened 2018-12-19
Fix for global aggregations without property

### LINKED ISSUE #1196: globalCount() is broken
Some dude broke `globalCount` [here](https://github.com/CartoDB/carto-vl/commit/451408f98d0fa25e1b0966621e43ad155918421b) by using unconditionally the property argument that, in the case of this expression, is not present.

The fix is a matter of

```diff
--- a/src/renderer/viz/expressions/aggregation/global/GlobalAggregation.js
+++ b/src/renderer/viz/expressions/aggregation/global/GlobalAggregation.js
@@ -28,7 +28,9 @@ export default class GlobalAggregation extends BaseExpression {
     }

     _resolveAliases (aliases) {
-        this.property._resolveAliases(aliases);
+        if (this.property) {
+            this.property._resolveAliases(aliases);
+        }
     }

     _bindMetadata (metadata) {
```

jgoizueta:
Fixes #1196 

This wasn't detected by the tests, so we should add some. There is a [unit test](https://github.com/CartoDB/carto-vl/blob/d2b2e1c02b3e7bc128c95f34e973c5e8e23c2535/test/unit/renderer/viz/expressions/aggregation/globalAggregation.test.js#L56-L60) already, but it wasn't failing; I guess `_resolveAliases` wasn't being call. We should either call `_resolveAliases` in that test or add some new render test using globalCount.

Pending

- [x] Add `globalCount()` test


jgoizueta:
I've introduced calls to `_resolveAliases` in the global aggregations unitary tests, so that the problem would now be detected for that test.

I have no idea why after that a e2e has broken 🤷‍♂️

jgoizueta:
That failing timerange e2e test seems to now be unstable. I have no permissions to rerun the test, but it's passing here on the very same code: https://circleci.com/gh/CartoDB/carto-vl/17687?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #1198, opened 2018-12-19
Control validity of each global aggregation of cluster aggregation case

### LINKED ISSUE #1193: Give us our clustered metadata back
Global aggregations used to work with cluster-aggregated values, but not anymore since the timeaggregation changes: https://github.com/CartoDB/carto-vl/commit/94d63823030c7f2d8cd5c12fed92af8ee6d868af

Back then, (before time aggregations), for an expression like `globalAvg(clusterAvg($column))`, the metadata of `$column` was used to compute the global aggregation, which was technically incorrect in general (the average of cluster averages is not the average of the raw data). Now this kind of expression causes an error.

This can be inconvenient mostly because `linear` without explicit limits uses `globalMin` and `globalMax` for the limits, sor for cluster aggregations we must put explicit limits in linear expressions.

Although it was technically incorrect in some cases, the old behaviour should be correct for min, max, and only slightly off in most cases for avg.

So it could be nice to get back to it and allow automatic `linear` limits for cluster expressions.
I think this could be achieved as simply as:

```diff
--- a/src/renderer/viz/expressions/aggregation/global/GlobalAggregation.js
+++ b/src/renderer/viz/expressions/aggregation/global/GlobalAggregation.js
@@ -36,7 +36,7 @@ export default class GlobalAggregation extends BaseExpression {
         this.property._bindMetadata(metadata);
         const propertyName = this.property.propertyName || this.property.name;

-        const stats = metadata.stats(propertyName);
+        const stats = metadata.stats(propertyName) || metadata.stats(this.property.name);
         // TODO improve type check
         if (!stats || stats[this._name] === undefined) {
             throw new CartoValidationError(`${cvt.MISSING_REQUIRED} Metadata ${this._name} for property ${propertyName} is not defined`);
```

Maybe be could restrict it to `min`, `max`, `avg` agregations (`this._name` in that function), or even only to `min`, `max` which are the *correct* cases and the ones needed for `limit`.

cc @makella

jgoizueta:
Fixes #1193

### DIFF
--- a/src/renderer/viz/expressions/aggregation/global/GlobalAggregation.js
+++ b/src/renderer/viz/expressions/aggregation/global/GlobalAggregation.js
@@ -8,11 +8,12 @@ export default class GlobalAggregation extends BaseExpression {
      * @param {*} property
      * @param {*} name
      */
-    constructor ({ property, name, type }) {
+    constructor ({ property, name, type, baseStats = false }) {
         super({ _value: number(0) });
         this.property = implicitCast(property);
         this._name = name;
         this.type = type;
+        this.baseStats = baseStats;
         super.inlineMaker = inline => inline._value;
     }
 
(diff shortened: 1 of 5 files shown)


jgoizueta:
The intent of this PR is to provide support for the use of `linear` on cluster aggregations without explicit limits, which used to work in the past. So I would some tests for that case.

Regarding the explicit combination of global aggregation expressions with cluster aggregations, I don't think that it's very useful for visualizations, so we could avoid officially supporting that case. We could just mention that global aggregations should have a property as the input parameter, and that perhaps we could also document that the *well-defined combinations* (globalMin of clusterMin, globalMax of clusterMax, globalSum of clusterCount and globalSum of clusterSum) are allow and work, but that any other will produce an error or inaccurate results.

jgoizueta:
Hey, the broken test seems to have some race condition with exquisite or something, see https://github.com/CartoDB/carto-vl/pull/1197#issuecomment-449474438

PR #1199, opened 2018-12-24
Improve Interactivity performance and fix unwanted feature events

### LINKED ISSUE #1191: interactivity bug on hover
As reported internally:

Having `featureEnter` and `featureLeave` set up, hover interactivity continues when the map is clicked and dragged, causing the browser to freeze up after a bit

dragvl

VictorVelarde:
### Closes #1191 
- [x] Fix featureEnter / featureLeave while dragPan (issue #1191), and in general while "moving" the map to improve performance.
- [x] Add enable / disable / isEnabled public methods to the API
- [x] Add interaction tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Add `getCentroid` method to features
 - Add `isPlaying` method to animations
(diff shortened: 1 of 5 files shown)


VictorVelarde:
@rochoa This should fix the problem with dragPan, but it would be nice to check the performance gain

rochoa:
👏 Good job, @VictorVelarde!

Issue #1215, opened 2019-01-04
Wrong centroid position for polygons


VictorVelarde:
Reviewing the ongoing work for labelling examples with polygons, I have seen some, apparently random, wrong positions when obtaining them with `getCentroid`. 

This looks like this:

VictorVelarde:
The getCentroid is attached to Features coming from:
- `viewportFeatures` (the case we are analyzing here...).
- `Interactivity` (to be analyzed)

After reviewing the calculations for a while, a situation has finally arisen: the `getCentroid` operation works at the **dataframe** level, but there is no current mechanism to consider the different 'pieces' that could lie in several dataframes and how the centroid should be calculated for the *whole geometry*

This is the current code in `ViewportAggCalculator.js` which runs the viewport aggregations:
```js
const featureAlreadyAccumulated = processedFeaturesIDs.has(featureId);
if (featureAlreadyAccumulated) {
    continue;
}
```
So if the feature in the dataframe (a piece of it) has already been included in the list of features to be processed later by the viewport expressions, new occurrences are ignored. That's fine for alphanumeric properties (and performance), but causes that, when using viewportFeatures, the 'first piece of feature (and its centroid)' wins, so that's the feature & centroid it is recovered later on. The current calculation for polygons is fine, but just considering inside of a single dataframe.

That's why:
- It doesn't happen with GeoJSON data sources (1 dataframe), like the ones used in the debug examples.
- It doesn't manifest so clearly when using a lower zoom (the feature lies inside just one dataframe)
- This doesn't affect point geometries, which always fit inside just 1 dataframe.

Alternatives

VictorVelarde:
#### Performance for viewportFeatures
- Points: are is not affected
- Lines / Polygons:
    - eg */hurricane-harvey.html* (about 2500 polygons animation). Adding previously a `@v_features: viewportFeatures($e_totpop)` to the viz, there is no remarkable drop in FPS.
    - eg. */debug/layer-updated.html* (about 29000 polygons anim). Adding previously `@v_features: viewportFeatures($beginning)` to the animation, drops from 11-12FPS to 10-11 FPS aprox.

So it's clear it needs more calculations, and performance must be a bit slower, but it doesn't look like a blocker.

PR #1217, opened 2019-01-04
Fix getCentroid for fragmented features and rename to getRenderedCentroid

VictorVelarde:
### Related to https://github.com/CartoDB/carto-vl/issues/1215

*Note*: starting from a Mamata's branch (nothing remarkable for the current issue). 

See the discussion below ⬇️ starting [here](https://github.com/CartoDB/carto-vl/pull/1217#issuecomment-451899937)

### DIFF
--- a/src/interactivity/commonFeature.js
+++ b/src/interactivity/commonFeature.js
@@ -1,0 +1,28 @@
+
+import { average } from '../renderer/viz/expressions/stats';
+
+/**
+ * Get a 'compound Feature'. It is a Feature created from a set of 'feature' pieces
+ * @param {Array} featurePieces
+ */
+export function getCompoundFeature (featurePieces) {
+    if (featurePieces.length === 1) return featurePieces[0];
+
+    const centroids = featurePieces.map(piece => piece.getCentroid());
+    const exemplar = featurePieces[0];
+
+    // Unify geometry-related properties, into an 'exemplar'.
+    delete exemplar.getCentroid;
+    Object.defineProperty(exemplar, 'getCentroid', {
+        get: function () {
+            const getCentroid = () => {
+                const avgX = average(centroids.map(c => c[0]));
+                const avgY = average(centroids.map(c => c[1]));
+                return [avgX, avgY]; // vs this._rawFeature._dataframe.getCentroid(this._rawFeature._index);
+            };
+            return getCentroid;
+        }
+    });
+
+    return exemplar;
+}
--- a/src/interactivity/commonFeature.js
+++ b/src/interactivity/commonFeature.js
@@ -6,13 +6,15 @@ import { average } from '../renderer/viz/expressions/stats';
  * @param {Array} featurePieces
  */
 export function getCompoundFeature (featurePieces) {
-    if (featurePieces.length === 1) return featurePieces[0];
-
-    const centroids = featurePieces.map(piece => piece.getCentroid());
     const exemplar = featurePieces[0];
+    if (featurePieces.length === 1) return exemplar;
 
     // Unify geometry-related properties, into an 'exemplar'.
-    delete exemplar.getCentroid;
+    const allCentroids = featurePieces.map(piece => piece.getCentroid());
+    const centroids = allCentroids.filter(c => {
+        return !(isNaN(c[0]) || isNaN(c[1]));
+    });
+
     Object.defineProperty(exemplar, 'getCentroid', {
         get: function () {
             const getCentroid = () => {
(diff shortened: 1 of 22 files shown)


VictorVelarde:
##  1. The problem
The getCentroid is attached to Features coming from:
- `viewportFeatures`
- `Interactivity` (to be analyzed)

After reviewing the calculations for a while, a situation has finally arisen: the `getCentroid` operation currently works at the **dataframe level**, but there is no mechanism yet to consider the different 'pieces' (that could lie in several dataframes) and how the centroid should be calculated for the **whole set of geometries**.

### a) viewportFeatures
This is the current code in `ViewportAggCalculator.js` which runs the viewport aggregations:
```js
const featureAlreadyAccumulated = processedFeaturesIDs.has(featureId);
if (featureAlreadyAccumulated) {
    continue;
}
```
So if the feature in the dataframe (a piece of it) has already been included in the list of features to be processed later by the viewport expressions, new occurrences are ignored. That's fine for alphanumeric properties (and performance), but causes that, when using viewportFeatures, the *first piece of feature (and its centroid) wins*, so that's the feature & centroid it is recovered later on. The current calculation for polygons looks fine but just when considering it for a single dataframe, which is not so good.

Commenting out the previous lines it is not a solution, but it displays the situation more clearly (displaying the centroid of all the 'pieces'):
several_centroids

That's why:
- This doesn't happen with GeoJSON data sources (1 dataframe), like the ones used in the debug examples.
- This doesn't manifest so clearly when using a lower zoom (when the features lie inside just one dataframe)
- This doesn't affect point geometries, which always fit inside just 1 dataframe.

### b) Interactivity

This gif shows the centroid obtained from a featureClick event, with Interactivity
centroid-interactivity

The problem is similar, in the sense that it just returns the centroid for a "portion of the feature" (the one under the click coordinates).

In this case query method is `getFeaturesAtPosition` at `RenderLayer`.


## 2. Approach
- Recalculate the centroids considering the whole geometry, for lines and polygons:
    - [x] viewportFeatures 
    - [ ] Interactivity
- Rename `getCentroid` to something like `getClientCentroid`, to better convey the idea of 'client-side calculation' (just considering visible tiles, not the whole feature in the server).

VictorVelarde:
#### Performance for viewportFeatures
- Points: are is not affected
- Lines / Polygons:
    - eg */hurricane-harvey.html* (about 2500 polygons animation). Adding previously a `@v_features: viewportFeatures($e_totpop)` to the viz, there is no remarkable drop in FPS.
    - eg. */debug/layer-updated.html* (about 29000 polygons anim). Adding previously `@v_features: viewportFeatures($beginning)` to the animation, drops from 11-12FPS to 10-11 FPS aprox.

So it's clear it needs more calculations, and performance must be a bit slower, but it doesn't look like a blocker.

VictorVelarde:
### Fix results

#### Step 1
Features with multiple feature pieces
image

#### Step 2
Multiple pieces, ponderated by area
image

src/renderer/RenderLayer.js
```diff
@@ -60,17 +63,90 @@ export default class RenderLayer {
...
+        }
+        return features;
+    }
+
+    /**
+     * Get all the pieces from rawFeatures.
+     * Return is a nested Array; each entry includes an array with (potentially) several feature pieces
+     */
+    _getPartialFeaturesFromSingle (rawFeatures) {
+        const idProperty = this.viz.metadata.idProperty;
```
rochoa: Should we use `this.idProperty`? Or is it unsafe?

src/interactivity/commonFeature.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Get a 'compound Feature'. It is a Feature created from a set of 'feature' pieces
+ * @param {Array} featurePieces
+ */
+export function getCompoundFeature (featurePieces) {
+    const exemplar = featurePieces[0];
+    if (featurePieces.length === 1) return exemplar;
+
+    const geomProperties = _getGeomPropertiesFromPieces(featurePieces);
+    Object.defineProperty(exemplar, 'getCentroid', {
```
rochoa: What's the reason to use a getter here?

VictorVelarde: I was trying to gain some performance here, by lazy calculating some stuff just when invoking the method, the same as when creating a Feature from scratch

src/renderer/RenderLayer.js
VictorVelarde: Good point, let me check

Issue #1216, opened 2019-01-04
MVT 'category' property returns a suspected numeric value


AdriSolid:
Hi all,

I'm testing your MVT creation using Mapnik. I'm trying to extract 'category' properties in order to do stuffs later with a 'featureEnter' event, but always it returns a numerical value. 

As you can see below (simplified example), your public table 'ne_10m_populated_places_simple' has a string column called 'name', which is the target. I feel like I am missing out... 

```JavaScript
<!DOCTYPE html>
<html>

<head>
  <title>MVT Test | CARTO</title>
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta charset="UTF-8">
  <script src="https://libs.cartocdn.com/carto-vl/v1.0.0/carto-vl.min.js"></script>
  <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v0.50.0/mapbox-gl.js"></script>
  <link href="https://api.tiles.mapbox.com/mapbox-gl-js/v0.50.0/mapbox-gl.css" rel="stylesheet" />
  <link href="https://carto.com/developers/carto-vl/examples/maps/style.css" rel="stylesheet">
</head>

<body>
  <div id="map"></div>
  <div id="loader">
    <div class="CDB-LoaderIcon CDB-LoaderIcon--big">
      <svg class="CDB-LoaderIcon-spinner" viewBox="0 0 50 50">
        <circle class="CDB-LoaderIcon-path" cx="25" cy="25" r="20" fill="none"></circle>
      </svg>
    </div>
  </div>
  <script>
    const map = new mapboxgl.Map({
      container: 'map',
      style: carto.basemaps.voyager,
      center: [8.234, 46.802],
      zoom: 7
    });

    const nav = new mapboxgl.NavigationControl({
      showCompass: false
    });
    map.addControl(nav, 'top-left');

    async function loadMVT() {
      const mapConfig = {
        buffersize: { mvt: 0 },
        layers: [
          {
            id: 'testLayer',
            type: 'mapnik',
            options: {
              sql: `SELECT cartodb_id,
                           the_geom,
                           the_geom_webmercator,
                           name
                    FROM ne_10m_populated_places_simple
                    WHERE adm0name = 'Switzerland'`
            }
          }
        ]
      }
      const response = await fetch('https://cartovl.carto.com/api/v1/map', {
        method: 'POST',
        headers: {
          'Accept': 'application/json',
          'Content-Type': 'application/json'
        },
[...]

PR #1222, opened 2019-01-08
Set codecs at constructor

rochoa:
Configuring the codecs should happen at the constructor because we already have the metadata.

However, this doesn't work because the codecs hold a reference to the metadata. Cloning the source will fail with a JSON circular reference in https://github.com/CartoDB/carto-vl/blob/a2fdec46c9707a3b7b3ab0cb584a256d37ebe9cd/src/sources/MVT.js#L75-L77

We already had [a branch that was removing those references](https://github.com/CartoDB/carto-vl/tree/codecs-metadata-avoid-references). The initial proposition was more about performance, but now we have a better reason to merge that branch: being able to clone the sources without that circular structure issue and configuring the codecs at the right place.

What do you think?

### DIFF
--- a/src/client/windshaft.js
+++ b/src/client/windshaft.js
@@ -401,9 +401,7 @@ export default class Windshaft {
 
         const idProperty = 'cartodb_id';
 
-        const metadata = new Metadata({ properties, featureCount, sample: stats.sample, geomType, isAggregated: aggregation.mvt, idProperty });
-        metadata.setCodecs();
-        return metadata;
+        return new Metadata({ properties, featureCount, sample: stats.sample, geomType, isAggregated: aggregation.mvt, idProperty });
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix issue sorting null key in `viewportHistogram`
 - Fix unwanted `Interactivity` events while using dragPan (and performance improvement)
 - Fix `linear` applied to cluster aggregations without explicit limits
+- Fix `MVT` category property returning a numeric value
 
 ### Changed
 - Change examples to use Mapbox GL version 0.52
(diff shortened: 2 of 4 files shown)


rochoa:
BTW, this will fix the issue described at https://github.com/CartoDB/carto-vl/issues/1216.

src/client/windshaft.js
```diff
@@ -401,9 +401,7 @@ export default class Windshaft {
 
         const idProperty = 'cartodb_id';
 
-        const metadata = new Metadata({ properties, featureCount, sample: stats.sample, geomType, isAggregated: aggregation.mvt, idProperty });
-        metadata.setCodecs();
-        return metadata;
+        return new Metadata({ properties, featureCount, sample: stats.sample, geomType, isAggregated: aggregation.mvt, idProperty });
```
VictorVelarde: Should this same change be applied to GeoJSON?

VictorVelarde: If base `Metadata` is not the place anymore to properly `setCodecs`, maybe we could 'defend' from unwanted uses creating there something like:

```js
    setCodecs() {
        throw new CartoRuntimeError(`You must call 'setCodecs' once you have determined the proper subclass`);
    }
```

Does it make sense to you?

rochoa: I might not be understanding your comment about GeoJSON. We are already doing something similar in the GeoJSONN source, see https://github.com/CartoDB/carto-vl/blob/a2fdec46c9707a3b7b3ab0cb584a256d37ebe9cd/src/sources/GeoJSON.js#L159-L160

rochoa: Fair point about the deffensive code in the base metadata class. I will add that.

rochoa:
I prefer to discuss #1223 and merge into master instead of mixing the two things in just one PR. If that other PR gets into master, I will update this to check e2e tests.

src/client/windshaft.js
VictorVelarde: Regarding to GeoJSON, I thought both lines would rise this very same 'deffensive error', but I have just seen that Metadata there is already a GeoJSONMetadata, so forget about it.

rochoa:
After merging #1223 into master and updating this PR with master, all tests are green. 🎉

PR #1223, opened 2019-01-08
Avoid holding references to the metadata object in the codecs

rochoa:


VictorVelarde:
Only for curiosity, what's the effect in the performance here?

rochoa:
This brach, per se, not much. This is the first step to avoid transferring metadata back and forth to the worker. The real performance gains are at https://github.com/CartoDB/carto-vl/tree/avoid-metadata-transfer. However, that requires much more work yet.

Issue #1149, opened 2018-11-16
Handle unsupported WebGL features


rochoa:
When, for instance, a WebGL extension is not supported, we don't handle the error and the map doesn't load, neither it does fail or show any error message.

<img width="899" alt="screen shot 2018-11-16 at 14 52 40" src="https://user-images.githubusercontent.com/45346/48627911-23476000-e9b6-11e8-8237-68c574ad32ae.png">

rochoa:
I was thinking about having a utility function to be able to check the minimum set of requirements: e.g. `OES_texture_float` extension, the minimum value for `gl.MAX_RENDERBUFFER_SIZE`, etc. So you can use it preemptively in your application instead of having to try-catch those errors on runtime.

Instead of doing:

```js
try {
  layer.addTo(map, 'watername_ocean');
} catch(err) {
  // check for runtime error and display a message to the user
}
```

You could call this function:

```js
if (!carto.isBrowserSupported()) {
  const reasons = carto.unsupportedBrowserReasons();
  // display a message to the user
}
```

VictorVelarde:
As per https://github.com/CartoDB/support/issues/1851, a clearer explanation on how to deal with CartoError would also be interesting (example / guide)

PR #1225, opened 2019-01-09
Add helpers to validate browser support

rochoa:
As per our previous conversation in https://github.com/CartoDB/carto-vl/issues/1149#issuecomment-441287468, I'm planning on adding `isBrowserSupported` and `unsupportedBrowserReasons` to be able to detect browser incompatibilities before creating any visualization.

This will make easier to display error messages to the end-user without having to use try/catch blocks.

### DIFF
--- a/examples/basics/browser-support.html
+++ b/examples/basics/browser-support.html
@@ -1,0 +1,74 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <title>Check for browser support | CARTO</title>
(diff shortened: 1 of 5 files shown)


rochoa:
Haha. Yes, it was WIP. I was finalizing some tests before making the publicly available.

VictorVelarde:
I think that's really nice. This is how Mapbox does a pretty similar stuff:
https://www.mapbox.com/mapbox-gl-js/example/check-for-support/

PR #1226, opened 2019-01-10
Versioning documentation

### LINKED ISSUE #1221: Clarify supported versions at Dev Center
Clarify the support for the different releases at Dev Center

rochoa:
This should close #1221.

### DIFF
--- a/docs/support/04-versions.md
+++ b/docs/support/04-versions.md
@@ -17,6 +17,8 @@ If you wanted to load a previous version, `v1.0.0` for instance, you could use t
  1. https://libs.cartocdn.com/carto-vl/v1.0/carto-vl.min.js.
  1. https://libs.cartocdn.com/carto-vl/v1.0.0/carto-vl.min.js.
 
+We also distribute CARTO VL through the [npm registry](https://www.npmjs.com/), the CARTO VL npm package is [@carto/carto-vl](https://www.npmjs.com/package/@carto/carto-vl).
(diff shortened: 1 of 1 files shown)


VictorVelarde:
IMO it's also interesting to link to https://www.npmjs.com/package/@carto/carto-vl

rochoa:
> IMO it's also interesting to link to https://www.npmjs.com/package/@carto/carto-vl

Just added that information. 👍

rochoa:
> LGTM, just a small comment, should we include a note about where to check for new releases at the github page or something like that?

@jsanz, that's a really good point. Apart from pointing to the Github releases page, what do you think if we listed the last three available/supported versions within this very same documentation page?

rochoa:
> > LGTM, just a small comment, should we include a note about where to check for new releases at the github page or something like that?
> 
> @jsanz, that's a really good point. Apart from pointing to the Github releases page, what do you think if we listed the last three available/supported versions within this very same documentation page?

After checking it, that's not an easy thing to accomplish on the developers' website. However, we plan to make the changelog available, see https://github.com/CartoDB/carto-vl/issues/1186, we should link to that once it's available.

jsanz:
> After checking it, that's not an easy thing to accomplish on the developers' website. However, we plan to make the changelog available, see #1186, we should link to that once it's available.

Nice, thanks for the update :heart:

Issue #1204, opened 2018-12-31
Improve editor


VictorVelarde:
We should improve the editor regarding to its:
- interface
- ux
- code experience

Algunenano:
One thing that would make my life easier is the option of not using basemaps in editor. Sometimes it's hard to compare different tiles o see certain features because of the noise introduced by the basemap / basemap tags.

Should I open a different issue?

VictorVelarde:
Good idea!. I think that could fit in this issue.

What do you think about a "None" option in the Basemaps, with a default white color?

elenatorro:
**Integrate an existing code editor for better interoperability**:

I think we should discuss also about adding a syntax highlighter, which I guess is one of the purposes of this task. I know @IagoLast did something in December, it'd be great to integrate these rules in the [Monaco Editor](https://github.com/Microsoft/monaco-editor), a tool we've already mentioned in the past, that can be very useful for the code editor.

PR #1229, opened 2019-01-11
Fix Layer.addTo

### LINKED ISSUE #1228: Editor - Error adding a layer to the map
There is a problem in Editor, currently using MGL **v0.52**, which manifests in the first load, as:

image

The error appears in the console, although the Layer is eventually added to the map successfully.

Note: This has been detected while preparing the release for v1.1, which has been aborted.

VictorVelarde:
## 1. Relevant facts
Some facts about this problem:
1. It manifests just with MGL v0.52 (not in v0.50 and v0.51)
1. It only has appeared in the *editor.html*, not in any of the examples (public / debug)
1. The editor defines dinamically the Map's style (`map.setStyle(selectedBasemap`) vs the examples defining it usually within map's constructor options, with something like `const map = new mapboxgl.Map({ style: carto.basemaps.positron ...`
1. The editor's code, regarding to map & layer loading is a bit convoluted when calling to `layer.addTo(map, "watername_ocean")`.
1. The common way to add layers to MGL map is:
    ```js
    map.on('load', () => {
         map.addLayer(yourLayer, beforeLayerID?);
    });
    ```
1. ... But in CARTO VL there is a convenient method to add a layer to the map, with just **Layer.addTo(map, beforeLayerID)**. This is its current implementation:
    ```js
    addTo (map, beforeLayerID) {
        const STYLE_ERROR_REGEX = /Style is not done loading/;

        const addTheLayer = () => { map.addLayer(this, beforeLayerID); };
        try {
            addTheLayer();
            // Note: map.isStyleLoaded() has been tested here without success
        } catch (error) {
            if (!STYLE_ERROR_REGEX.test(error)) {
                throw new CartoRuntimeError(`Error adding layer to map: ${error}`);
            }

            map.on('load', addTheLayer);
        }
    }
    ```
1. So as we can see, that method has a fallback inside, just in case 'the map' is not ready yet (still loading'), to wrap the more common `map.on('load', addTheLayer);`... and that looks exactly like the situation here, so...

## 2. What's happening?
*Why isn't that catch working?*
  The error raising from the first addLayer is not been caught. The try & catch doesn't work here, as the error is an **Event**, not a common thrown Error (see https://www.mapbox.com/mapbox-gl-js/api/#map.event:error). So there is no way to catch it like that, and that looks like the error in the console. Moreover, the error 'message' doesn't correspond in this case with current regex. 

So, has v0.52 changed the way to raise the errors here... ? Yes, it has in this particular case!. It looks like it was not evented before. See: https://github.com/mapbox/mapbox-gl-js/pull/7539/

There is also a new warning, which corresponds to these changes: https://github.com/mapbox/mapbox-gl-js/pull/7562 but it doesn't look harmful.

*And why is the layer eventually loading?*
  The editor code has two calls to addLayer, and the second one seems succesful, as the basemap is then ready. To clean that is something we should address in https://github.com/CartoDB/carto-vl/issues/1204, so it is out of scope here.

## 3. Fix
Make Layer.addTo able to deal with common error (as before) and evented error (new in MGL v0.52)


VictorVelarde:
Note: all failing test 🔝 were `test:user` failing tests. In all those commits, the corresponding `test:user:watchc` passed all tests succesfully. The "fix" has been to test for the spy call without any surrounding map or layer .on events

PR #1240, opened 2019-01-17
Fix MVT Interactivity for polygons

### LINKED ISSUE #1239: Interactivity fails on MVT with a polygon layer
As reported internally, this is a regression detected on v1.1.1

VictorVelarde:
### Closes #1239 

---

This error is related to the last changes made to Interactivity, to allow a better management of multiple feature pieces (polygons / lines) within `_getPartialFeaturesAtPosition`.

If using `carto.source.Dataset` (or `carto.source.SQL`), **idProperty** is `cartodb_id` by default.
When that happens, the internal id for a feature in each dataframe, for a particular index i, is:
```js
const currentId = dataframe.properties[this.idProperty][i]
// >> and that returns a perfectly common numeric identifier
```
In those cases, there is an equivalence between the 'internal id' and the 'featureId':
`dataframe.properties[this.idProperty][0] === dataframe.getFeature(0)[this.idProperty]`

But that equivalence is not followed in **carto.source.MVT**, where idProperty can be set to an arbitrary PK. 

In a feature coming from a `carto.source.MVT` one can get:
```js
  const currentId = dataframe.properties[this.idProperty][i];
  // >> returns, for example, 4237

  // while this...
  const currentFeatureId = dataframe.getFeature(i)[this.idProperty]
  // >> returns something very different, such as "44009990200"
```

This PR fixes the problem, favouring the second form, which is consistent among different kinds of data sources


test/integration/render/scenarios/regression/mvt-interactivity/scenario.js
```diff
@@ -0,0 +1,76 @@
...
+    strokeWidth: 1
+`);
+const layer = new carto.Layer('my_mvt_layer', source, viz);
+
+const interactivity = new carto.Interactivity(layer);
+
+interactivity.on('featureClick', async (event) => {
+    const aFeature = event.features[0];
+
+    console.log(aFeature.id);
```
rochoa: I'm gonna remove these lines.

VictorVelarde: The original problem was related to `id` properties, so that's why I left that complementary logs.

PR #1257, opened 2019-02-08
Validate idProperty types

### LINKED ISSUE #1256: idProperty should check type is valid
When using the `MVT` source, you can define what property to use for the ID of the features through the `idProperty` metadata option. However, if the property type is not specified in the metadata `properties` is supposed to be a `number`. If that's not the case and it's a `category` property, it will mess up any other category property indexed values.

The user can work-around the problem by defining the `category` type for the ID property.

OTOH, we should raise an error as we do for any other type mismatch.

rochoa:
This will fix #1256.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fix regression in `Interactivity` using MVT and polygons
-- Raise an error if when the idProperty is not specified in the metadata and is not a number
+- Raise an error when the idProperty is not specified in the metadata and is not a number
 - Fix getting aggregation expression values passed to `viewportFeatures` expression
 
 ## [1.1.1] - 2019-01-16
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Allow the use of variables for all classifiers: `globalQuantiles`, `globalEqIntervals`, `globalStandardDev`, `viewportQuantiles`, `viewportEqIntervals` and `viewportStandardDev`
+- Improve MVT decoding error messages
 
 ### Fixed
 - Fix regression in `Interactivity` using MVT and polygons
+- Raise an error if when the idProperty is not specified in the metadata and is not a number
 
 ## [1.1.1] - 2019-01-16
 
--- a/src/codecs/mvt/Category.js
+++ b/src/codecs/mvt/Category.js
@@ -1,16 +1,15 @@
 import CategoryCodec from '../Category';
-import CartoRuntimeError, { CartoRuntimeTypes as crt } from '../../errors/carto-runtime-error';
+import CartoRuntimeError, { CartoRuntimeTypes as runtimeErrors } from '../../errors/carto-runtime-error';
 
 export default class MVTCategoryCodec extends CategoryCodec {
     sourceToInternal (metadata, propertyValue, propertyName) {
         const propertyValueType = typeof propertyValue;
         if (propertyValue !== null && propertyValueType !== 'undefined' && propertyValueType !== 'string') {
-            // TODO: It would be nice to include the name of the source property in the error message.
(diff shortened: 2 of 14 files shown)


rochoa:
I was thinking about how we could improve the error messages in these scenarios.

The problem I see is that we tell the expected property `type`, the received `value` and the actual value `type`, but we don't say a word about what is the property we are having the issue with.

So instead of throwing `MVT decoding error. Metadata property is of type 'number' but the MVT tile contained a feature property of type 'string': 'id-0'`, we should say something like `MVT decoding error. The "country" property expected the type 'number' but the MVT tile contained the value "id-0" with type "string"`.

What do you think? This would change some error messages, something I don't like.

I would do that in a separate PR.

VictorVelarde:
How about just always requiring the user to specify the type when using a custom idProperty? 

We could then throw a new, much more clear, error (such as "missing type for custom idProperty", whatever...), without affecting the current messages for other situations.

rochoa:
I like your idea for this specific use-case, but the improvements I was proposing were for any type mismatch.

E.g., you have a metadata object with two properties, `city=category` and `province=category`, right now you can receive an error message like `MVT decoding error. Metadata property is of type 'category' but the MVT tile contained a feature property of type 'number': 13`. You won't be able to know if the problem is with the `city` or with the `province` property.

VictorVelarde:
It's clear that if the name of the property is not included in the error itslef, its value is quite lower, so I agree, we should improve them, even if that implies a break.

elenatorro:
Hi @VictorVelarde 👋 could you please re-review it? :)

src/codecs/mvt/Category.js
```diff
@@ -2,15 +2,15 @@ import CategoryCodec from '../Category';
 import CartoRuntimeError, { CartoRuntimeTypes as crt } from '../../errors/carto-runtime-error';
```
VictorVelarde: You can change the alias of the import to 'errors' for consistency with your changes in other files. 
I don't care much about the alias itself, but it's nicer if we are consistent with them all across the library.

test/unit/source/MVT.test.js
```diff
@@ -45,8 +45,8 @@ describe('sources/MVT', () => {
             });
             metadata.setCodecs();
             expect(() => {
-                metadata.codec('wadus').sourceToInternal(metadata, 'this is not a number');
-            }).toThrowError(/MVT decoding error. Metadata property is of type \'number\' but the MVT tile contained a feature property of type 'string': \'this is not a number\'/);
+                metadata.codec('wadus').sourceToInternal(metadata, 'this is not a number', 'wadus');
```
VictorVelarde: Nice test

src/codecs/mvt/Category.js
```diff
@@ -2,15 +2,15 @@ import CategoryCodec from '../Category';
...
 
 export default class MVTCategoryCodec extends CategoryCodec {
-    sourceToInternal (metadata, propertyValue) {
+    sourceToInternal (metadata, propertyValue, propertyName) {
         const propertyValueType = typeof propertyValue;
         if (propertyValue !== null && propertyValueType !== 'undefined' && propertyValueType !== 'string') {
             // TODO: It would be nice to include the name of the source property in the error message.
```
VictorVelarde: And now I think we could remove this comment... 😄

src/codecs/mvt/Category.js
elenatorro: You're right, I didn't realize 👍

CHANGELOG.md
```diff
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Allow the use of variables for all classifiers: `globalQuantiles`, `globalEqIntervals`, `globalStandardDev`, `viewportQuantiles`, `viewportEqIntervals` and `viewportStandardDev`
+- Improve MVT decoding error messages
 
 ### Fixed
 - Fix regression in `Interactivity` using MVT and polygons
+- Raise an error if when the idProperty is not specified in the metadata and is not a number
```
VictorVelarde: if or when?

elenatorro: 👀 when?

PR #1258, opened 2019-02-12
Reduce and clean-up examples section

makella:


makella:
closing so i can address when this cycle's examples are put together and merged into master. once that is done, I can reorganize/remove according to what's outlined in the ticket.

Issue #1244, opened 2019-01-25
Make aggregated values available for labels


ztephm:
Use case:
We want to make a cluster map. We can use our clusterCount function for that, for example:

```
const viz = new carto.Viz(`
  @c_count: clusterCount()
  
  color: red
  width: @c_count
  resolution: 16
`);
```

It's a common convention for cluster maps to be labeled with the number of features represented by the cluster marker:

![screen shot 2019-01-24 at 6 52 59 pm](https://user-images.githubusercontent.com/1779444/51716095-63a40980-2009-11e9-9b45-70ce6add1b5c.png)

However we can't use the clusterCount in a label.

Right now VL requires us to use an original dataset attribute for labels instead, like in this example:
https://carto.com/developers/carto-vl/examples/#example-point-labels

```
const viz = new carto.Viz(`
  @c_count: clusterCount()
  @v_features: viewportFeatures($poi_name)
  
  color: red
  width: @c_count
  resolution: 16
`);

....
const features = viz.variables.v_features.value;
...
"label_field": `${f.properties['poi_name']}`,

```

Can we create a way to use aggregated values for labels, so something like this would work?

```
const viz = new carto.Viz(`
        color: fuchsia
        width: 6
        strokeColor: white
        strokeWidth: 2
        filter: $major == "YES"
        @cc: clusterCount()
        @v_features: viewportFeatures($poi_name, @cc)
      `);
```

elenatorro:
Hi @ztephm 👋 I added two examples in this PR https://github.com/CartoDB/carto-vl/pull/1259, but I'd like to know if there's another dataset we can use in order to have a better one. Thanks! ✨ 

![screenshot 2019-02-13 at 10 56 27](https://user-images.githubusercontent.com/3824953/52704840-c9890e80-2f81-11e9-97e2-f9a3cf7b8c44.png)


cc @makella

ztephm:
Thanks, it looks great! 

Here's an open dataset of Montreal parking signs I used for a cluster map. Depending on the settings it can have clusters with numbers up to 3 digits—in the example it could be good to show clusters with 1-, 2- and 3-digit numbers:

[signalisation_rd_tt.zip](https://github.com/CartoDB/carto-vl/files/2862595/signalisation_rd_tt.zip)

[...]

PR #1259, opened 2019-02-13
Make aggregated values available for labels

elenatorro:
Related issue https://github.com/CartoDB/carto-vl/issues/1244

In this PR we're solving two issues at the same time. This approach adds a new property to the Base expression object: `variableName`. The problem is that when getting the value from an expression assigned to a variable, we use the `propertyName` of the expression, which returns the name of the property we pass as an argument to the expression. For example:

`@v_features: viewportFeatures($poi_name)`

Will get the `poi_name` value from the feature.

However, if do the following:

```
@poi: $poi_name
@v_features: viewportFeatures(@poi)
```

Then it fails, because you don't have access to the property through the variable name, although it should work as well. The same situation happens with expressions where we don't pass a property as an argument, as the `ClusterCount` expression:

```
@clusterCount: clusterCount()
@v_features: viewportFeatures(@clusterCount)
```

Because we get this value through `_cdb_feature_count` from the properties.

**ClusterCount example**
[examples/labeling/point-aggregation-cluster-count.html](https://github.com/CartoDB/carto-vl/blob/1244-make-aggregated-values-available-for-labels/examples/labeling/point-aggregation-cluster-count.html)

screenshot 2019-02-13 at 10 56 27

TODO:

- [x] Improve Cluster Count example
- [x] Fix specs
- [x] Add specs

### DIFF
--- a/examples/labeling/point-aggregation-cluster-count.html
+++ b/examples/labeling/point-aggregation-cluster-count.html
@@ -23,7 +23,7 @@
 <script>
     const map = new mapboxgl.Map({
         container: 'map',
-        style: 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json',
+        style: carto.basemaps.darkmatter,
         center: [2.1694138650888135, 41.37286330558891],
         zoom: 21.611075639862133
     });
@@ -44,7 +44,7 @@
         strokeWidth: 0
         resolution: 0.25
         width: @cc * 15
-        @v_features: viewportFeatures($speed, @cc);
+        @v_features: viewportFeatures(@cc);
     `);
     const layer = new carto.Layer('layer', source, viz);
 
--- a/examples/labeling/point-aggregation-cluster-count.html
+++ b/examples/labeling/point-aggregation-cluster-count.html
@@ -1,0 +1,104 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Exported map | CARTO VL</title>
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<script src="../../dist/carto-vl.js"></script>
+<script src="https://api.tiles.mapbox.com/mapbox-gl-js/v0.52.0/mapbox-gl.js"></script>
+<link href="https://api.tiles.mapbox.com/mapbox-gl-js/v0.52.0/mapbox-gl.css" rel="stylesheet" />
+<link rel="stylesheet" type="text/css" href="../style.css">
+<style>
+    html, body {
+        margin: 0;
+    }
+    #map {
+        position: absolute;
+        width: 100%;
+        height: 100%;
+    }
+</style>
+</head>
+<body>
+<div id="map"></div>
+<script>
+    const map = new mapboxgl.Map({
+        container: 'map',
+        style: 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json',
+        center: [2.1694138650888135, 41.37286330558891],
+        zoom: 21.611075639862133
+    });
+
+    carto.setDefaultConfig({
+        serverURL: 'https://{user}.carto.com'
+    });
+
+    carto.setDefaultAuth({
+        username: 'cartovl',
+        apiKey: 'default_public'
+    });
+
+
+    const source = new carto.source.SQL(`select *, 1 as co from gecat_geodata_copy`);
+    const viz = new carto.Viz(`
+        @cc: clusterCount()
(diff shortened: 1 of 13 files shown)


examples/labeling/point-aggregation-cluster-count.html
```diff
@@ -0,0 +1,104 @@
...
+        height: 100%;
+    }
+</style>
+</head>
+<body>
+<div id="map"></div>
+<script>
+    const map = new mapboxgl.Map({
+        container: 'map',
+        style: 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json',
```
VictorVelarde: We can now use carto.basemap enum for the style, eg. carto.basemaps.voyager

elenatorro: You're right! I've copy-pasted other example, so I'll change it there as well.

makella:
SUPER nice one!! 

Going to try this out!

makella:
@elenatorro this is a great start! awesome that we can get some of the aggregated values like this!

I did a little testing with this and I have some comments/questions.

- Is it possible to be "smart" about what values we grab for viewport aggregations at different zoom levels? 

For example, in this map: [debug/cluster-label-zoom-no-filter.html](https://github.com/CartoDB/carto-vl/blob/1244-make-aggregated-values-available-for-labels/debug/cluster-label-zoom-no-filter.html), would it be possible to put some thresholds where, for example, clustercount values of < 5 wouldn't get labeled at zoom 3 and have the ability to make those rules through zoom?

In this image, you can see that having low clustercount values labeling, really makes the map hard.

screen shot 2019-02-13 at 3 12 51 pm

Having something like this would also help with the cases where lower values are labeled, but higher ones are not (see for example the red circle between 10 and 18 is a higher value but not labeled... this is a Mapbox label placement thing but I'm wondering if we can be smart about the information returned to the centroid label layer)
screen shot 2019-02-13 at 3 15 07 pm

One idea, was to filter ClusterCount based on zoom level, but that doesn't work because it is actually filtering not just the labels, but the features too... 
- [debug/cluster-label-zoom-with-filter.html](https://github.com/CartoDB/carto-vl/blob/1244-make-aggregated-values-available-for-labels/debug/cluster-label-zoom-with-filter.html)

... still thinking about all of this! .... just some intial tests, definitely want to explore the possibilities in more detail!

elenatorro:
I've not found a way to improve the labels visualization :/  In this [example](https://docs.mapbox.com/mapbox-gl-js/example/cluster/) Mapbox uses the `cluster` parameter when adding a source, but it doesn't work for us. I've been checking also the filters [here](https://docs.mapbox.com/mapbox-gl-js/example/filter-markers-by-input/), but I didn't find a great solution.

VictorVelarde:
Just as a comment, if we have a custom function in which we can filter the 'viewportFeatures' (which results into 'centroids' for labelling), we could check for map.getZoom inside, and make some decisions there. So not in the viz expression itself, but in that JS part of the 'getCentroid > make Geojson' for labelling

src/renderer/viz/expressions/viewportFeatures.js
```diff
@@ -118,5 +121,6 @@ function _childrenFromProperties (properties) {
 }
 
 function validProperty (property) {
-    return property.isA(Property) || property.isA(ClusterAggregation) || property.isA(ClusterTimeDimension);
+    const validExpressions = [Property, ClusterAggregation, ClusterCount, ClusterTimeDimension];
```
VictorVelarde: 👍 nice!

src/renderer/viz/expressions/viewportFeatures.js
```diff
@@ -89,8 +89,11 @@ export default class ViewportFeatures extends BaseExpression {
             if (!this._requiredProperties.every(p => validProperty(p))) {
                 throw new CartoValidationError(`${cvt.INCORRECT_TYPE} viewportFeatures arguments can only be properties`);
             }
-            const columns = this._requiredProperties.map(p => p.propertyName);
-            this._FeatureProxy = genLightweightFeatureClass(columns, renderLayer);
+
+            const propertyNames = this._requiredProperties.map((p) => {
+                return { property: p.propertyName, variable: p._variableName };
+            });
+            this._FeatureProxy = genLightweightFeatureClass(propertyNames, renderLayer);
```
VictorVelarde: Good addition

makella:
I ❤️ this PR

Issue #1033, opened 2018-10-08
add globalHistogram()


davidmanzanares:
Only viewportHistogram exists, which is an inconsistency since the rest of widget related functions have both a viewport version and a global version

rochoa:
💥

davidmanzanares:
:heart:

PR #1262, opened 2019-02-14
Add GlobalHistogram expression

elenatorro:
Related issue: https://github.com/CartoDB/carto-vl/issues/1033

ToDo
- [x] Add GlobalHistogram expression
- [x] Example to test GlobalHistogram with category values
     * [x] `examples/misc/legends/legend-categorical-chart-global.html`
- [x] Example to test GlobalHistogram with numerical values
     * [x] `examples/widgets/global-vs-viewport-histogram.html`
- [x] Use both ViewportHistogram and GlobalHistogram with classifiers (i.e.: top)
- [x] Add / change specs
- [x] Clean & Improve examples --> TBD https://github.com/CartoDB/carto-vl/issues/1266
- [x] Improve 'Reference' documentation

### Demo 1: Global vs Viewport Numeric Histogram

viewport-vs-global-histogram

### Demo 2: Global Categorical Histogram

categorical-data

Notes:

1. For numeric values, we're using the `sample` values from the metadata, as we're doing in other global expressions such as `GlobalQuantiles`. As we've discussed, we'll talk about it during this cycle.

2. We should add somewhere in the guides how to use the `getJoinedValues` method when using any histogram expression, and how to use it combined with `getLegendData` from a ramp -->  https://github.com/CartoDB/carto-vl/issues/1267


elenatorro:
I still have to add tests, but meanwhile, we could add a good example for these new expression. The examples I've linked are only for manual testing, so we should provide better ones. I think we should add, at least, one that shows the difference between globalHistogram & viewportHistogram expressions. Feedback & ideas are welcome!

cc @makella

makella:
@elenatorro will take a closer look!

elenatorro:
After talking with @jgoizueta I'm going to follow a simpler approach to get the histogram data for numeric values, which is using the `sample` array from the `metadata` 👍

elenatorro:
There's an issue with `viewportHistogram` expression I'm working on now which I think should be fixed in this PR. When using `viewportHistogram` with an expression, the results are not consistent.

Let's say we have 6 features in the viewport: 3 features 'a', 2 features 'b' and 1 feature 'c'

* Using viewportHistogram with a categorical property `@histogram: viewportHistogram($category)` returns:

```js
[
   { x: 'a', y: 3 },
   { x: 'b', y: 2 },
   { x: 'c', y: 1 }
]
```

* Using viewportHistogram with a categorical property and a `top` expression to get the top 2 features: `@histogram: viewportHistogram(top($category, 2))`  returns:

```js
[
   { x: 0, y: 3}, // feature a
   { x: 1, y: 2}, // feature b
   { x: -1, y: 1 } // others
]
```

And we've tests validating this, so we should change them as well. What do you think?

PR #1274, opened 2019-02-22
Increase wait milliseconds

elenatorro:
Tests were failing both in master and locally. By increasing the wait time it seems to be fixed, but I don't really know why they were failing.


VictorVelarde:
It looks like browser tests work fine, although that pesky `1) /timerange` from e2e is annoying us again

PR #1277, opened 2019-02-23
Add cluster label examples

makella:
This PR:

- adds `clusterCount()` labeling example would like to review some of the process with team to decide final symbology
- modifies label styling for the conditional label example so it follows more recent GL styling specs
- adds title boxes to all labeling examples


makella:
@elenatorro is there a way that we can get rounded values back to the label centroids? I want to label using clusterAVG, and there are many significant digits for all values.

see: examples/labeling/cluster-labels-average.html

elenatorro:
Several (😅) notes from the examples:

* **Cluster avg values**:
> is there a way that we can get rounded values back to the label centroids?

Yes! We've to round the values directly when getting the labels from the viz variable: 

screenshot 2019-02-26 at 09 56 07

```js
"properties": {
    "label_field": Math.round(f.properties['ca']),
}
```

By looking at this example, and after reading deeply the mapbox documentation, I _think_ it's possible to change the label color depending on the cluster background color, just in case we wanted to add an example with that (although I think we shouldn't put much more effort explaining how their API works, it depends on the number of examples we want to have and maintain...)

* **About the 'string' & 'number' thing**:

As @rochoa said in the meeting, we're saving the label values as strings when using them in the Mapbox gl API, that's why the filter wasn't working when setting it as a `number`. By doing the following:

```js
"properties": {
    "label_field" f.properties['ca'], // without `${}`
}
```

And then, in the filter (which is what you had at the beginning):

```js
filter: ['>', ['number', ['get', 'label_field']], 1]
```

It works as expected ✨ 

* **About variable names**

We're using `ca` for `clusterAvg`, `cc` for `clusterCount`, etc. I think it'd be better if we use `v_clusterAvg`, for instance. I know it may seem longer, but it's more descriptive, and it's always a good pratice to have meaningful names in variables. It's just a suggestion, we can check how we're using variables in other examples.

* **About label size**

I tested also a way of increasing the font size depending on the cluster value (I think we talked about how could we do that, but I'm not sure):

```js
map.addLayer({
    // ...
    "layout": {
        // ...
         "text-size": ['*', ['number', ['get', 'label_field']], 0.35],
       // ...
    }
})
```
 
screenshot 2019-02-26 at 09 52 13

Ok, I agree the final result is not very beautiful 😅 , but maybe we can use it in a labelling guide or in an advanced example, what do you think?

makella:
@elenatorro awesome review!

Thank you so much for the rounding. As discussed, in the future, we might want to also consider ways to do things like take a value of 2467 and return it as 2.4K.

- [ ] To do: update clusterAVG example to rounded numbers and modify symbology

I will use number for the cluster label, and update the filter. That definitely makes more sense and will also "show" users that they can do custom buckets on the GL side if they want to style particular labels differently.

- [ ] To do: convert cluster count labels to number, update filter

Once we get the global calcs implemented:

- [ ] To do: update clusterCount example's width to global vs. viewport

Also, as we discussed, once we have global, we should definitely try some examples with all labelcluster maps!

elenatorro:
Hi @makella, as a result of [this](https://github.com/CartoDB/carto-vl/pull/1281#issuecomment-470048194), we won't be able to have this example :( :

_"To do: update clusterCount example's width to global vs. viewport"_

elenatorro:
Hi @makella 👋 just updated this branch with the latest fixes in variables + cluster issues, so we can keep working on this examples :)

makella:
great!

thank you.

makella:
@elenatorro ok, I went ahead and added the two examples to `examples.json` and double checked them one last time. basically, we have once clusterAVG example and one clusterCount() example for both labeling and symbology.

I think they should be ok to merge.

makella:
@elenatorro actually, i need to change something. sorry! will ping for review in a few minutes

makella:
@elenatorro ok, all ready for your review. thanks!

elenatorro:
I did a couple of small changes only for HTML and JS styling, but it LGTM 🌟 . I think we can merge it, let me know what do you think :) @makella

Issue #1172, opened 2018-11-30
Enhance filter examples/create guide on Dev Center


ramiroaznar:
`filter` is a property quite important when dealing with widgets. So more docs about this property are needed:
* a guide explaining the concept, how it works
* I was not able to get to the reference of this property (searching in Developers with `ctrol+f` is hard cc @inigomedina ),
* there is [one basic example](https://carto.com/developers/carto-vl/examples/#example-filter-based-styling), but another more complex one (or two) would be great,
* I like how easy to understand is [@rochoa's code](https://gist.github.com/rochoa/d61774b4086c5ade34555455284f9a3b#file-app-js-L40) in this respect.

It would be great to have also a mixed doc/example in Airship.

cc @makella @VictorVelarde

makella:
a first enhancement could be to make the filtering example more robust with a few different radio buttons for different kinds of filters in the examples section. We can also link to the reference from that example in the legend box text.

I will also spend some time thinking about a Guide that would be robust enough to demonstrate the use of filter in a variety of ways. Thinking here, that you can filter the data (SQL source) or filter the features in the browser (filter). We do a lot of explanation in the animation guide about filtering in that context, so it might be a matter of reworking content from there into a guide all about ways to filter data.

VictorVelarde:
Most of the issues / doubts came from filtering the current viz using some kind of widget in the client (mostly an airship category widget to choose one or several categories). Regarding to the 'SQL source', we had some conversations regarding to 're-using' the same source to get server updates and that could fit into the current 'sources guide' (maybe just as a Note)

makella:
ahhhh i see. 

Ok, so another layer of this then is maybe adding a section in the Widgets Guide to address this:
> Most of the issues / doubts came from filtering the current viz using some kind of widget in the client (mostly an airship category widget to choose one or several categories).

Or build off the category widget created in the Widgets Guide in a Filter with Widgets guide... (maybe?)

While still adding a few more meaningful options to the Filtering Example.

ramiroaznar:
Yes. Helping common use cases (filtering by histogram bins, categories...) would be very useful.

VictorVelarde:
In CARTO VL guides, I would try to illustrate how to update or reset a filter dinamically using simple controls (like an HTML Select with several options), to exemplify a simple 'selection updated' >> 'update filter' flow. 

[...]

PR #1279, opened 2019-02-25
Add filter example

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/1172

In this first approach, we're adding a widget that allows filter data by category using the `filter` expression and the JavaScript API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Allow the use of variables for all classifiers: `globalQuantiles`, `globalEqIntervals`, `globalStandardDev`, `viewportQuantiles`, `viewportEqIntervals` and `viewportStandardDev`
 - Improve MVT decoding error messages
+- Add interactive filter example
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


makella:
@elenatorro this is good!

I have a couple of comments/questions:

- can we remove the comments that were specific to this map as a legend example and add in a few comments for the pieces related to the filter and selection buttons?

- on line 85 we call out the others labels (I need to ask you about this in the original legend example as well) since there are no other categories showing, can we remove this?
```
        layer.on('updated', () => {
            if (!initialized) {
                const legend = layer.viz.color.getLegendData({
                    othersLabel: 'Other Weather'
                });
```

elenatorro:
> can we remove the comments...

Totally, actually the main idea is to take a final look to the code once we finish the example, clean it and keep the consistency with other examples 👌 

> others labels

Yes, we can remove this setting for sure, good catch!

makella:
@elenatorro thanks so much for all the comment clarifications! 

As discussed, our final to do here:
- [ ] change bars back to points (as to avoid confusion with a category-like widget since information is static)

makella:
@elenatorro ok, awesome! did a review and added some description in the legend box. LGTM. thanks for your work on this!

PR #1286, opened 2019-03-04
Fix  method and add regression test

elenatorro:
This PR:

- fixes `viz.toString` method
- uses a common constant for the supported viz properties
- uses resolution property as a _number expression_

### DIFF
--- a/src/Viz.js
+++ b/src/Viz.js
@@ -468,16 +468,18 @@ export default class Viz {
         }
     }
 
-    _checkVizSpec (vizSpec) {
-        // Apply implicit cast to numeric style properties
-        vizSpec.width = implicitCast(vizSpec.width);
-        vizSpec.strokeWidth = implicitCast(vizSpec.strokeWidth);
-        vizSpec.symbolPlacement = implicitCast(vizSpec.symbolPlacement);
-        vizSpec.transform = implicitCast(vizSpec.transform);
-        vizSpec.symbol = implicitCast(vizSpec.symbol);
+    _applyImplitCastToNumericProperties (vizSpec) {
         vizSpec.filter = implicitCast(vizSpec.filter);
         vizSpec.resolution = implicitCast(vizSpec.resolution);
+        vizSpec.strokeWidth = implicitCast(vizSpec.strokeWidth);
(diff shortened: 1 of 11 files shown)


src/Viz.js
```diff
@@ -523,7 +523,6 @@ export default class Viz {
             order: ${this.order.toString()}
             symbol: ${this.symbol.toString()}
             symbolPlacement: ${this.symbolPlacement.toString()}
-            offset: ${this.offset.toString()}
             ${variables}`.replace(/ {4}/g, '');
```
VictorVelarde: Besides removing offset, we should add the **transform** that was its replacement. Indeed I think we're also missing **resolution**. 

It would also be great if we could refactor a bit this, so we could use a proper array with these SUPPORTED_PROPERTIES (eg. from https://github.com/CartoDB/carto-vl/blob/0880f32303cce95220aaaf481d9dda7ea3035f94/src/Viz.js#L38)

elenatorro: This is a nice catch! I changed it accordingly and added more meaningful tests.

src/Viz.js
```diff
@@ -222,26 +204,26 @@ export default class Viz {
     _updateRootExpressionList () {
         this._rootExpressions = [
```
VictorVelarde: Nice refactor of the properties!
Could we bring also some of that into this method, so we can avoid also the repetition here?

src/Viz.js
```diff
@@ -512,18 +501,14 @@ export default class Viz {
...
-            filter: ${this.filter.toString()}
-            order: ${this.order.toString()}
-            symbol: ${this.symbol.toString()}
-            symbolPlacement: ${this.symbolPlacement.toString()}
-            ${variables}`.replace(/ {4}/g, '');
+        const variables = Object
+            .keys(this.variables)
+            .map(varName => `@${varName}: ${this.variables[varName].toString()}`);
+
+        const properties = SUPPORTED_VIZ_PROPERTIES
```
VictorVelarde: Nice! No forgotten properties anymore

elenatorro:
Merged 🚀

PR #1302, opened 2019-03-21
Revert path filter

elenatorro:
Context:
---

When publishing to `npm`, the `carto-vl.min.js` file was not updated to the registry. It seems to be a problem with `npm`, because, when running `npm pack`, the file is added correctly.

We've already sent this issue to npm support.

Meanwhile, we've decided to upload the `dist` file as we were doing in the past to avoid having this problem.


VictorVelarde:
As a related note: It looks like publishing the library to npm, with npm 6, misses the `carto-vl.min.js` file, even that file is included when using `npm pack`

Issue #1304, opened 2019-03-21
Built-in SVG Images are not loading on Firefox


elenatorro:
Context
---
If we load a map with our own icons in **Firefox**, it does not work. But if it's a url, then it works fine.
* Not working: [Example 1](https://cartodb.github.io/carto-vl/examples/styling/image-multiple.html)
* Working: [Example 2](https://cartodb.github.io/carto-vl/examples/styling/image-external.html)

elenatorro:
After some research, we may should change the way we encode svg files to be used in the canvas:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs

>  Security issues
    A number of security issues (e.g. phishing) have been associated with data URLs, and navigating to them in the browser's top level. To mitigate such issues, top-level navigation to data:// URIs has been blocked in Firefox 59+ (release version, Nightly/Beta from 58), and we hope to see other browsers follow suit soon. See Blocking Top-Level Navigations to data URLs for Firefox 58 for more details.

PR #1316, opened 2019-03-28
Fix Firefox icons [WIP]

elenatorro:
Fixes https://github.com/CartoDB/carto-vl/issues/1304

Context
---

🦊

- Our built-in icons use the `svg` expression underneath. This expression parses the svg to a `data-uri` format. 
- We need with / height properties in our svg's images.
- Loading the image to a canvas is a problem in Firefox  when the source is a `data-uri`. This is solved using an html `image`, not a `canvas`, in the texture:

```js
gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, this.image);
```

Screenshot 2019-03-28 at 11 05 33

Although this works now in Firefox, some render tests are failing because the resolution has changed now. I still have to figure out how to fix this, but a first review would be very helpful.

Useful links:

- https://webglfundamentals.org/webgl/lessons/webgl-render-to-texture.html
- https://delphic.me.uk/tutorials/webgl-text
- https://wdobbie.com/post/gpu-text-rendering-with-vector-textures/


src/renderer/viz/expressions/Image.js
```diff
@@ -33,11 +33,11 @@ export default class Image extends Base {
         this.type = 'image';
         this.canvas = null;
```
VictorVelarde: I think we should remove `this.canvas = null;`

src/renderer/viz/expressions/Image.js
```diff
@@ -80,43 +80,26 @@ export default class Image extends Base {
...
+        drawMetadata.freeTexUnit++;
+
+        if (this._loaded) {
             gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
             gl.bindTexture(gl.TEXTURE_2D, this.texture);
-            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, this.canvas);
-            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR_MIPMAP_LINEAR);
-            gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
+
+            this.image.width = this.image.width || 150;
```
VictorVelarde: In the old `_getCanvasFromImage` a default 256 value is used for canvas size (and it was just using the original `img.width` and `img.height`). Does it make sense to use here that 256 / 2 = 128 size ?

VictorVelarde:
The overall change looks fine to me. Could you post later on an example of that 'resolution change' that is breaking those render tests?

PR #1319, opened 2019-03-29
[fix] svg icons not working in firefox

elenatorro:
Fixes https://github.com/CartoDB/carto-vl/issues/1304

Context
---

🦊

- Our built-in icons use the `svg` expression underneath. This expression parses the svg to a `data-uri` format. 
- We need with / height properties in our svg's images.
- Loading the image to a canvas is a problem in Firefox  when the source is a `data-uri`. This is solved using an html `image`, not a `canvas`, in the texture:

```js
gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, this.image);
```

Screenshot 2019-03-28 at 11 05 33

Although this works now in Firefox, some render tests are failing because the resolution has changed now. I still have to figure out how to fix this, but a first review would be very helpful.

Useful links:

- https://webglfundamentals.org/webgl/lessons/webgl-render-to-texture.html
- https://delphic.me.uk/tutorials/webgl-text
- https://wdobbie.com/post/gpu-text-rendering-with-vector-textures/


VictorVelarde:
@elenatorro I guess this should be closed right?

Issue #1291, opened 2019-03-15
Let users customize the string 'CARTO_VL_OTHERS'


rjimenezda:
As a user, if I use a `top` expression, or similar, I would like to be in control of what the string will be for the 'rest' values.

Right now the workaround is applying a map to the resulting data (histogram, for instance) and replacing it there, but I think it would be nice to solve the problem at the root.

jorgesancha:
as discussed @rochoa

elenatorro:
I think we can close this because the use case was for Airship bridges and it's being solved here https://github.com/CartoDB/airship/pull/606. 

If in we need or want to add it later, we can reopen it.

Issue #1288, opened 2019-03-06
Review all values, eval and getLegendData methods


elenatorro:
## Context

There is a common issue that people usually struggles with: how to get data from **properties** and **variables**.

This is a quick answer from an issue comment:

---

### 1. When you need the value of any viz property, you have to use `.value`

Viz:
```
width: 5
```

Value:
`layer.viz.width.value`=> returns 5

### 2. When you assign an **expression** to a property and you want to now the value for certain feature, you have to use the `eval(feature)` method:

Viz:

```
width: sqtr($num)
```

Value:
`layer.viz.width.eval(feature)` => will return the sqtr value of the $num feature you've passed

_(note: you can get the feature, for instance, by using the `viewportFeature` expression, or in the event you get when adding interactivity)_

### 3. When the **expression** has a `getLegendData()` method

You can check if an expression has this method in our [Reference](https://carto.com/developers/carto-vl/reference/). It will return an array of objects with the data associated with each key, and it'd depend on the property type (numeric o categorical), the expression you used (ramp, buckets, top...).

Viz:
```
width: ramp(top($product, 3), [1, 2, 3])
```

Values:
`layer.viz.width.getLegendData()` => [{key: 'first top name', value: 1, key: 'second top name', value: 2, key: 'third top name', value: 3}]

---

However, the `value` is not actually consistent in all of the expressions and therefore can mislead the suer. I think we should:

- [x] Review all `value`, `eval()` methods, and other methods to get data such `getLegendData`
- [x] Add documentation for those methods
- ~Add a guide / post (I think a post is better) to explain those methods as above but with more detail and examples.~

---

Related issues:
* https://github.com/CartoDB/carto-vl/issues/1208

Issue #1208, opened 2019-01-02
getLegendData should return an empty array if there's no data


elenatorro:
Context
---
Using `getLegendData` method when there's no data returns an array with a length of `n`, but all the key values are `0`. 

<img width="479" alt="screen shot 2019-01-02 at 10 25 48" src="https://user-images.githubusercontent.com/3824953/50586266-11901e00-0e79-11e9-82dd-48379d898cf7.png">

I think the _data_ array should be empty in this case, when _min_ and _max_ values are the same (0 in this case). By taking a look to each `_getLegendData` private method, it **seems** to be happening only when using the `linear` expression, where by default it'll always return an array of length 10 (DEFAULT_SAMPLES).

https://github.com/CartoDB/carto-vl/blob/35bf5af37a0e4a04f79d04ddc950b419891aec5b/src/renderer/viz/expressions/linear.js#L209-L225

elenatorro:
Note: I don't know if if min and max parameters in the linear expressionare the same it's correct to expect this result:

[example](https://cartodb.github.io/carto-vl/examples/editor/#eyJhIjoic3BlbmRfZGF0YSIsImIiOiIiLCJjIjoiY2FydG92bCIsImQiOiJodHRwczovL3t1c2VyfS5jYXJ0by5jb20iLCJlIjoiY29sb3I6IHJhbXAobGluZWFyKCRhbW91bnQsIGdsb2JhbE1pbigkYW1vdW50KSwgZ2xvYmFsTWluKCRhbW91bnQpKSwgW2JsdWUsIHllbGxvd10pXG53aWR0aDogcmFtcChsaW5lYXIoJGFtb3VudCwgZ2xvYmFsTWluKCRhbW91bnQpLCBnbG9iYWxNaW4oJGFtb3VudCkpLCBbMTAwLCAxMF0pXG5vcmRlcjogZGVzYyh3aWR0aCgpKSIsImYiOnsibG5nIjoyLjE3LCJsYXQiOjQxLjM4fSwiZyI6MTMsImgiOiJEYXJrTWF0dGVyIiwiaSI6ImRhdGFzZXQifQ==) 

Screenshot:

![Screenshot 2019-05-24 at 16 02 08](https://user-images.githubusercontent.com/3824953/58333347-80c4d980-7e3d-11e9-9362-7d36878dcb36.png)

viz string:

```
color: ramp(linear($amount, globalMin($amount), globalMin($amount)), [blue, yellow])
[...]

PR #1323, opened 2019-04-01
Unify value methods

elenatorro:
Related issue https://github.com/CartoDB/carto-vl/issues/1288

- Unify `value` getter for all the expressions
- Remove `.expr` and use `value` instead
- Add `getLegendData` to base expression
- Add `values` getter (get the `data` from `getLegendData`). Also related: https://github.com/CartoDB/carto-vl/issues/1208
- Remove `eval()` when it should be `value` instead (`eval()` should be used to *evaluate* features)
- Add documentation about how to get values from expressions
- Allow users to define `othersLabel` in `top` and `buckets` expressions, and `samples` in `linear` expression, related with https://github.com/CartoDB/carto-vl/issues/1291


docs/guides/03-style-with-expressions.md
```diff
@@ -99,13 +99,85 @@ To access feature properties from a dataset, use the dollar sign (`$`). For exam
...
+Once a visualization has been created, you can access and modify all styling properties and variables through the returned `Visualization` object.
 
 Once you have a `Visualization` object, you can:
-- Get its property values by using the `.value` getter (`const currentColor = viz.color.value;`)
-- Modify its property values by using the `.blendTo()` method (`viz.color.blendTo(‘green’);`)
+
+
+#### Access
+
+There three main use cases:
```
VictorVelarde:
```suggestion
There are three main use cases:
```

src/renderer/viz/expressions/RampImage.js
```diff
@@ -43,9 +47,8 @@ export default class RampImage extends Base {
     }
 
     getLegendData (options) {
-        const config = Object.assign({}, DEFAULT_OPTIONS, options);
         const type = this.input.type;
-        const legendData = this.input.getLegendData(config);
+        const legendData = this.input.getLegendData(options);
```
VictorVelarde: Why have we omitted here the previous DEFAULT_OPTIONS?

src/renderer/viz/expressions/Zoomrange.js
```diff
@@ -26,24 +26,25 @@ import { implicitCast, checkType, checkExpression } from './utils';
  * @api
  */
 export default class Zoomrange extends BaseExpression {
-    constructor (zoomBreakpointList) {
-        zoomBreakpointList = implicitCast(zoomBreakpointList);
+    constructor (_zoomBreakpointList) {
```
VictorVelarde: Shouldn't the ctor param be named without underscore?

src/renderer/viz/expressions/base.js
```diff
@@ -65,19 +65,30 @@ export default class Base {
...
+
+    /**
+     * Get the expression values
+     *
+     * @api
+     * @memberof Expression
+     * @name values
+     */
+    get values () {
+        return this.getLegendData().data;
```
VictorVelarde: Does this always guarantee a result? Should we consider a runtime error + a proper message?

src/renderer/viz/expressions/basic/constant.js
```diff
@@ -29,17 +29,24 @@ export default class Constant extends BaseExpression {
         checkNumber('constant', 'x', 0, x);
 
         super({});
-        this.expr = x;
+        this.value = x;
```
VictorVelarde: Are we missing here the underscore?
```suggestion
        this._value = x;
```

src/renderer/viz/expressions/linear.js
```diff
@@ -3,7 +3,8 @@ import { checkExpression, implicitCast, checkType, checkMaxArguments, clamp } fr
 import { globalMin, globalMax } from '../expressions';
 import { castTimeRange, msToDate } from '../../../utils/util';
 import IdentityCodec from '../../../codecs/Identity';
-import TZDate from '../../../utils/time/TZDate';
```
VictorVelarde: I find the new name much clearer, but I guess that we should consider this a breaking change

src/renderer/viz/expressions/linear.js
```diff
@@ -30,7 +31,7 @@ import TZDate from '../../../utils/time/TZDate';
 * @api
 */
 export default class Linear extends BaseExpression {
-    constructor (input, min, max, range) {
+    constructor (input, min, max, range, samples = DEFAULT_SAMPLES) {
```
VictorVelarde: What is the use case for introducing here this optional param?

src/renderer/viz/expressions/linear.js
elenatorro: It's not an expression, why should it be then a breaking change?

src/renderer/viz/expressions/RampImage.js
elenatorro: By default, the default options (`othersLabel` and `samples`) are the same, but they're defined in the constructor of the nested expression. I forgot to added in the PR description (now it's added), but this comes from this issue: https://github.com/CartoDB/carto-vl/issues/1291.

src/renderer/viz/expressions/Zoomrange.js
elenatorro: You're right, this is a mistake when replacing the string.

src/renderer/viz/expressions/linear.js
elenatorro: Same as here: https://github.com/CartoDB/carto-vl/pull/1323#discussion_r275337504

src/renderer/viz/expressions/linear.js
VictorVelarde: All right then, I thought it was exposed somehow.

VictorVelarde:
Overall, it LGTM but I would feel more comfortable if we would fix the e2e docker problem first and bring it here, before we merge this in

rjimenezda:
Some (undocumented) part of the bridge relies on the fact that getLegendData was only present on certain expressions, but I can refactor it to use something else, like check the actual return value now, or the expression type.

PR #1328, opened 2019-04-10
Pg11 test support

ethervoid:
Added support to test against PG10 and PG11


ethervoid:
@elenatorro I have a failure in one test due to timeout, is this normal, maybe a rebuild could make it work?

ethervoid:
Working. Now it test against PG10 and PG11

@Algunenano @javitonino

ethervoid:
@Algunenano on it!

PR #1329, opened 2019-04-15
Fix built in svgs firefox

elenatorro:
This PR fixes https://github.com/CartoDB/carto-vl/issues/1304 and it's related with https://github.com/CartoDB/carto-vl/pull/1316

Context
===

The main issue is that in Firefox 🦊, local svgs aren't loaded in a canvas. As explained in the other PR:

* Our built-in icons use the svg expression underneath. This expression parses the svg to a data-uri format.
* We need with / height properties in our svg's images.
* Loading the image to a canvas is a problem in Firefox when the source is a data-uri

In this solution, we get our default icons from the CDN and use the `Image` expression.


src/renderer/viz/expressions/SVG.js
```diff
@@ -4,6 +4,7 @@ import { checkMaxArguments } from './utils';
 export default class SVG extends Image {
     constructor (svg) {
         checkMaxArguments(arguments, 1, 'svg');
+        // This doesn't work in Firefox for local SVG images
```
VictorVelarde: This is fine for our internal reference, but we should probably replicate this note in public jsdoc

elenatorro: This expression is not documented, we're using it internally. If we want to make it public in the future we can, taking into account that it will imply a minor release.

Issue #1309, opened 2019-03-25
Improve JavaScript error messaging


elenatorro:
* Related issues: https://github.com/CartoDB/carto-vl/issues/743
* Related project: CARTOFrames

Context
---

In order to provide a better error feedback, specially for CARTOFrames, we need to test how errors are sent, identify how can we improve them and then implement the required changes.

jorgesancha:
@kukukaka Could we have a chat about this one? Also related to: https://github.com/CartoDB/cartoframes/issues/599

elenatorro:
🚀

PR #1331, opened 2019-04-17
Add types to errors

elenatorro:
Related with https://github.com/CartoDB/carto-vl/issues/1309 and https://github.com/CartoDB/carto-vl/issues/743

This PR adds a `type` parameter to our custom errors in order to display better errors from external applications.

This would be a simple example of displaying the errors in HTML:

```html
<section class="error-section">
    <span id="error-name"></span>
    <span id="error-type"></span>
    <span id="error-message"></span>
</section>
```
```js
try {
    init()
} catch (e) {
    const error$ = document.getElementById('error-container');
    const errors$ = error$.getElementsByTagName('span');

    errors$[0].innerHTML = e.name;
    errors$[1].innerHTML = e.type;
    errors$[2].innerHTML = e.message;
}
```

### DIFF
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -4,7 +4,7 @@ import Viz from './Viz';
 import SourceBase from './sources/Base';
 import Renderer from './renderer/Renderer';
 import RenderLayer from './renderer/RenderLayer';
-import CartoValidationError, { CartoValidationTypes as cvt } from '../src/errors/carto-validation-error';
+import CartoValidationError, { CartoValidationTypes } from '../src/errors/carto-validation-error';
 import CartoRuntimeError from '../src/errors/carto-runtime-error';
 
 import { cubic } from './renderer/viz/expressions';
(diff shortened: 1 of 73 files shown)


VictorVelarde:
Could you please paste here a small example of the try / catch you're using this for?

elenatorro:
This would be a simple example of displaying the errors in HTML:

```html
<section class="error-section">
    <span id="error-name"></span>
    <span id="error-type"></span>
    <span id="error-message"></span>
</section>
```
```js
try {
    init()
} catch (e) {
    const error$ = document.getElementById('error-container');
    const errors$ = error$.getElementsByTagName('span');

    errors$[0].innerHTML = e.name;
    errors$[1].innerHTML = e.type;
    errors$[2].innerHTML = e.message;
}
```

src/Layer.js
```diff
@@ -558,42 +558,42 @@ export default class Layer {
 
     _checkId (id) {
         if (id === undefined) {
-            throw new CartoValidationError('\'id\'', cvt.MISSING_REQUIRED);
+            throw new CartoValidationError('\'id\'', CartoValidationTypes.MISSING_REQUIRED);
```
VictorVelarde: Hi @elenatorro I see the value in separating the error type in a new field, apart from the message. That's a good addition.

But I'm not sure about 2 things:
- error type names (and this was there before this PR)
- not using an alias

Regarding the first, I think we could use the format *BaseNameErrorTypes* (so `CartoValidationErrorTypes` instead of `CartoValidationTypes, `CartoMapsAPIErrorTypes` instead of just `CartoMapsAPITypes`... and so on).

About the alias, don't you think it is more practical to keep them? using `throw new CartoValidationError('\'id\'', cve.MISSING_REQUIRED);` instead of `throw new CartoValidationError('\'id\'', CartoValidationTypes.MISSING_REQUIRED);` I see this as quite similar to the 'exp' alias we use extensively.

elenatorro: I thought the same about adding the `Error` in the constant name, I totally agree with that!

About the alias, I did it to improve readability because `cvt` or `exp` don't mean anything, having long meaningful names is always better than having abbreviations:

https://readlearncode.com/code-and-stuff/clean-code-variables/

> Use words that can be pronounced and avoid abbreviations
There are many common abbreviations that are well known by everyone, but many that are not. Avoid using any abbreviations, for example, its better to use clientDateOfBirth rather that cDOB. Imagine that you must talk about your code in a meeting with people who are not familiar with what your code does. It’s much clearer to say “clientDateOfBirth” than “cDOB” every time you want to refer to that variable.

elenatorro:
I think that right now it's good enough to the error types in the reference. If we add a guide, it'd be hard to maintain, and error handling is something that has to be handled from a

elenatorro:
I think that right now it's good enough to have the error types in the reference. If we'd add a guide, in mi opinion it'd be hard to maintain, and error handling is something that's usually handled from a developer perspective, so I don't think we need to explain "How to handle errors".

Maybe we could add a link to the reference from one of the guides, how does it sound?

PR #1339, opened 2019-05-02
Enable macOS tests after saucelabs fix

### LINKED ISSUE #1272: Fix SauceLabs tests to run on macOS

VictorVelarde:
Closes #1272 
saucelabs has notified us their bug is fixed

### DIFF
--- a/test/browsers/browsers-target.js
+++ b/test/browsers/browsers-target.js
@@ -2,12 +2,11 @@
 const browsers = [
     { 'platform': 'Windows 10', 'browserName': 'chrome', 'version': 'latest' },
     { 'platform': 'Windows 10', 'browserName': 'firefox', 'version': 'latest' },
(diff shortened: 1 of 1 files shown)


VictorVelarde:
But it looks like it doesn't fix our case... 
Reporting them back!

VictorVelarde:
- They have confirmed us it doesn't work in our use case

alejandraarri:
They finally gave us a a fix!

>There may be a workaround for the 500 error issue! Try this:

>Add an entry in the hosts file on the machine running SC (and the localhost app) for a dummy hostname and 127.0.0.1, e.g.
```127.0.0.1     localtestsite```
and then test using localtestsite:9001 rather than localhost:9001

Tried a manual Live test and running our automatic tests and it worked in both cases.

PR #1349, opened 2019-05-21
Remove layer from interactivity when layer is removed

jesusbotella:
This PR solves a bug caused by interactivity with multiple layers whenever any of those layers is removed.

The bug is present whenever a layer with features is removed from the map and you move the mouse over any of the features that were previously there.

Interactivity calculates diff points using the layers within its layers' list, and as the layer removed is not actually removed from interactivity, it still gets points from that removed layer.

featureghost

So this PR:
- Fires `removed` event for layer objects whenever they're removed.
- Listens to that event in Interactivity and removes layer from Interactivity layers' list.
- If layers in interactivity are empty, removes listeners for `mousemove` and `click` in map instance.

### DIFF
--- a/src/interactivity/Interactivity.js
+++ b/src/interactivity/Interactivity.js
@@ -128,6 +128,7 @@ export default class Interactivity {
         this._prevHoverFeatures = [];
         this._prevClickFeatures = [];
         this._numListeners = {};
+        this._isAutoChangePointerEnabled = options.autoChangePointer;
 
         const allLayersReadyPromises = layerList.map(layer => layer._context);
         return Promise.all(allLayersReadyPromises)
@@ -135,8 +136,8 @@ export default class Interactivity {
                 postCheckLayerList(layerList);
                 this._subscribeToLayerEvents(layerList);
                 this._subscribeToMapEvents(layerList[0].map);
-            }).then(() => {
-                if (options.autoChangePointer) {
+
+                if (this._isAutoChangePointerEnabled) {
                     this._setInteractiveCursor();
                 }
             });
@@ -164,7 +165,7 @@ export default class Interactivity {
         this._onMouseMoveBound = this._onMouseMove.bind(this);
         map.on('mousemove', this._onMouseMoveBound);
 
-        this._onClickBound = this._onMouseMove.bind(this);
+        this._onClickBound = this._onClick.bind(this);
         map.on('click', this._onClickBound);
 
         this._disableWhileMovingMap(map);
@@ -204,7 +205,10 @@ export default class Interactivity {
         const map = this._layerList[0].map;
 
         this._removeLayerFromInteractivity(layer);
-        this._onFeatureHoverBound({ features: [] });
+
+        if (this._isAutoChangePointerEnabled) {
+            this._onFeatureHoverBound({ features: [] });
+        }
 
         if (!this._layerList.length) {
             this._unsubscribeToMapEvents(map);
--- a/src/interactivity/Interactivity.js
+++ b/src/interactivity/Interactivity.js
@@ -147,14 +147,17 @@ export default class Interactivity {
         if (!map.__carto_interactivities) {
             map.__carto_interactivities = new Set();
         }
-        this.on('featureHover', event => {
-            if (event.features.length) {
-                map.__carto_interactivities.add(this);
-            } else {
-                map.__carto_interactivities.delete(this);
-            }
-            map.getCanvas().style.cursor = (map.__carto_interactivities.size > 0) ? 'pointer' : '';
-        });
+        this._onFeatureHoverBound = this._onFeatureHover.bind(this, map);
+        this.on('featureHover', this._onFeatureHoverBound);
+    }
+
+    _onFeatureHover (map, event) {
+        if (event.features.length) {
+            map.__carto_interactivities.add(this);
+        } else {
+            map.__carto_interactivities.delete(this);
+        }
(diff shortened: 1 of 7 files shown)


jesusbotella:
The PR is ready for review, but I have two main concerns for the reviewer to check:

- `.then` remove in `_init`.
  I thought of removing it because there isn't any reason to put it off to a different execution as we're not returning any promise to wait for in the `then` execution after the layers initialization.

- Latest `_onFeatureHover` execution before event unsubscription.
  I had to do that last execution of the method because if the mouse is over any of the features before the layer removal, the pointer won't be changed to the initial state. I don't know if you feel like putting it there or doing anything else.

This is a huge performance boost for our lasso tool. :)

src/interactivity/Interactivity.js
```diff
@@ -147,22 +148,34 @@ export default class Interactivity {
...
         }
-        this.on('featureHover', event => {
-            if (event.features.length) {
-                map.__carto_interactivities.add(this);
-            } else {
-                map.__carto_interactivities.delete(this);
-            }
-            map.getCanvas().style.cursor = (map.__carto_interactivities.size > 0) ? 'pointer' : '';
-        });
+        this._onFeatureHoverBound = this._onFeatureHover.bind(this, map);
```
elenatorro: I think we should avoid declaring the function we're going to use to listen the event, which is used in `_onLayerRemoved` where we're not checking if it's defined.

Instead, I would save the `map` in a "private" value for the Interactivity (`this._map = layerList[0].map`) and use it on `_onFeatureHover` method instead of passing it as a parameter, so we could use this function in `_onLayerRemoved` method. We wouldn't need then `_onFeatureHoverBound`.

src/interactivity/Interactivity.js
```diff
@@ -180,13 +193,38 @@ export default class Interactivity {
...
+            layer.on('removed', this._onLayerRemoved.bind(this));
         });
     }
 
     _onLayerUpdated () {
         this._onMouseMove(this._mouseEvent, true);
     }
 
+    _onLayerRemoved (layer) {
+        const map = this._layerList[0].map;
```
elenatorro: We could use `this._map` here (from the previous comment)

src/interactivity/Interactivity.js
jesusbotella: Yes, I thought it too when doing my own code review and it's better to change it like you said. We're checking if the auto cursor option is enabled, but it's better to hold the map in a private variable :)

elenatorro:
Nice! ✨ We'd need to add it to the CHANGELOG.md and we'd be ready to merge it 👍

src/interactivity/Interactivity.js
```diff
@@ -128,41 +128,54 @@ export default class Interactivity {
...
+        this._onMouseMoveBound = this._onMouseMove.bind(this);
+        map.on('mousemove', this._onMouseMoveBound);
+
+        this._onClickBound = this._onClick.bind(this);
+        map.on('click', this._onClickBound);
+
         this._disableWhileMovingMap(map);
     }
 
+    _unsubscribeToMapEvents (map) {
```
VictorVelarde: maybe a rename to `_unsubscribeFromMapEvents` here?

test/integration/user/test/interactivity.test.js
```diff
@@ -455,6 +455,101 @@ describe('Interactivity', () => {
...
+            layer2._contextInitialize();
+
+            waitForLayersInitialization
+                .then(() => {
+                    const layerListLength = interactivity._layerList.length;
+
+                    layer1.remove();
+
+                    expect(interactivity._layerList.length).toBe(layerListLength - 1);
+                    expect(interactivity._layerList.indexOf(layer1)).toBe(-1);
```
VictorVelarde: Nice test

test/integration/user/test/interactivity.test.js
```diff
@@ -455,6 +455,101 @@ describe('Interactivity', () => {
...
+        });
+
+        describe('and it is the last layer with interactivity', () => {
+            it('should unsubscribe from map events', done => {
+                const waitForLayersInitialization = Promise.all([
+                    layer1._context,
+                    layer2._context
+                ]);
+
+                spyOn(interactivity, '_unsubscribeToMapEvents').and.callThrough();
```
VictorVelarde: `unsubscribeFromMapEvents`?

src/interactivity/Interactivity.js
jesusbotella: I had the same doubt the other day, and I searched for it 😂 
StackExchange says this: https://english.stackexchange.com/questions/102994/unsubscribe-to-vs-unsubscribe-from. (Take into account that sanctioned is a false friend here).

So, do you prefer `_unsubscribeFromMapEvents`?

VictorVelarde: My ear was used to 'from', but it's a good thing to know that 'to' works fine too. So let's keep that as it is (and keep learning good ol' english!) 😄

test/integration/user/test/interactivity.test.js
VictorVelarde: forget it 😄

VictorVelarde:
LGTM. Don't forget to add you as author at README.md 💥

jesusbotella:
That was my ultimate goal, to be listed in package contributors 😂

elenatorro:
VAMOS 🚀

PR #1351, opened 2019-05-22
Add full screen control to all examples

rjimenezda:
Fix https://github.com/CartoDB/developers/issues/585

Add mapbox's full screen control to all public examples.

We'll need to release a new version to update the dev center, so we should decide whether it's worth it just for this or if we wait until there's more fixes on the way.


VictorVelarde:
I think this should have been merged to `develop` not master, unless it is part of a quick fix to be released ASAP, right? @elenatorro @rjimenezda

rjimenezda:
We're making a patch, I'm just having some issues with the release process on my linux machine 👀

elenatorro:
Yes, as @rjimenezda says, we're releasing a patch. This is one of the things we decided to consider as "quick fix".

Issue #1354, opened 2019-05-23
CARTO_VL_OTHERS is always shown


Jesus89:
It should be included only when "others" exist.

Jesus89:
Closed in #1355

PR #1355, opened 2019-05-24
Do not always add CARTO_VL_OTHERS when getting legend data

elenatorro:
Fixes https://github.com/CartoDB/carto-vl/issues/1354

### DIFF
--- a/src/renderer/viz/expressions/buckets.js
+++ b/src/renderer/viz/expressions/buckets.js
@@ -150,6 +150,13 @@ export default class Buckets extends BaseExpression {
             }
         });
 
+        if (this.input.type === 'category') {
+            const property = metadata.properties[this.input.propertyName];
+            this._numDatasetCategories = property
+                ? property.categories.length
+                : this.list.elems.length - 1;
+        }
+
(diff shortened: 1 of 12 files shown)


src/renderer/viz/expressions/buckets.js
```diff
@@ -150,6 +150,13 @@ export default class Buckets extends BaseExpression {
             }
         });
 
+        if (this.input.type === 'category') {
+            const property = metadata.properties[this.input.propertyName];
+            this._numDatasetCategories = property
+                ? property.categories.length
+                : this.list.elems.length - 1;
```
Jesus89: When is this else condition reached?

elenatorro: This is an specific use case where the buckets input doesn't depend on a feature, and you can use a string as an input. This is used for getting a value. We've tests with this use case in the `ramp.test.js` file, for example:

```js
ramp(buckets('Madrid', ['Pontevedra', 'Zaragoza', 'Cordoba', 'Alicante', 'Murcia']), palettes.PRISM, red);
```

Issue #1359, opened 2019-05-27
Improve getLegendData with opacity example


VictorVelarde:
After per https://github.com/CartoDB/carto-vl/pull/1357 we have now a method to obtain colors inside an opacity expression, so we should modify this example to show it: https://carto.com/developers/carto-vl/examples/#example-feature-and-legend-opacity

PR #1363, opened 2019-05-31
Update legend opacity example

Jesus89:
Fixes https://github.com/CartoDB/carto-vl/issues/1359


VictorVelarde:
I missed the guide, there is a replica of this inside the legends guide. I'll make a PR

Jesus89:
Good catch!

PR #1365, opened 2019-05-31
Update guide with opacity & getLegendData improv.

VictorVelarde:
The same changes made for #1359 but applied to the Legends guide

### DIFF
--- a/docs/guides/10-add-legends.md
+++ b/docs/guides/10-add-legends.md
@@ -4,7 +4,7 @@ Maps that symbolize data without the necessary information to decode the symbols
 
 ### Overview
 
-CARTO VL itself doesn't provide the functionality to _draw_ legends. Instead, it provides the functionality necessary to _build_ them. What this means is that CARTO VL provides the data you need to create a legend, but drawing that data on the screen (in the form of a legend), is the responsibility of the application developer. The benefit of this is that you have more control over customizing legends for the needs of your specific application.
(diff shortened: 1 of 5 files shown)


docs/guides/10-add-legends.md
```diff
@@ -223,9 +221,11 @@ With that change, the map labels other categories as "Other weather" in the lege
 </div>
 You can explore this step [here](/developers/carto-vl/examples/maps/guides/add-legends/step-4.html)
 
-### Assign opacity with variables
+### Working with opacity
+
+To assign opacity to your mapped features, you will usually work with the `opacity` expression, combined with your `ramp`.
 
-If you want to assign opacity to mapped features, you will construct your `ramp` using [variables](/developers/carto-vl/guides/Glossary/#variables).
+If you want, for this example, you can try using [variables](/developers/carto-vl/guides/Glossary/#variables) for both expressions.
```
Jesus89: I would remove this line

examples/guides/add-legends/step-5.html
```diff
@@ -46,12 +46,12 @@ <h1 id="light">Rail Accidents by Weather</h1>
 
         const source = new carto.source.Dataset("railroad_accidents");
         const viz = new carto.Viz(`
-            @myRamp: ramp($weather,[darkorange,darkviolet,darkturquoise])
+            @myRamp: ramp($weather, [darkorange, darkviolet, darkturquoise])
```
Jesus89: I would use here a variable for the palette and then the ramp and opacity in the color expression

VictorVelarde: Yeah, that sounds as a better approach

Issue #1366, opened 2019-05-31
Bug in animation speed


Jesus89:
We have seen that in this example `/debug/animation/multi-anim.html` the animation interval is broken. This bug has appeared in 591530b47877b3f2df5ca9438427c600428724cf, so we need to add also a test to cover this case.

Jesus89:
Fixed in https://github.com/CartoDB/carto-vl/pull/1367

PR #1367, opened 2019-05-31
Fix windshaft filtering

Jesus89:
Fixes https://github.com/CartoDB/carto-vl/issues/1366

### DIFF
--- a/test/unit/client/windshaft.test.js
+++ b/test/unit/client/windshaft.test.js
@@ -1,3 +1,138 @@
+import Windshaft from '../../../src/client/windshaft';
+import Viz from '../../../src/Viz';
+
+const base = {
+    _apiKey: '1234',
+    _username: 'wadus',
+    _serverURL: 'https://wadus.carto.com',
+    _getFromClause: () => 'table_name'
+};
+
+const vizMock = `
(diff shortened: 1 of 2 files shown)


elenatorro:
Good catch! 💪

test/unit/client/windshaft.test.js
```diff
@@ -1,3 +1,138 @@
...
+    describe('_getServerInfoFrom', () => {
+        it('should convert viz into vizInfo', () => {
+            const viz = new Viz(vizMock);
+            const vizInfo = client._getServerInfoFrom(viz);
+            expect(vizInfo).toEqual(vizInfoMock);
+        });
+    });
+
+    describe('_needToInstantiateMap', () => {
+        it('should detect that it requires instantiation', () => {
```
VictorVelarde: It would be nice to test again, so we see no need for a new instantiation

VictorVelarde: And maybe a third change, altering one of the relevant properties. So it needs a new inst. again... Something so understand better the implement. from the test

VictorVelarde:
Great @Jesus89 	I think we can 🚀

Issue #1348, opened 2019-05-21
Change eval methods for those expressions that may depend in multiple features


elenatorro:
Context
===

`.eval` methods are used internally to evaluate a feature from an expression, for example, to get the "legend data" (through `.getLegendData`). Not all the methods need to evaluate a feature (for example, `globalAggregation`). However, there are some cases where there could be more than one feature to evaluate. For instance, in all the **binary expressions** (`mul`, `add`, `sub`...) evaluate **two** features.

To Do
===

We have to fix all the eval methods that require in two or more features to be evaluated. This will also imply to test those methods to build legends.

elenatorro:
Note: it currently has low priority but this is a must if we want to include bivariate legends in CARTOframes.

PR #1376, opened 2019-06-24
Change eval methods for multiple feature

elenatorro:
Related issue #1348 

This PR adds:
- Refactor binary operation expressions
- Fix `value` and `eval` for binary operation expressions
- Add `getLegendData` for binary operation expressions

This means that we can also have Bivariate legends in Airship and CARTOframes.

Screenshot 2019-06-24 at 16 51 07

Nice to have:
- Better and public example of Bivariate legend (different task)


debug/advanced/bivariate-legends.html
```diff
@@ -0,0 +1,107 @@
...
+    layer.addTo(map);
+
+    layer.on('loaded', () => {
+      hideLoader();
+
+      const legends = layer.viz.color.getLegendData();
+      let content = '';
+
+      legends.data.forEach((legend) => {
+        const colorHex = rgbToHex(legend.value);
```
VictorVelarde: @elenatorro I think we can now remove this `rgbToHex` thing, and just usa `rgba` as in `feature-and-legend-opacity.html` example

VictorVelarde: I would probably move this 'debug' example to public ones, what do you think? Or maybe we should do a small blog post?

src/client/windshaft-filtering.js
```diff
@@ -1,4 +1,11 @@
-import { And, Or, Equals, NotEquals, LessThan, LessThanOrEqualTo, GreaterThan, GreaterThanOrEqualTo } from '../renderer/viz/expressions/binary';
+import GreaterThan from '../renderer/viz/expressions/binary/GreaterThan';
```
VictorVelarde: Why this small refactor here? Was it maybe based on some tooling?

VictorVelarde: Oh, I see it now... you deleted the binary.js

test/unit/renderer/viz/expressions/binary.test.js
```diff
@@ -75,134 +76,496 @@ describe('src/renderer/viz/expressions/binary', () => {
...
+            describe('NUMBERS_TO_NUMBER', () => {
+                testValue('mul', 0, 0, 0);
+                testValue('mul', 1, 0, 0);
+                testValue('mul', 1, 1, 1);
+                testValue('mul', 1, 2, 2);
+                testValue('mul', -1, 2, -2);
+            });
+
+            describe('NUMBER_AND_COLOR_TO_COLOR', () => {
+                testValue('mul', 10, s.rgba(255, 15, 12, 1), { r: 10, g: 1, b: 0, a: 1 });
```
VictorVelarde: When does this NUMBER * COLOR is useful? do you know any example?

test/unit/renderer/viz/expressions/binary.test.js
```diff
@@ -75,134 +76,496 @@ describe('src/renderer/viz/expressions/binary', () => {
...
-            test('and', 0.5, 0.5, 0.25);
+    describe('.value', () => {
+        function testValue (fn, param1, param2, expected) {
+            it(`${fn}(${param1}, ${param2}) should return ${expected}`, () => {
+                const actual = s[fn](param1, param2).value;
+                expect(actual).toEqual(expected);
+            });
+        }
+
+        describe('mul', () => {
```
VictorVelarde: Shouldn't we split also the tests into different files? Especially now that we have 1 file per operation.

debug/advanced/bivariate-legends.html
elenatorro: Totally! I'd like to have this in a separate task (since this one is a bit big), and it'd be great to have more than one example (but not too much, I mean, like two or three). At leas this one and another one that combines with and color properties.

test/unit/renderer/viz/expressions/binary.test.js
elenatorro: Ok, I agree, I'll change that :)

test/unit/renderer/viz/expressions/binary.test.js
elenatorro: Actually, no, but it's been there for a while and I _think_ there was an example. I'll do a bit of research.

debug/advanced/bivariate-legends.html
VictorVelarde: Fair enough. We can keep the rgbToHex if it is just a debug example.

PR #1379, opened 2019-07-01
Support mixed multi geometries

Jesus89:
Fixes https://github.com/CartoDB/cartoframes/issues/815


Jesus89:
This change only affects the GeoJSON geometry decoding. GeoJSON geom types are converted to Dataframe types (point, line, polygon) after the decoding. In the interactivity, this final geom type is the one used.

PR #1384, opened 2019-07-12
Support histogram expression with dates

elenatorro:
Related with https://github.com/CartoDB/airship/pull/636

Context
===
We weren't supporting dates in histogram expressions, this PR fixes that. We need this feature to fix both Histogram and Time Series widgets in CARTOframes

### DIFF
--- a/src/renderer/viz/expressions/histogram/Histogram.js
+++ b/src/renderer/viz/expressions/histogram/Histogram.js
@@ -82,7 +82,7 @@ export default class Histogram extends BaseExpression {
     }
 
     _getBucketsValue ([...histogram], buckets) {
-        buckets = buckets.length && buckets.length !== 0 ? this._genBreakpoints(buckets) : buckets;
+        buckets = buckets.length && buckets.length === 0 ? this._genBreakpoints(buckets) : buckets;
         const nBuckets = buckets.length;
         const hist = Array(nBuckets).fill(0);
 
--- a/src/renderer/viz/expressions/histogram/Histogram.js
+++ b/src/renderer/viz/expressions/histogram/Histogram.js
@@ -7,9 +7,18 @@ export default class Histogram extends BaseExpression {
     }
 
     get value () {
-        return this.input.type === 'number'
-            ? (this._hasBuckets ? this._getBucketsValue(this._histogram, this._sizeOrBuckets) : this._getNumericValue(this._histogram, this._sizeOrBuckets))
-            : this._getCategoryValue(this._histogram);
+        switch (this.input.type) {
+            case 'number':
+                return this._hasBuckets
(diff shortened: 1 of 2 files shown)


src/renderer/viz/expressions/histogram/Histogram.js
```diff
@@ -7,9 +7,18 @@ export default class Histogram extends BaseExpression {
...
     get value () {
-        return this.input.type === 'number'
-            ? (this._hasBuckets ? this._getBucketsValue(this._histogram, this._sizeOrBuckets) : this._getNumericValue(this._histogram, this._sizeOrBuckets))
-            : this._getCategoryValue(this._histogram);
+        switch (this.input.type) {
+            case 'number':
+                return this._hasBuckets
+                    ? this._getBucketsValue(this._histogram, this._sizeOrBuckets)
+                    : this._getNumericValue(this._histogram, this._sizeOrBuckets);
+            case 'date':
```
VictorVelarde: 👌

src/renderer/viz/expressions/histogram/Histogram.js
```diff
@@ -90,6 +82,7 @@ export default class Histogram extends BaseExpression {
     }
 
     _getBucketsValue ([...histogram], buckets) {
+        buckets = buckets.length === undefined ? this._genBreakpoints(buckets) : buckets;
```
VictorVelarde: `.length === undefined ?`    Is then buckets an object been initialized before and not a simple array? From outside I was expecting something like `buckets === undefined```...

src/renderer/viz/expressions/histogram/Histogram.js
```diff
@@ -111,4 +104,93 @@ export default class Histogram extends BaseExpression {
...
+
+        if (frequencyDifference === 0) {
+            const categoryA = a.x;
+            const categoryB = b.x;
+
+            if (!categoryA && !categoryB) { return 0; } // both null or undefined
+            if (!categoryA) { return 1; } // categoryB first
+            if (!categoryB) { return -1; } // categoryA first
+
+            if (typeof categoryA === 'string' && typeof categoryB === 'string') {
```
VictorVelarde: How about adding a small explaining comment? It feels a bit strange to '_sortNumerically' en then dealing with 'string'

src/renderer/viz/expressions/histogram/Histogram.js
elenatorro: That's because the `buckets` parameter can be a number or an array, i'm going to change the condition to make it clearer

src/renderer/viz/expressions/histogram/Histogram.js
elenatorro: This is how it was before, I can change it to make it more meaningful.

PR #1392, opened 2019-07-19
Fix blend to variables

elenatorro:
- Fix blend value / eval
- Fix toString method in animations
- Support date format in some binary operations
- Add specs

### DIFF
--- a/src/renderer/viz/expressions/between.js
+++ b/src/renderer/viz/expressions/between.js
@@ -50,8 +50,8 @@ export default class Between extends BaseExpression {
     _bindMetadata (meta) {
         super._bindMetadata(meta);
 
-        checkType('between', 'input', 0, 'number', this.input);
-        checkType('between', 'lowerLimit', 1, 'number', this.lowerLimit);
-        checkType('between', 'upperLimit', 2, 'number', this.upperLimit);
+        checkType('between', 'input', 0, ['number', 'date'], this.input);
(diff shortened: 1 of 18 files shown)


src/renderer/viz/expressions/between.js
```diff
@@ -50,8 +50,8 @@ export default class Between extends BaseExpression {
     _bindMetadata (meta) {
         super._bindMetadata(meta);
 
-        checkType('between', 'input', 0, 'number', this.input);
-        checkType('between', 'lowerLimit', 1, 'number', this.lowerLimit);
-        checkType('between', 'upperLimit', 2, 'number', this.upperLimit);
+        checkType('between', 'input', 0, ['number', 'date'], this.input);
```
VictorVelarde: 👍 nice addition

src/renderer/viz/expressions/utils.js
```diff
@@ -247,15 +247,33 @@ export function clamp (x, min, max) {
...
+function _mixDate (x, y, a) {
+    const xTime = x.getTime();
+    const yTime = y.getTime();
+    const timeDiff = Math.abs(_mixNumber(xTime, yTime, a));
+
+    return new Date(timeDiff);
+}
+
+function _mixNumber (x, y, a) {
+    return x * (1 - a) + y * a;
```
VictorVelarde: ✨ Nice refactor!

Issue #1247, opened 2019-01-31
Error in ramp with column name 'length'


VictorVelarde:
As reported internally by @andy-esch an unexpected error is thrown when styling some lines with a ramp using a column named **length**. 

The error is:
`> RangeError: invalid array length`

This comes, for example, from a column in a SQL made with `ST_Length(the_geom::geography) as length` and then using `ramp($length, sunset)` in the viz. Using other column name works fine.

andy-esch:
Hey @VictorVelarde, we have another issue seen first in CartoDB/cartoframes#533 that looks similar to what's above.

andy-esch:
To add to the last comment a bit more, we can reproduce in cartoframes that if interactivity is enabled for a column called `length`, then we get the following error in the console:
`TypeError: legend.key[0].toFixed is not a function[Learn More]`

If the column name is changed to, for example, `length_` then the map renders as expected.

andy-esch:
Nope, that previous comment is wrong -- that error was from something else. As far as I can tell, there is no error thrown.

Here's a reproducible output: https://bl.ocks.org/andy-esch/c74297f7fbedaae5c42cc48eab5cbc6c

Changing the `length` to `length_` will allow the map to render.

VictorVelarde:
We'll check it out, but probably the best thing to do here is to detect those JS 'reserved words' in VL and raise a descriptive error

Jesus89:
Fixed in #1410. It will be ready in the next patch release

PR #1410, opened 2019-11-06
Fix using length as a property name

Jesus89:
Fixes https://github.com/CartoDB/carto-vl/issues/1247

### DIFF
--- a/src/renderer/dataframe/Dataframe.js
+++ b/src/renderer/dataframe/Dataframe.js
@@ -152,13 +152,13 @@ export default class Dataframe extends DummyDataframe {
     }
 
     getPropertyTexture (propertyName) {
-        const encodedPropertyName = `_${propertyName}`;
+        const encodedPropertyName = this._encodePropertyName(propertyName);
 
         if (this.propertyTex[encodedPropertyName]) {
             return this.propertyTex[encodedPropertyName];
         }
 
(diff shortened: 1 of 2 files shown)


Jesus89:
It seems that render tests are broken because the docker image is not updated.

elenatorro:
Yes, you're right, we're trying to fix the image 👍

src/renderer/dataframe/Dataframe.js
```diff
@@ -152,12 +152,14 @@ export default class Dataframe extends DummyDataframe {
...
     getPropertyTexture (propertyName) {
-        if (this.propertyTex[propertyName]) {
-            return this.propertyTex[propertyName];
+        const encodedPropertyName = this._encodePropertyName(propertyName);
+
+        if (this.propertyTex[encodedPropertyName]) {
+            return this.propertyTex[encodedPropertyName];
         }
 
         this._loadPropertyValuesToTexture(propertyName);
```
elenatorro: We can pass here `encodedPropertyName` and avoid to encode it again in the method, right? It's only called here.

Jesus89: Nope, because `_loadPropertyValuesToTexture` uses `propertyName` to access the `properties` and the encodedPropertyName is only for the textures.

PR #1417, opened 2019-12-11
Update browser support documentation

csubira:
https://github.com/CartoDB/support/issues/2296

### DIFF
--- a/docs/support/03-browser-support.md
+++ b/docs/support/03-browser-support.md
@@ -18,6 +18,6 @@ For a good experience, >= 8GB or RAM is recommended and also having a dedicated 
 
 Some browsers, specially in mobile devices, only have experimental or partial WebGL support. We cannot guarantee that CARTO VL will work on those cases.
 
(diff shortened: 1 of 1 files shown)


pablomoniz:
I'm ok too. Thank you!!

oriolbx:
@csubira I'm not seeing this change on the Dev Center 😨 

[https://carto.com/developers/carto-vl/support/browser-support/](https://carto.com/developers/carto-vl/support/browser-support/)

Screenshot 2019-12-24 at 15 13 28

csubira:
It's not in production yet, it would be when the new patch version of CARTO VL has been released. _(Planned before the end of this week)_

csubira:
@oriolbx in production! Check it out!

oriolbx:
Thanks @csubira !!

Eh2406:
So before this change when our (a Carto customer) map shows the **wrong** data, on a device that passes `isBrowserSupported`, I can report it as a Carto vl Bug. Admittedly a hard to track down difficult to fix bug, but a bug nonetheless.  After this change if the map does not work, it may not be a Bug it may be a silently unsupported device. Under what conditions is it still a bug for the map to be completely wrong?

PR #1426, opened 2020-01-02
Remove default 'key' color

elenatorro:
Related issue https://github.com/CartoDB/airship/issues/657

Although it seemed to be an Airship issue, this should be fixed in CARTO VL. This change doesn't have backwards compatibility issues, and it will be fixed in both Airship and CARTOframes once we release a patch with this change


makella:
woo hoo!!

